PR #19, opened 2016-06-24
thanks to irill & Mozilla

sylvestre:


jankeromnes:
Great idea! Thank you @sylvestre, I should have added this way earlier.

PR #36, opened 2017-01-24
Oauth2

jankeromnes:
@bnjbvr Following your advice, I stopped trying to make OAuth2 authentication (client ID + client secret) a common middleware that intercepts all requests, because this authentication will only be accepted for a few specific requests anyway. Instead, I will just authenticate OAuth2 clients where it's actually useful.

Note: In Janitor-land, the OAuth2 "service" (where user accounts and API live) is the main server, and OAuth2 "clients" are secondary cluster hosts (where users access some resources, but an OAuth2 handshake with the service is required to authenticate them).

PTAL 😀

### DIFF
--- a/api/hosts-api.js
+++ b/api/hosts-api.js
@@ -10,11 +10,8 @@ let log = require('../lib/log');
 let machines = require('../lib/machines');
 let users = require('../lib/users');
 
-
 // API resource to manage Janitor cluster hosts.
-
-let hostsAPI = selfapi({
-
+let hostsAPI = module.exports = selfapi({
   title: 'Hosts',
 
   beforeTests: (callback) => {
@@ -34,20 +31,14 @@ let hostsAPI = selfapi({
       callback(error);
     });
   }
-
 });
 
-module.exports = hostsAPI;
-
-
 hostsAPI.get({
-
   title: 'List hosts',
-
   description: 'List all cluster hosts owned by the authenticated user.',
 
   handler: (request, response) => {
-    let user = request.user;
+    let { user } = request;
     if (!users.isAdmin(user)) {
       response.statusCode = 403; // Forbidden
       response.json({ error: 'Unauthorized' });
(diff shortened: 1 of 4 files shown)


api/hosts-api.js
```diff
@@ -67,32 +58,36 @@ hostsAPI.get({
...
-
   title: 'Get a single host',
 
   handler: (request, response) => {
-    let user = request.user;
-    let scope = request.scope;
-    let hostname = request.query.hostname;
+    let { user } = request;
+    let { hostname, client_id = null, client_secret = null } = request.query;
+    let authenticatedHostname = hosts.authenticate(client_id, client_secret);
```
bnjbvr: Also, would it make sense to hide that `hosts.authenticate` uses the query `client_id` and `client_secret`, by having `hosts.authenticate` take the `request` or `request.query` as a parameter instead?

api/hosts-api.js
```diff
@@ -120,20 +115,16 @@ hostAPI.get({
...
-
   description: 'Create a new host and add it to the cluster.',
 
   handler: (request, response) => {
-    let user = request.user;
-    let scope = request.scope;
-    let hostname = request.query.hostname;
+    let { user } = request;
+    let { hostname, client_id = null, client_secret = null } = request.query;
+    let authenticatedHostname = hosts.authenticate(client_id, client_secret);
```
bnjbvr: (same question as above for the `hostname` checking)

api/hosts-api.js
```diff
@@ -67,32 +58,36 @@ hostsAPI.get({
...
 
-    // User or host OAuth2 authentication.
-    if (users.isAdmin(user) || scope.hostname === hostname) {
+    // Host OAuth2 authentication.
+    if (authenticatedHostname && authenticatedHostname === hostname) {
+      let host = hosts.get(hostname);
+      if (host) {
+        response.json(host.properties);
+        return;
+      }
```
bnjbvr: This whole chunk is *almost* the same as the one below, craving for a refactoring... 😛

api/hosts-api.js
jankeromnes: Yes to both suggestions, thanks!

api/hosts-api.js
jankeromnes: I've been there too 😛 however, I believe the code is clearer un-refactored. The slight duplication of code makes it easier to read, understand, and later tweak (I have reasons to believe that the host- and user-authenticated paths will subtly diverge over time, as users will be able to own multiple hosts, but hosts will only be able to control their own data).

api/hosts-api.js
jankeromnes: (addressed in follow-up commit)

jankeromnes:
@bnjbvr PTAL (take 2) 😀

jankeromnes:
👍

PR #37, opened 2017-01-30
Issue OAuth2 authorizations and access tokens

jankeromnes:


jankeromnes:
This commit starts to implement an OAuth2 single-sign-on scheme between all Janitor cluster hosts: The main Janitor server acts as the identity provider (because every user has to be signed in with it), and all the secondary Janitor servers initiate a transparent OAuth2 handshake with the main server to know which users are allowed to access which resources.

An example for how this works goes as follows:
- User tries to access https://moz1.janitor.technology/abcde1234567890/8080 (supposed proxy access to container `abcde1234567890` at port `8080`, if it exists on that host and the user actually owns it)
- https://moz1.janitor.technology doesn't know User yet, so it 302-redirects them to https://janitor.technology/login/oauth/authorize for authorization (with a few query parameters, like the OAuth2 `client_id` and the `redirect_url`)
- https://janitor.technology knows User, so it issues an authorization code, and 302-redirects back to https://moz1.janitor.technology/abcde1234567890/8080?code=1234
- This time, https://moz1.janitor.technology has a code, so it sends a separate POST request behind the scenes to https://janitor.technology/login/oauth/access_token (with the query parameters `code`, and secret OAuth2 credentials that User shouldn't know)
- While User still waits for https://moz1.janitor.technology/abcde1234567890/8080?code=1234, the parallel POST request comes back to moz1.janitor.technology with an `access_token`, that moz1.janitor.technology associates to User's session (it then uses the `access_token` with Janitor's API to verify User actually has ownership of container `abcde1234567890` and port `8080`, and then proceeds to proxy the request)

@bnjbvr if you have the time, could you please take a look at this code change? (No need to deeply verify the OAuth2 logic if you don't want to, I'll request a separate security review of the whole OAuth2 implementation.) Thanks!

app.js
```diff
@@ -138,6 +138,49 @@ boot.executeInParallel([
...
+        return;
+      }
+      routes.redirect(query.res, data.redirect_url);
+    });
+  });
+
+  // OAuth2 access token request.
+  app.route(/^\/login\/oauth\/access_token\/?$/, (data, match, end, query) => {
+    let { req: request, res: response } = query;
+    let { user } = request;
```
bnjbvr: nit: `user` is only used once, no need for a temporary?

app.js
```diff
@@ -138,6 +138,49 @@ boot.executeInParallel([
...
+    if (request.method !== 'POST') {
+      routes.notFoundPage(user, end, query);
+      return;
+    }
+
+    let authenticatedHostname = hosts.authenticate(query.req);
```
bnjbvr: nit: `query.req` => `request`

lib/hosts.js
```diff
@@ -121,22 +133,98 @@ exports.destroy = function (hostname, callback) {
...
+// Pre-authorize a host to make OAuth2 requests on behalf of a user.
+exports.issueOAuth2AuthorizationCode = function (request, callback) {
+  let { client_id, scope, state, redirect_url = null } = request.query;
+  let hostname = exports.identify(client_id);
+  if (!hostname) {
+    callback(new Error('No such OAuth2 client ID: ' + client_id));
+    return;
+  }
+
+  if (redirect_url && !redirect_url.startsWith('https://' + hostname + '/')) {
```
bnjbvr: Is it a bit strict to check for the trailing slash?

lib/hosts.js
```diff
@@ -121,22 +133,98 @@ exports.destroy = function (hostname, callback) {
...
+    callback(new Error('Invalid OAuth2 redirect URL: ' + redirect_url));
+    return;
+  }
+
+  if (!redirect_url) {
```
bnjbvr: nit: I'd merge this `if` with the one above, to make it clear they're mutually exclusive.

```
if (!redirect_url) {
  ...
} else if (!redirect_url.startsWith(...)) {

}
```

lib/hosts.js
```diff
@@ -121,22 +133,98 @@ exports.destroy = function (hostname, callback) {
...
+    redirect_url = 'https://' + hostname + '/';
+  }
+
+  let grant = {
+    email: request.user.email,
```
bnjbvr: Maybe safe to check that `request.user` is defined here (even if it's just an assertion)? (this function is only called by another that checks that it's actually defined at the moment, but that'll make future debugging easier)

lib/hosts.js
```diff
@@ -121,22 +133,98 @@ exports.destroy = function (hostname, callback) {
...
+
+    let { scope: grant, token, tokenHash } = data;
+    let authorization = {
+      date: Date.now(),
+      email: grant.email,
+      scopes: grant.scopes
+    };
+
+    let oauth2tokens = db.get('oauth2tokens');
+    oauth2tokens[tokenHash] = authorization;
```
bnjbvr: Not sure how the hash is computed, but in case of collision, what's the worst that could happen? Could a collision realistically happen? Is there a way to just not have to worry about this at all?

app.js
```diff
@@ -138,6 +138,49 @@ boot.executeInParallel([
     routes.loginPage(end);
   });
 
+  // User OAuth2 authorization.
+  app.route(/^\/login\/oauth\/authorize\/?$/, (data, match, end, query) => {
+    let { user } = query.req;
+    if (!user) {
+      routes.notFoundPage(user, end, query);
```
bnjbvr: Shouldn't this redirect the user to the login page, with a `?next` query parameter which value is this route? (maybe there's already a middleware doing this)

lib/hosts.js
```diff
@@ -54,11 +53,26 @@ exports.authenticate = function (request) {
...
 
+// Find a Docker hostname that matches just the given OAuth2 client ID.
+exports.identify = function (clientId) {
+  let hosts = db.get('hosts');
+  for (let hostname in hosts) {
+    let host = hosts[hostname];
+    if (!host || !host.oauth2client) {
+      continue;
+    }
+    if (String(clientId) === host.oauth2client.id) {
```
bnjbvr: A sanity check (assertion) that `host.oauth2client.id` is not an empty string might be nice here.

jankeromnes:
Thanks a lot for the review! Much helpful and appreciated :)

> Haven't looked in detail at the oauth2 code, but I haven't spotted anything deeply wrong.
> Please make sure to have somebody do a thorough review though ;-)

Sure thing, I'll request a separate security review of all the OAuth2 mechanism soon.

app.js
jankeromnes: Indeed! To quote my uncommitted comment on the very next line:

> // FIXME redirect to login. then retry?

Thanks for your suggestion, I'll fix this now rather than later.

lib/hosts.js
jankeromnes: I don't think it's too strict, because we wouldn't want to accept a `redirect_url` like `"https://moz1.janitor.technology.free.fr/steal/codes"`, and requiring a trailing `/` in automatically-generated `redirect_url` parameters seems like a cheap way to prevent this (as opposed to further URL parsing). I may be wrong though.

lib/hosts.js
jankeromnes: Fair enough, I'll use your suggested variant.

lib/hosts.js
jankeromnes: The used [hash function](https://github.com/jankeromnes/oauth2provider/blob/master/index.js#L140) is a hexadecimal SHA256 digest, and the hashed data is a cryptographically strong pseudo-random 40-digit hexadecimal string. There is currently no hash involved.

Hash collisions could only happen if 1) there is a major bug in the token generator library causing it to issue the same value twice (the [current version](https://github.com/jankeromnes/oauth2provider/blob/master/index.js#L103) looks about sound), or 2) there is 40-digit random hex collision, or 3) there is an accidental SHA256 collision. 2 and 3 seem unrealistic to me, but nevertheless I agree that collisions should be looked out for. I'll add a sanity check (no need for fancy collision-prevention mechanisms I believe).

lib/hosts.js
jankeromnes: Ok, I'll add a `|| !host.oauth2client.id` condition to the above bail-out `if` block.

app.js
jankeromnes: On second thought, I left this for later because the current sign-in mechanism involves an email with a sign-in link, so when you're actually signing in you're in a different tab (possibly even in a different browser or device).

I'll try to find a good UX for this when there are more sign-in options, in the meantime it should not happen very often that you try to access a secondary host while not being signed in to the main host (rather, it's probably the main host that was redirecting you) so I guess it's ok if you have to repeat your request after signing in (sorry users! I know this is frustrating. If this bothers you, please suggest a technical solution).

app.js
```diff
@@ -138,6 +138,49 @@ boot.executeInParallel([
...
+      return;
+    }
+
+    hosts.issueOAuth2AuthorizationCode(query.req, (error, data) => {
+      if (error) {
+        log('[fail] oauth2 authorize', error);
+        routes.notFoundPage(user, end, query);
```
jankeromnes: You're right, this failure should probably show a more informative page to the user, explaining the conditions that triggered it, but then again this failure should be pretty rare (mostly OAuth2 sanity problems, which can happen during OAuth2 refactors, or if a blacklisted host tries to use the service). I'll simply add a comment so we can reconsider if this error starts becoming somewhat frequent.

jankeromnes:
🎉

PR #38, opened 2017-02-10
Authenticate access token, scope and user in OAuth2 requests to the Janitor API.

jankeromnes:
Hi @bnjbvr, could you please take a look at this pull request when you have time?

Now that we can issue OAuth2 access tokens, we also need to make them useful for something. This pull request implements the authentication of access tokens, which, if successful, gives us an email (the user who authorized this access) and a scope (which limits the access that is granted).

It also changes the `/api/hosts/:hostname/:container/:port` API endpoint to accept authentication by OAuth2 access tokens, provided that the scope includes `user:ports` (or simply `user`, which is a super-scope for all `user:*` scopes).

Again, please don't worry too much about verifying the OAuth2 algorithm itself and its security, because I'll request a security audit of those anyway.

I'm mostly worried I might have made an obvious mistake, or made a bad decision somewhere. Also, I'm not too happy with the function names `users.getOAuth2Access()` and `hosts.authenticateOAuth2Request()`. Both return a bunch of things associated to a valid access token, but the former actually just calls the latter to replace `email` (saved with the access token) by an actual `user` representation. If you have other naming or implementation suggestions, I'm all ear!

And again, thank you so much for reviewing my changes! Please feel free to delay or ignore my pull requests if you don't have much time.

### DIFF
--- a/app.js
+++ b/app.js
@@ -72,13 +72,7 @@ boot.executeInParallel([
 
   // Authenticate OAuth2 requests with a server middleware.
   app.use((request, response, next) => {
-    let access = users.getOAuth2Access(request);
-    if (access) {
-      let { hostname, scope, user } = access;
-      let scopes = scope.split(',');
-      request.oauth2access = { hostname, scopes, user };
-    }
-
+    request.oauth2scope = users.getOAuth2ScopeWithUser(request);
     next();
   });
 
--- a/app.js
+++ b/app.js
@@ -62,12 +62,24 @@ boot.executeInParallel([
     next();
   });
 
-  // Authenticate all user requests with a server middleware.
+  // Authenticate signed-in user requests with a server middleware.
   app.use((request, response, next) => {
     users.get(request, user => {
       request.user = user;
       next();
     });
+  });
+
+  // Authenticate OAuth2 requests with a server middleware.
+  app.use((request, response, next) => {
+    let access = users.getOAuth2Access(request);
+    if (access) {
+      let { hostname, scope, user } = access;
+      let scopes = scope.split(',');
+      request.oauth2access = { hostname, scopes, user };
+    }
+
+    next();
   });
 
   // Mount the Janitor API.
@@ -82,18 +94,15 @@ boot.executeInParallel([
 
   // Public blog page.
   app.route(/^\/blog\/?$/, (data, match, end, query) => {
-    let user = query.req.user;
-
+    let { user } = query.req;
     log('blog');
-
-    return routes.blogPage(user, end);
+    routes.blogPage(user, end);
   });
 
   // Public live data page.
   app.route(/^\/data\/?$/, (data, match, end, query) => {
-    let user = query.req.user;
-
-    return routes.dataPage(user, end);
+    let { user } = query.req;
+    routes.dataPage(user, end);
   });
 
   // Public project pages.
@@ -103,7 +112,8 @@ boot.executeInParallel([
 
     if (!projectUri) {
       // No particular project was requested, show them all.
-      return routes.projectsPage(user, end);
+      routes.projectsPage(user, end);
+      return;
     }
 
     var projectId = projectUri.slice(1);
@@ -121,7 +131,7 @@ boot.executeInParallel([
   app.route(/^\/logout\/?$/, (data, match, end, query) => {
     users.logout(query.req, error => {
       if (error) {
-        log('logout', error);
+        log('[fail] logout', error);
       }
       routes.redirect(query.res, '/');
     });
@@ -129,12 +139,11 @@ boot.executeInParallel([
 
   // User login.
   app.route(/^\/login\/?$/, (data, match, end, query) => {
-    let user = query.req.user;
+    let { user } = query.req;
     if (user) {
       routes.redirect(query.res, '/');
       return;
     }
-
     routes.loginPage(end);
   });
 
@@ -185,7 +194,7 @@ boot.executeInParallel([
 
   // User contributions list.
   app.route(/^\/contributions\/?$/, (data, match, end, query) => {
-    let user = query.req.user;
+    let { user } = query.req;
     if (!user) {
       routes.loginPage(end);
       return;
@@ -196,7 +205,7 @@ boot.executeInParallel([
 
   // User settings.
   app.route(/^\/settings(\/\w+)?\/?$/, (data, match, end, query) => {
-    let user = query.req.user;
+    let { user } = query.req;
     if (!user) {
       routes.loginPage(end);
       return;
@@ -229,7 +238,7 @@ boot.executeInParallel([
 
   // Admin sections.
   app.route(/^\/admin(\/\w+)?\/?$/, (data, match, end, query) => {
-    let user = query.req.user;
+    let { user } = query.req;
     if (!users.isAdmin(user)) {
       routes.notFoundPage(user, end, query);
       return;
@@ -246,7 +255,7 @@ boot.executeInParallel([
 
   // Secure VNC connection proxy.
   app.route(/^\/vnc\/(\w+)\/(\d+)(\/.*)$/, (data, match, end, query) => {
-    let user = query.req.user;
+    let { user } = query.req;
     if (!user) {
       routes.notFoundPage(user, end, query);
       return;
@@ -315,7 +324,7 @@ boot.executeInParallel([
 
   // 404 Not Found.
   app.notfound(/.*/, (data, match, end, query) => {
-    let user = query.req.user;
+    let { user } = query.req;
 
     log('404', match[0]);
 
@@ -348,7 +357,7 @@ boot.executeInParallel([
 
   // Alpha version invite.
   app.ajax.on('invite', (data, end, query) => {
-    let user = query.req.user;
+    let { user } = query.req;
     if (!users.isAdmin(user)) {
       end();
       return;
@@ -373,7 +382,7 @@ boot.executeInParallel([
 
   // Request a log-in key via email.
   app.ajax.on('login', (data, end, query) => {
-    let user = query.req.user;
+    let { user } = query.req;
     if (user) {
       end({ status: 'logged-in' });
       return;
@@ -393,7 +402,7 @@ boot.executeInParallel([
 
   // Change the parameters of a project.
   app.ajax.on('projectdb', (data, end, query) => {
-    let user = query.req.user;
+    let { user } = query.req;
     if (!users.isAdmin(user)) {
       end();
       return;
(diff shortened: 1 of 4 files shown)


lib/hosts.js
```diff
@@ -238,3 +238,28 @@ exports.issueOAuth2AccessToken = function (request, callback) {
...
+
+  let tokenHash = oauth2provider.hash(token);
+  let oauth2tokens = db.get('oauth2tokens');
+  let authorization = oauth2tokens[tokenHash];
+  if (!authorization) {
+    return null;
+  }
+
+  let { client, email, scopes } = authorization;
+  let hostname = exports.identify(client);
```
bnjbvr: Can `hostname` be null here?

lib/hosts.js
```diff
@@ -238,3 +238,28 @@ exports.issueOAuth2AccessToken = function (request, callback) {
   // Attempt to generate an access token with hash.
   oauth2provider.generateAccessToken(client_id, code, state, onToken);
 };
+
+// Find the authorized scope of a request's OAuth2 access token.
+exports.authenticateOAuth2Request = function (request) {
+  let token = request.query.access_token || null;
+  if ('authorization' in request.headers) {
+    token = request.headers['authorization'].split(/\s+/)[1];
```
bnjbvr: The intent is not clear here: do you want to clobber the token if there's `query.access_token` but an Authorization header is present? If so, a comment might clarify that.

app.js
```diff
@@ -62,14 +62,26 @@ boot.executeInParallel([
...
       next();
     });
   });
 
+  // Authenticate OAuth2 requests with a server middleware.
+  app.use((request, response, next) => {
+    let access = users.getOAuth2Access(request);
+    if (access) {
+      let { hostname, scope, user } = access;
+      let scopes = scope.split(',');
```
bnjbvr: Would it make sense that `users.getOAuth2Access().scope` just be the array of scopes? This code is undoing (split) what the callee did (join).

lib/users.js
```diff
@@ -56,16 +57,30 @@ exports.isAdmin = function (user) {
...
+
   return false;
 };
 
+// Find the user that authorized a given OAuth2 request, and its access scope.
+exports.getOAuth2Access = function (request) {
+  let authorization = hosts.authenticateOAuth2Request(request);
+  if (authorization) {
+    let { email, hostname, scope } = authorization;
+    let user = getUser(email);
```
bnjbvr: Can `user` be null?

app.js
bnjbvr: Or even better, maybe: will scopes get attached to them, later? If not, you could just use a `Set` of scopes, making you avoid using `Array.includes` for a lookup.

lib/hosts.js
```diff
@@ -238,3 +238,28 @@ exports.issueOAuth2AccessToken = function (request, callback) {
   // Attempt to generate an access token with hash.
   oauth2provider.generateAccessToken(client_id, code, state, onToken);
 };
+
+// Find the authorized scope of a request's OAuth2 access token.
+exports.authenticateOAuth2Request = function (request) {
```
bnjbvr: Two general comments:

- I think the fact that you're using OAuth2 for authentication is almost an implementation detail (that doesn't mean it's easy to implement). Considering this, maybe `authFromRequest` would be a simpler name here. The other function could then just be `authWithUserFromRequest`.
- it seems to me that since the other function is just augmenting this function's return value, it should be in that file too, right thereafter.

bnjbvr: Alternative names: `oauth2FromRequest` / `oauth2WithUserFromRequest`, requestToOauth2` / `requestToOauth2AndUser`... (I usually don't fear long and overly descriptive names, since they're just a few keystrokes away, thanks to auto-completion)

lib/users.js
bnjbvr: I'd also invert the condition, because having an authorization seems to be the general case, not the particular case, here:

```
if (!authorization)
  return null;

let { ... } = authorization
etc();
```

jankeromnes:
Thanks a lot for the great review!

> I've seen a few ES6 changes that distracted me from the initial review and made it look much longer than it actually was; perhaps these changes could go in a separate commit next time, to reduce review friction.

Sorry about that! I do tend to refactor some parts of a file while I make changes in it, thinking that the changes are trivial and don't deserve their own commits. However, I agree that it adds unrelated commit messages to the blame of these lines, and that they can be distracting. I'll avoid such noise in the future by making separate clean-up commits. Thanks for the suggestion!

lib/hosts.js
jankeromnes: It's not supposed to be `null`, unless something changes in the token or host DB representations (and it already happened during manual testing...). Good catch! I'll add a safeguard for this.

lib/users.js
jankeromnes: The function `getUser()` is more of a `getOrCreateUser()`, so `user` should never be `null`. Maybe I can rename the function to make that more explicit?

lib/hosts.js
julienw: nit: you can use `request.headers.authorization` instead of the array notation.

I believe the intent is to support `Authorization: Bearer <token>` for API calls, and a query parameter for in-browser calls. I agree a comment would be a good idea.

lib/users.js
julienw: nit: you can use `const` instead of `let`

lib/hosts.js
```diff
@@ -238,3 +238,28 @@ exports.issueOAuth2AccessToken = function (request, callback) {
...
+  if (!authorization) {
+    return null;
+  }
+
+  let { client, email, scopes } = authorization;
+  let hostname = exports.identify(client);
+  let scope = scopes.join(',');
+
+  return { email, hostname, scope };
+};
```
julienw: nit:you can use `const` instead of `let` for everything except `token`

app.js
julienw: I believe this is what other oauth2 providers are doing.

julienw:
It's not clear to me which part you're modifying here. Is it the OAuth2 provider part (that's also a OAuth2 client to github and others ? Or is it the OAuth2 client part ?

notriddle:
Janitor doesn't have an OAuth2 client (yet). This is for Janitor's OAuth2 provider.

jankeromnes:
Thanks a lot for your review and useful suggestions @julienw !

@notriddle is right, there are no OAuth2 clients yet in Janitor (just the `/lib/oauth2.js` OAuth2 client helper module, but no clients for third party services like GitHub, nor clients for Janitor-as-a-provider). This pull request implements the authentication of OAuth2 access tokens on the provider side (for janitor.technology, the main Janitor server), and the clients using these tokens will be other Janitor hosts (Docker hosts like moz1.janitor.technology) who want to join the cluster (that part will be my next pull request).

lib/hosts.js
jankeromnes: Thanks for your suggestions! I think I'll go with `getOAuth2Scope()` / `getOAuth2ScopeWithUser()`, because I believe it's important to explicitly emphasize that these methods determine the "OAuth2 access scope" of each request.

> it seems to me that since the other function is just augmenting this function's return value, it should be in that file too, right thereafter.

The reason both methods are in separate modules is that they use features that are private to each module (no other module than `/lib/hosts.js` should manage host-issued access tokens, and no other module than `/lib/users.js` should be able to manage users). FYI, it's the situation I asked you about in Brussels, while drawing imaginary diagrams on the wall.

lib/hosts.js
jankeromnes: Julien is right, the goal is to support OAuth2 authentication via both query parameters and authorization headers. I'll add comments to explain this.

bnjbvr: Makes sense. Perhaps worth documenting as well that authorization headers are privileged over query parameters, just to not have strange surprises later ;)

app.js
jankeromnes: Good suggestion, I'll change it to a `Set` when loaded, and back to a `String` when saving it in `db.json`. However, I don't get how this prevents "attaching scopes later" (even though it's not something I plan to do).

lib/users.js
jankeromnes: Addressed all nits locally, revised patch coming soon.

lib/hosts.js
jankeromnes: Good idea, I've never used `const` in this code base before, but this is a great place to start!

lib/hosts.js
jankeromnes: Actually now that I think of it, it's probably less surprising to make query parameters prime over headers (since there more obvious). I'll reverse the priority, which should also make the code a bit clearer.

jankeromnes:
@bnjbvr and/or @julienw, care to take another look?

I believe that I addressed all your nits. If you agree with these new changes, I'll fixup the commits and merge the pull request (but feel free to request more changes if needed).

lib/users.js
```diff
@@ -194,7 +194,7 @@ exports.sendInviteEmail = function (email, callback) {
 };
 
 // Find an existing user, or create a new one.
-function getUser (email) {
+function getOrCreateUser (email) {
```
bnjbvr: This is more explicit, thanks.

lib/hosts.js
```diff
@@ -239,27 +241,57 @@ exports.issueOAuth2AccessToken = function (request, callback) {
...
   }
 
-  let { client, email, scopes } = authorization;
-  let hostname = exports.identify(client);
-  let scope = scopes.join(',');
+  const { client, email, scope } = authorization;
+  const hostname = exports.identify(client);
+  if (!hostname) {
+    // The authorized OAuth2 client doesn't exist anymore.
+    delete oauth2tokens[tokenHash];
```
bnjbvr: Good catch. Does it need to be saved in the db as well? Maybe it could lead to the error (above) `OAuth2 token hash already exists` if it's not saved.

lib/hosts.js
```diff
@@ -239,27 +241,57 @@ exports.issueOAuth2AccessToken = function (request, callback) {
...
+    return null;
+  }
 
-  return { email, hostname, scope };
+  const scopes = parseScopes(scope);
+  if (!scopes) {
+    // The authorized scope is invalid.
+    delete oauth2tokens[tokenHash];
```
bnjbvr: (and here too)

lib/hosts.js
jankeromnes: It's not really necessary to call `db.save()` here, because the DB might be persisted by some other code later, and even if this change is not persisted and a server crash happens, any follow-up authentication attempt would re-trigger the same code path and delete the token again, guaranteeing consistent behavior and eventual DB consistency.

Also I believe it wouldn't trigger the `token hash already exists` error, which may only happen just after a brand-new token is generated (which should never have the same hash as a previously generated token).

Still, maybe it's cleaner to persist it here directly. I'll do that before merging this PR.

jankeromnes:
Thanks again for the great reviews! Squashed and merged as 7e42a0b273a6b550ff0252087927a2f3c6446ed0 and 0f708246354de6238f5bc28fc8010764799c377f.

PR #40, opened 2017-02-16
Implement an authenticated proxy to protect Docker containers on Janitor cluster hosts.

jankeromnes:


jankeromnes:
This pull request implements the OAuth2 client part (which interacts with the OAuth2 provider part from the previous pull requests), as well as a basic Janitor host-level proxy that:

- takes requests like `https://moz1.janitor.technology/abcdef0123456789/8080/index.html`,
- authenticates the user via OAuth2,
- uses the Janitor API to verify that user actually owns container `abcdef0123456789`,
- (also checks what host-level port the container's `8080` port is mapped to, and what type of proxy should be used),
- and if all went well, proxies the request to `http://localhost:8080/index.html` in container `abcdef0123456789`.

@bnjbvr, could you please have a look when you get the chance?

notriddle:
Just to make sure I'm understanding correctly, this is going to be running c9 in the container? Or is it for working on web apps in Janitor?

jankeromnes:
@notriddle I agree my explanations are unclear, but you're correct! It's for any kind of container port that need to be proxied. It could allow running Cloud9 IDE in the containers directly (which me might do using their SDK), and also to work on web apps in Janitor (e.g. Janitor, or the upcoming Kresus project), but the reason I need it now for the cluster is that noVNC's web interface is also exposed on an insecure container port.

In the future, it will be possible to configure such port mappings for each project, e.g. if you want to route additional ports (like temporary web servers running in tests, Firefox does this and developers might want to access these types of servers).

jankeromnes:
Ok, this new version seems to work better.

@bnjbvr, could you please have look when you have the time? 😄

jankeromnes:
P.S. Sorry for the distracting ES6-fixups in `/lib/boot.js`! This commit pre-dates your comment about not liking these, and I haven't taken the time to split it up.

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+      response.statusCode = 400; // Bad Request
+      response.end();
+    }
+    return;
+  }
+
+  // Use the Janitor API to check which local host port the requested Docker
+  // container port is mapped to. This also verifies that the container exists
+  // on this host, and that the authenticated user is allowed to access it.
+  getMappedPort(oauth2Tokens[session.id], container, port, (error, data) => {
```
bnjbvr: nit: `(error, { port, proxy })`, not `data`

Then further down:

```
const proxyParameters = { port, proxy };
```

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+      response.end();
+      return;
+  }
+}
+
+// Use the Janitor API to get the mapping information of a given container port.
+function getMappedPort (accessToken, container, port, callback) {
+  let parameters = {
+    provider: 'janitor',
+    accessToken: accessToken,
```
bnjbvr: nit: ES6 destructuring (or so) allows you to write `{ a: a }` as just `{ a }`.

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+}
+
+// Get the Janitor OAuth2 provider's authorization URL for scope 'user:ports'.
+function getOAuth2AuthorizationUrl (redirectUrl, state, callback) {
+  const parameters = {
+    provider: 'janitor',
+    options: {
+      redirect_url: 'https://' + hostname + redirectUrl,
+      scope: [ 'user:ports' ],
+      state: state
```
bnjbvr: nit: just `state` here too

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+    }
+  });
+}
+
+// Get the Janitor OAuth2 provider's authorization URL for scope 'user:ports'.
+function getOAuth2AuthorizationUrl (redirectUrl, state, callback) {
+  const parameters = {
+    provider: 'janitor',
+    options: {
+      redirect_url: 'https://' + hostname + redirectUrl,
```
bnjbvr: Do you need a slash between hostname and redirectUrl? If not, can you assert it's already there?

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+  });
+}
+
+// Request an OAuth2 access token in exchange of an authorization code.
+function getOAuth2AccessToken (code, state, callback) {
+  const parameters = {
+    provider: 'janitor',
+    code: code,
+    options: {
+      state: state
```
bnjbvr: nit: two opportunities for destructuring here

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+    }
+  };
+
+  oauth2.getAccessToken(parameters, (error, accessToken, refreshToken) => {
+    callback(error, accessToken, refreshToken);
+  });
```
bnjbvr: could just be `oauth2.getAccessToken(parameters, callback);`

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+    oauth2Tokens[session.id] = accessToken;
+    next();
+  });
+}
+
+// Ensure that all requests are authenticated via OAuth2.
+function ensureOAuth2Access (request, response, next) {
+  const { session } = request;
+  if (oauth2Tokens[session.id]) {
+    // This session has an OAuth2 access token, so it's authenticated.
```
bnjbvr: What if it's been invalidated because it's too old?

app.js
```diff
@@ -247,72 +247,59 @@ boot.executeInParallel([
...
     }
 
-    let projectId = match[1];
-    let machineId = parseInt(match[2], 10);
-    let uri = path.normalize(match[3]);
-
-    log('vnc', projectId, machineId, uri);
+    const container = match[1];
+    const port = String(match[2]);
+    const path = nodepath.normalize(match[3]);
```
jankeromnes: Good catch! The better behavior is to check if `path` is empty, and to perma-redirect to a URL with a trailing `/`, as `join.js` does, but for this (deprecated) handler it's OK to make the `path` regex group non-optional (I'll remove the `?` from the regex).

app.js
```diff
@@ -247,72 +247,59 @@ boot.executeInParallel([
...
     }
 
+    // Remember this port for the WebSocket proxy (see below).
+    user.lastProxyPort = mappedPort.port;
+    routes.webProxy({ port: mappedPort.port, path }, query.req, query.res);
+  });
+
+  // FIXME: Remove this deprecated handler (see comments above).
+  // Proxy WebSocket connections to local containers.
+  app.on('upgrade', (request, socket, head) => {
```
jankeromnes: Node.js servers receive two major types of connections: `request` connections (which are handled by Camp with the usual stack of internal + custom middlewares, which can process the request and trigger the relevant `app.route()`, `app.path()`, `app.notfound()` and `app.ajax.on()` handlers), and `upgrade` connections (used for WebSockets, which completely bypass Camp).

So yes, in a sense, this is a sort of middleware that gets triggered for all WebSocket connections to this server, in order to either proxy them to one of the local containers, or drop them.

join.js
```diff
@@ -1,13 +1,18 @@
-// Copyright © 2016 Jan Keromnes. All rights reserved.
+// Copyright © 2017 Jan Keromnes. All rights reserved.
```
jankeromnes: I used to update the year whenever I made "significant" changes to the file (e.g. major rewrite or refactor). If that's not a common practice, I'll leave it unchanged. Thanks for telling me!

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+      // port, and a path that starts with a '/'. These anonymous patterns are
+      // captured in `request.query` as '1', '2' and '3', respectively.
+      const { query } = request;
+      const path = query['3'];
+      if (!path) {
+        // We want the proxied `path` to always begin with a '/'.
+        // However `path` is empty in URLs like '/abc123/8080?p=1', so we
+        // redirect them to '/abc123/8080/?p=1' (where `path` is '/').
+        let url = request.url;
+        url = (url.includes('?') ? url.replace('?', '/?') : url + '/');
```
jankeromnes: I'll remove the unnecessary parenthesis.

If `url` already had a trailing `/`, then the captured `path` pattern would not be empty (but `'/'`), so we wouldn't enter this whole if block. We only enter here when there is nothing after `port` in the URL, so no trailing `/`.

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+    proxy.path('/*', proxyRequest);
+
+    // Proxy WebSocket connections to local containers.
+    proxy.on('upgrade', (request, socket, head) => {
+      // Note: 'upgrade' requests bypass any installed middlewares, so we need
+      // to call them here manually to authenticate WebSocket connections.
+      request.query = {};
+      ensureSession(request, socket, () => {
+        ensureOAuth2Access(request, socket, () => {
+          proxyRequest(request, socket);
```
jankeromnes: Yes, all middlewares and handlers here can decide that something is an error case, and terminate the request with an error of their choice (for now, it's simply an error code, but in the future we should implement more explicit / helpful error pages).

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
   });
 });
+
+// Associate some non-persistent data to sessions.
+const oauth2States = {};
+const oauth2Tokens = {};
+const lastProxyParameters = {};
+
+// Assign a stable session to all requests.
+function ensureSession (request, response, next) {
```
jankeromnes: I used to be disturbed by this, but then it stopped bothering me too much, and in this case I think it even improves the readability of the file (first, the "core behavior" of the file, then the implementations of "major functions", then all the little "helpers" that aren't as interesting). Would you find the file easier to read if all functions and helpers were implemented before the "boot sequence" and proxy server?

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+  const { session } = request;
+  const { code, state } = request.query;
+  if (!code) {
+    next();
+    return;
+  }
+
+  // Compare the provided OAuth2 state parameter with actual session state.
+  const expectedState = oauth2States[session.id];
+  if (state !== expectedState) {
```
jankeromnes: How about I coerce them both to Strings?

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+    }
+
+    request.session = session;
+    next();
+  });
+}
+
+// Handle any OAuth2 authorization codes.
+function handleOAuth2Code (request, response, next) {
+  const { session } = request;
```
jankeromnes: Hm, sure. I pinned it at first line like `let { user } = request` in `app.js` handlers, because it's probably the most important variable that all handlers will use, but I agree it's cleaner to create variables just before they're actually used.

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+  if (!code) {
+    next();
+    return;
+  }
+
+  // Compare the provided OAuth2 state parameter with actual session state.
+  const expectedState = oauth2States[session.id];
+  if (state !== expectedState) {
+    log('[fail] bad oauth2 state: got', state, 'but expected', expectedState);
+    response.statusCode = 403; // Forbidden
```
jankeromnes: That map would have to be a separate module used everywhere in the code base, and essentially be a re-implementation of the various RFCs that define all [HTTP status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes). I think I prefer the simpler / less robust option of using on-the-spot codes, along with helpful reminder comments.

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+    response.end();
+    return;
+  }
+
+  // If they match, use the code to request an OAuth2 access token.
+  getOAuth2AccessToken(code, state, (error, accessToken, refreshToken) => {
```
jankeromnes: It's also not generated by the provider side yet. I'll add a comment about supporting that one day.

join.js
jankeromnes: True, or if it's been revoked by the user / the provider. Maybe we should test if that token is still valid here.

But there is probably no need to block the request on this check, as we can probably run that verification asynchronously, and simply nuke the token if it's been invalidated, so that future requests ask for a new token?

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+  });
+}
+
+// Proxy a request to a local Docker container.
+function proxyRequest (request, response) {
+  const { session } = request;
+  const { container, port } = request.query;
+  if (!container || !port) {
+    // No container port was explicitly requested, so we re-use the session's
+    // last proxy parameters.
```
jankeromnes: Yes, exactly. It happens regularly that servers in the Docker container make requests to unprefixed/root-based URLs, e.g. for WebSocket requests (noVNC hits `/websockify`), HTTP redirections (Cloud9's SDK redirects to `/ide.html`), or simple for resources with absolute URLs.

The same thing happens with requests like https://hostname/?id=id&port=port, where `id` and `port` are remembered for all the follow-up requests that won't forward the original query parameters.

join.js
jankeromnes: What if `error` is defined and `data` is undefined? Wouldn't that cause a runtime error?

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+function routeRequest (proxyParameters, request, response) {
+  const path = nodepath.normalize(request.url);
+  const { port, proxy } = proxyParameters;
+  switch (proxy) {
+    case 'https':
+      routes.webProxy({ port, path }, request, response);
+      return;
+
+    case 'none':
+      const url = 'https://' + hostname + ':' + port + path;
```
jankeromnes: In theory, the regex pattern capturing `path` makes sure that it always begins with a `/` (the only case it doesn't is when no `path` pattern is captured, as it is optional, but in this case we perma-redirect to a URL where `path === '/'` and bail out).

join.js
jankeromnes: Good catch! I'll add a sanity check.

join.js
jankeromnes: Still, I'll add a sanity check because it's good practice.

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+    options: {
+      redirect_url: 'https://' + hostname + redirectUrl,
+      scope: [ 'user:ports' ],
+      state: state
+    }
+  };
+
+  oauth2.getAuthorizationUrl(parameters, (error, url) => {
+    callback(error, url);
+  });
```
jankeromnes: I repeated the arguments here in order to make the code more transparent about what to expect in your callback. I guess I could just use `callback` straight, but that would force you to go read the code of another file in order to know what to expect. What do you think is best?

lib/boot.js
```diff
@@ -11,11 +11,8 @@ let oauth2 = require('./oauth2');
 
 let hostname = db.get('hostname', 'localhost');
 
-
 // Run given tasks in parrallel, and only call `next()` when all have succeeded.
-
-function executeInParallel (tasks, next) {
-
+exports.executeInParallel = function (tasks, next) {
```
jankeromnes: With this new style, I sought to remove all unnecessary repetitions of function names.

Compare:

```js
function executeInParallel () {
  // stuff
}
exports.executeInParallel = executeInParallel;
```

with:

```js
exports.executeInParallel = function () {
  // stuff
};
```

Additionally, I'm happy to confirm that V8 is pretty good at using property names in these situations. See this recent example:

```bash
[2017-02-07T14:31:41.709Z] [fail] oauth2 authorize Error: No such OAuth2 client ID: undefined
    at Object.exports.issueOAuth2AuthorizationCode (/home/janx/janitor/lib/hosts.js:161:14)
```

join.js
jankeromnes: My explanation is wrong, because this function gets `path` by normalizing `request.url` directly (no regex pattern involved at this point). I believe this always starts with a `/`, but I'll add a sanity check anyway because you asked about the `/` :)

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+      secure: true,
+      key: https.key,
+      cert: https.crt,
+      ca: https.ca
+    });
+
+    log('[ok] proxy → https://' + hostname + ':' + ports.https);
+
+    // Convenient express-like alias.
+    proxy.use = proxy.handle;
```
jankeromnes: Fair enough, switched back to using `proxy.handle` without an alias.

join.js
jankeromnes: Added a TODO comment for this.

join.js
```diff
@@ -25,5 +30,284 @@ boot.executeInParallel([
...
+function ensureSession (request, response, next) {
+  sessions.get(request, (error, session, token) => {
+    if (error) {
+      log('[fail] session', error);
+      response.statusCode = 500;
+      response.end();
+      return;
+    }
+
+    request.session = session;
```
jankeromnes: (Done in the `if` check above.)

jankeromnes:
In the interest of time, I'll merge this pull request in its current state, but there is some follow-up work left:

- [ ] decide whether to rely on JS hoisting to resolve https://github.com/jankeromnes/janitor/pull/40#discussion_r102187338
- [ ] decide if it's ok to destructure optional callback arguments to resolve https://github.com/jankeromnes/janitor/pull/40#discussion_r102192689
- [ ] decide whether to explicitly repeat arguments or use `callback` directly to resolve https://github.com/jankeromnes/janitor/pull/40#discussion_r102186335 and https://github.com/jankeromnes/janitor/pull/40#discussion_r102193242
- [ ] TODO: check whether client tokens are still valid, discard or rew them if not
- [ ] TODO: save and use `refreshToken`s when they're implemented

app.js
bnjbvr: Makes sense, thanks.

app.js
bnjbvr: OK.

join.js
bnjbvr: Yes, I definitely would. The main reason why I dislike this ordering is that it doesn't allow straight forward reading of the code; one has to read the code, then look up where a given function is defined and jump there; then get back to the initial position of where the code does, think back about the current context, etc.
In practice, the back-and-forth also happens when the code is properly ordered, but I really think it happens much less often. It's a personal taste, of course, so your call :)

join.js
bnjbvr: Ha you're right, nevermind me :)

join.js
bnjbvr: You make a good point; having the long form sounds good.

bnjbvr:
I've answered your previous comments without giving a full look at the code, but your answers sound reasonable to me. Good job, Jan!

PR #43, opened 2017-03-06
Allow spawning multiple containers per project.

jankeromnes:
@bnjbvr PTAL :)

### DIFF
--- a/lib/machines.js
+++ b/lib/machines.js
@@ -13,7 +13,7 @@ const streams = require('./streams');
 exports.getAvailableMachines = function (user) {
   const machines = {};
   for (const projectId in db.get('projects')) {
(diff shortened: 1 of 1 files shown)


lib/machines.js
```diff
@@ -272,7 +267,7 @@ exports.destroy = function (user, projectId, machineId, callback) {
       }
 
       log('destroy', containerId.slice(0, 16), 'success');
-      machine.docker.image = '';
+      delete machine.docker.image;
```
jankeromnes: Sorry, this drive-by change isn't actually related with this commit (I'm just sneaking it in discreetly).

jankeromnes:
Thanks a lot for digging so far into it to double-check the soundness of this change! I should copy/paste this IRC discussion somewhere to identify which details should be documented :)

PR #47, opened 2017-03-15
Refactor server response routes in order to pre-compile all templates.

jankeromnes:
In theory, there should be no functional change here (except for the small `http.ServerResponse` bug-fix I sneaked in).

This pull request switches to [ScoutCamp](https://github.com/espadrine/sc)'s new preferred way to deal with templates, which allows pre-compiling templates with `camp.template()`, and serving templated responses by calling `response.template()` directly instead of relying on `camp.route()`'s `end()` callback (note: `camp.route()` is slowly being deprecated, as is `query`: eventually, every route should use `camp.path(pattern, (request, response) => {})`, but this will happen in future pull requests).

@bnjbvr, could you please verify that I didn't make any obvious mistakes?

@espadrine, could you please have a look at how I handle templates, and if this the fastest way to serve templates? (I'd like to have a look at template/scoutcamp performance in the near future.)

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,18 +1,18 @@
 // Copyright © 2016 Jan Keromnes. All rights reserved.
 // The following code is covered by the AGPL-3.0 license.
 
-let camp = require('@jankeromnes/camp');
-let nodepath = require('path');
-let selfapi = require('selfapi');
+const camp = require('@jankeromnes/camp');
+const nodepath = require('path');
+const selfapi = require('selfapi');
 
-let api = require('./api/');
-let boot = require('./lib/boot');
-let db = require('./lib/db');
-let hosts = require('./lib/hosts');
-let log = require('./lib/log');
-let machines = require('./lib/machines');
-let routes = require('./lib/routes');
-let users = require('./lib/users');
+const api = require('./api/');
+const boot = require('./lib/boot');
+const db = require('./lib/db');
+const hosts = require('./lib/hosts');
+const log = require('./lib/log');
+const machines = require('./lib/machines');
+const routes = require('./lib/routes');
+const users = require('./lib/users');
 
 boot.executeInParallel([
   boot.forwardHttp,
@@ -37,11 +37,8 @@ boot.executeInParallel([
 
   log('Janitor → https://' + hostname + ':' + ports.https);
 
-  // Convenient express-like alias.
-  app.use = app.handle;
-
   // Protect the server and its users with a security policies middleware.
-  app.use((request, response, next) => {
+  app.handle((request, response, next) => {
     // Only accept requests addressed to our hostname, no CDN here.
     if (request.headers.host !== hostname) {
       log('dropping request for', request.headers.host);
@@ -63,7 +60,7 @@ boot.executeInParallel([
   });
 
   // Authenticate signed-in user requests with a server middleware.
-  app.use((request, response, next) => {
+  app.handle((request, response, next) => {
     users.get(request, user => {
       request.user = user;
       next();
@@ -71,7 +68,7 @@ boot.executeInParallel([
   });
 
   // Authenticate OAuth2 requests with a server middleware.
-  app.use((request, response, next) => {
+  app.handle((request, response, next) => {
     request.oauth2scope = users.getOAuth2ScopeWithUser(request);
     next();
   });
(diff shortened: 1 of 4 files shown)


app.js
```diff
@@ -81,44 +78,43 @@ boot.executeInParallel([
...
+    const projectId = projectUri.slice(1);
+    const project = db.get('projects')[projectId];
+    if (!project) {
+      // The requested project doesn't exist.
+      routes.notFoundPage(query.res, user);
+      return;
     }
 
-    return routes.notFoundPage(user, end, query);
+    // Show the requested project-specific page.
```
bnjbvr: nit: not sure of the value of this comment

app.js
```diff
@@ -127,25 +123,27 @@ boot.executeInParallel([
...
   app.route(/^\/login\/?$/, (data, match, end, query) => {
-    let { user } = query.req;
+    const { user } = query.req;
     if (user) {
       routes.redirect(query.res, '/');
       return;
     }
-    routes.loginPage(end);
+
+    routes.loginPage(query.res);
```
bnjbvr: Personal preference, but I'd invert the `if (user)` condition to the opposite `if (!user) { routes.loginPage(query.res); return } routes.redirect...`, because it seems going to the login page is the exception, not the general case.

lib/routes.js
```diff
@@ -1,266 +1,294 @@
...
+const http = require('http');
 
-let db = require('./db');
-let machines = require('./machines');
-let metrics = require('./metrics');
+const db = require('./db');
+const machines = require('./machines');
+const metrics = require('./metrics');
+
+const cwd = process.cwd();
```
bnjbvr: I'd rather use the absolute location of file and build a relative path from it, because the cwd can change according to where you start the application (unless cwd is forced somewhere else), and it's generally useful to not implicitly depend on this.

lib/routes.js
```diff
@@ -1,266 +1,294 @@
...
+
+// Teach the templating system how to generate IDs (matching /[a-z0-9_-]*/).
+camp.templateReader.parsers.id = text => {
+  return text.replace(/[^\w-]/g, '').toLowerCase();
+};
+
+// Redirect to a target URL.
+exports.redirect = function (response, url, permanently = false) {
+  // We can only redirect actual `http.ServerResponse`s, not raw `net.Socket`s.
+  if (response instanceof http.ServerResponse) {
```
bnjbvr: Isn't it that calling `redirect` with a `net.Socket` should even be a bad API usage, thus report something, if not throw?

lib/routes.js
```diff
@@ -1,266 +1,294 @@
...
-  }, { template: [
-    '../templates/header.html',
-    '../templates/blog.html',
-    '../templates/footer.html'
-  ]});
+  }, [
+    appHeader,
+    blogSection,
+    appFooter
+  ]);
```
bnjbvr: I'd even use a small wrapper that concats the `appHeader`, given `section` and `appFooter`, since it seems to be a frequent use case.

jankeromnes:
Thanks a lot for the very fast and very useful review! 😀

> Do you have some kind of linting of the code? Errors like undefined variables or const variables being reassigned are better caught by these tools.

Good point! I'd love to use https://github.com/feross/standard at some point. The style is already pretty close (with the notable exception of semi-colons everywhere) and getting closer, but I think this should be a collective decision, so I'll propose this soon.

app.js
jankeromnes: I brought it down by swapping the if/else blocks, but you're right, it doesn't help and I'll remove it.

app.js
jankeromnes: Actually, the link to `/login` is only visible in the header when there is no `user` signed in, so this login handler's the general case is when there is no `user`.

However, I agree that it's the opposite of every other handler in this file, so it's probably less confusing to invert them anyway.

lib/routes.js
jankeromnes: I'll add a `FIXME` comment for this, but will keep depending on the CWD for now, because `camp` currently resolves template paths relative to where its `camp.js` module lives (usually in `node_modules/camp/lib/`), which is confusing.

See https://github.com/espadrine/sc/issues/59 .

lib/routes.js
jankeromnes: I agree this is not very clean.

A bug currently happens when your first connection is a WebSockets connection (e.g. all of the noVNC page was cached by the browser, then your token got invalidated, and you try to resume noVNC). To authenticate the connection, Janitor tries to redirect the raw Socket into an OAuth flow, which breaks because Sockets can't be redirected. Maybe we should just drop the connection in that case, and let users refresh the page?

jankeromnes: I'll change this commit and punt it to the next pull request, no need to delay this one.

lib/routes.js
jankeromnes: Thanks for the suggestion! But I believe that abstracting out a small wrapper wouldn't bring enough value compared to these slightly repetitive but simple, transparent and immediately hackable template arrays, and I'd generally prefer duplication over [risking the wrong abstraction](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction?duplication).

espadrine:
Looks rad.

Issue #31, opened 2016-10-02
Put The Janitor in The Janitor


notriddle:
* I've done developing with your own developer tools, and if nothing else it gives you rapid feedback on all the problems. You will fix all the annoying crap just to be able to not deal with it yourself.

* The most painful part of setting up dev environments for webapps like this is dealing with the external services like GitHub that you can't just host yourself. If you can automate all the setup work needed for a developer instance of an app that runs on GitHub, I'd be very interested in getting [aelita](https://github.com/AelitaBot/aelita) running on janitor.

* I want to be able to run the Janitor in the Janitor so I can code while I code.

jankeromnes:
Great idea! Thanks, added to #24. I remember @sylvestre and @Coder206 also requested this. I used to think it didn't make sense, because the Janitor web app was fairly simple (so better suited for vanilla Cloud9 IDE development, rather than full fledged Docker automation on Janitor), but it is becoming more complex now as it grows to multiple servers, plus setting it up for local development already takes several non-trivial steps, and I definitely agree about the dogfooding benefits in general.

jankeromnes:
This is now live! 🎉🎆🍾

PR #48, opened 2017-03-16
Allow forcing insecure HTTP for local development only

jankeromnes:
This pull request introduces two `db.json` security parameters:

```json
{
  "security": {
    "forceHttp": true,
    "forceInsecure": true
  }
}
```

They can be used by Janitor developers to completely disable HTTPS support, Let's Encrypt certificate generation, and several security policies (like hostname verification and some HTTP headers).

Additionally, forcing "insecure mode" adds an obvious warning message to all web pages, so that any accident on a production app becomes immediately visible to all users.

It also prints all sign-in keys in the server logs, because signing in via email is generally not an option when developing locally.

Note: I also sneaked in the unrelated `http.ServerResponse` bug fix again, respecting your nits from the previous pull request. Sorry for making the diff view confusing! (It might help to review comments separately).

### DIFF
--- a/app.js
+++ b/app.js
@@ -21,24 +21,26 @@ boot.executeInParallel([
 ], () => {
 
   // You can customize these values in './db.json'.
-  let hostname = db.get('hostname', 'localhost');
-  let https = db.get('https');
-  let ports = db.get('ports');
+  const hostname = db.get('hostname', 'localhost');
+  const https = db.get('https');
+  const ports = db.get('ports');
+  const security = db.get('security');
 
   // The main Janitor server.
-  let app = camp.start({
+  const app = camp.start({
     documentRoot: process.cwd() + '/static',
     port: ports.https,
-    secure: true,
+    secure: !security.forceHttp,
     key: https.key,
     cert: https.crt,
     ca: https.ca
   });
 
-  log('Janitor → https://' + hostname + ':' + ports.https);
+  log('[ok] Janitor → http' + (security.forceHttp ? '' : 's') + '://' +
+    hostname + ':' + ports.https);
 
   // Protect the server and its users with a security policies middleware.
-  app.handle((request, response, next) => {
+  const enforceSecurityPolicies = (request, response, next) => {
     // Only accept requests addressed to our hostname, no CDN here.
     if (request.headers.host !== hostname) {
       log('dropping request for', request.headers.host);
@@ -50,14 +52,20 @@ boot.executeInParallel([
     // Tell browsers to only use secure HTTPS connections for this web app.
     response.setHeader('Strict-Transport-Security', 'max-age=31536000');
 
-    // Prevent browsers from accidentally detecting scripts where they shouldn't.
+    // Prevent browsers from accidentally seeing scripts where they shouldn't.
     response.setHeader('X-Content-Type-Options', 'nosniff');
 
     // Tell browsers this web app should never be embedded into an iframe.
     response.setHeader('X-Frame-Options', 'DENY');
 
     next();
-  });
+  };
+
+  if (!security.forceInsecure) {
+    app.handle(enforceSecurityPolicies);
+  } else {
+    log('[warning] disabled all https security policies');
+  }
 
   // Authenticate signed-in user requests with a server middleware.
   app.handle((request, response, next) => {
--- a/join.js
+++ b/join.js
@@ -2,6 +2,7 @@
 // The following code is covered by the AGPL-3.0 license.
 
 const camp = require('@jankeromnes/camp');
+const http = require('http');
 const nodepath = require('path');
 
 const boot = require('./lib/boot');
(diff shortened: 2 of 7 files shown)


jankeromnes:
@bnjbvr Could you please have a look at your convenience? :) Thanks!

jankeromnes:
Note: Combined with https://github.com/JanitorTechnology/dockerfiles/pull/15, this will allow us to [Put The Janitor in The Janitor](https://github.com/JanitorTechnology/janitor/issues/31) 🎉🎈🍕🍾

bnjbvr:
A few questions:

- are there useful modes where `forceHttp !== forceInsecure`?
- could the websocket redirection happen by a special message sent over the websocket, indicating that the client code must redirect through `window.location`?

jankeromnes:
Thanks a lot for having a look!

> are there useful modes where `forceHttp !== forceInsecure`? If yes, can you explain them? I have issues identifying what should be guarded by each.

- `forceHttp` is just there to make the server serve HTTP instead of HTTPS (and disable Let's Encrypt certificate generation). This can be useful on its own for example with a reverse proxy that handles caching and renewing HTTPS certificates
- `forceInsecure` basically disables all other security features, e.g. requested hostname verification, security-related HTTP headers like HSTS, CSP, iframe blocking, secure cookies, and goes deeper into security hacks by printing email-login keys in the server logs etc. It should only be used for development (which is why a visible warning is added at the top of each page), but you could use it by itself to hack on something like https://localhost (Let's Encrypt generation would fail, but we could make `forceInsecure && !forceHttp` generate self-signed certificates)

> could the websocket redirection happen by a special message sent over the websocket, indicating that the client code must redirect through `window.location`?

Good question! From what I understand, once the WebSocket handshake is complete and you upgrade to binary transfer, there is no way to push a special message (e.g. a script) to browsers and make them redirect, except if you add such a feature to your own custom protocol (but we're proxying noVNC connections, and I don't think they implement such a redirect feature).

However, you could in theory reply with a "302 Found" HTTP status *during* the WebSocket handshake, but it seems that browsers are allowed to ignore these (see https://github.com/websockets/ws/issues/812) and I believe there would be worrying security implications if browsers accepted redirections.

In practice, dropping the connection will just show a "Connection failed" error in the cached noVNC interface, and a simple page refresh will allow the first request (which is not a WebSocket upgrade) to be hijacked for a transparent OAuth2 redirection and round-trip. ("Connection failed" error, the user refreshes, now it works again).

jankeromnes:
Note: Rebased!

app.js
```diff
@@ -21,24 +21,26 @@ boot.executeInParallel([
...
+  const ports = db.get('ports');
+  const security = db.get('security');
 
   // The main Janitor server.
-  let app = camp.start({
+  const app = camp.start({
     documentRoot: process.cwd() + '/static',
     port: ports.https,
-    secure: true,
+    secure: security.forceHttp !== true,
```
bnjbvr: Just an opinion, but checking only for `!security.forceHttp` would allow 0 and 1 as possible values, which could be intuitively used in this context. (There are a few other places where it is assumed that `security.forceHttp` is either `true` or anything else; if you want to enforce a certain subset of possible values, maybe worth asserting these somewhere and documenting this)

lib/sessions.js
```diff
@@ -47,7 +48,7 @@ exports.get = function (request, callback) {
       if (request.cookies) {
         request.cookies.set('token', token, {
           expires: new Date('2038-01-19T03:14:07Z'),
-          secure: true
+          secure: security.forceInsecure !== true
```
bnjbvr: You can also hoist the value of `secure` in the global context, since if I understand correctly, `forceInsecure` can't be switched at runtime.

lib/sessions.js
```diff
@@ -63,7 +64,7 @@ exports.destroy = function (request, callback) {
       // Destroy the cookie.
       request.cookies.set('token', '', {
         overwrite: true,
-        secure: true
+        secure: security.forceInsecure !== true
```
bnjbvr: (see above remark)

lib/users.js
```diff
@@ -110,17 +111,26 @@ exports.sendLoginEmail = function (email, request, callback) {
         return 'Janitor Sign-in link';
       },
       htmlMessage (key) {
+        const url = (security.forceHttp === true ? 'http' : 'https') + '://' +
```
bnjbvr: Here the prefix could be hoisted as well.

join.js
```diff
@@ -164,6 +165,13 @@ function ensureOAuth2Access (request, response, next) {
     return;
   }
 
+  // We can only use `http.ServerResponse`s to initiate OAuth2 authentication,
+  // not raw `net.Socket`s (as in WebSocket connections).
+  if (!(response instanceof http.ServerResponse)) {
```
bnjbvr: Maybe log something on the console? Then you'll know if it's a recurrent issue and it means more attention.

jankeromnes:
💯👍🎆 Yay! Thanks a lot for yet another great review @bnjbvr 😄 Your remarks are very helpful, and I'll try to reply to and act on them shortly.

app.js
jankeromnes: I thought it would be safer to only accept `security.forceInsecure === true` as the exact value that disables important security features, but maybe it isn't much more secure. I'll make it so that both forcing variables can be truthy/falsy regardless of the exact value.

jankeromnes:
Note: Rebased + all nits addressed!

join.js
jankeromnes: Done.

PR #51, opened 2017-03-30
Add a new Settings tab to control third-party service integrations

jankeromnes:
The first commit lays the foundation for this change by tweaking some CSS rules and classes here and there (mostly to prevent Settings tabs and other pages to become extremely wide horizontally on large screens).

The second commit changes the Settings page templates (and associated route in `lib/routes.js`) in order to create a tabbed menu with two tabs: "Account" (what originally existed in Settings, i.e. email and Cloud9 account info), and "Integrations" (a new tab where users will be able to "link" their external accounts to enable advanced features and automation).

The new Settings tab should look like this:

settings-integrations


jankeromnes:
@Coder206 If you have time to spare one of these days, and you're still interested in reviewing this, could you please have a look at this pull request? If you do, please double-check that I didn't make any obvious mistakes, and if you have any thoughts/feedback/questions about my general approach or a particular piece of code, please share them. I'll be happy to reply 😄

There is absolutely no rush, so please feel free to take your time before replying (even a week or so). Also, if you don't think you'll have the time soon, or you're not interested anymore, that's ok too: in that case please reset the "Reviewers" flag on this PR, and someone else will do the review.

Coder206:
@jankeromnes After pulling your code from `jankeromnes:integrations`, I tried running your code and I received an error. 

``
module.js:472
    throw err;
    ^

Error: Cannot find module '@jankeromnes/camp'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/ubuntu/workspace/app.js:4:14)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
``

My version of `node` is `7.8.0` and I commented out both https://github.com/JanitorTechnology/janitor/blob/master/app.js#L19 and https://github.com/JanitorTechnology/janitor/blob/master/app.js#L32. Is there something I am missing from the list of hacks?

lib/routes.js
```diff
@@ -153,10 +153,11 @@ exports.contributionsPage = function (response, user = null) {
 };
 
 // User settings page.
+const settingsHeader = camp.template('./templates/settings-header.html');
 const settingsSections = {
-  account: camp.template('./templates/settings-account.html')
+  account: camp.template('./templates/settings-account.html'),
+  integrations: camp.template('./templates/settings-integrations.html')
 };
-const settingsHint = camp.template('./templates/settings-hint.html');
```
jankeromnes: @Coder206 I don't want to display this hint in every Settings section, just in Account, so I moved this template to the bottom of `settings-account.html`.

Coder206: @jankeromnes Ah, that makes sense. Thanks for the clarification :)

jankeromnes:
> I had some time today to review your code.

Thanks a lot for taking the time!

> some of the changes seemed to me to be unrelated to the settings tab (the text changes in landing.html for example)

Yes sorry, I included two commits in this Pull Request: https://github.com/JanitorTechnology/janitor/pull/51/commits/94318e1394b4f6d264f9584d7fbb313c2f9c12c1 ("Slightly improve the visual appearance of most web app pages") and https://github.com/JanitorTechnology/janitor/pull/51/commits/a712487f802760e1f5f30b452130a897c34dd054 ("Add a user settings tab to control third-party service integrations"), which you can review separately (the default pull request preview shows the combined changes from all commits, which is confusing).

> If it's ok with you, do you mind if I try the changes out in C9 before another review ?

Sure! That's a great idea, and I can help you troubleshoot any runtime problems so that you can preview the changes from this pull request.

> `Error: Cannot find module '@jankeromnes/camp'`

This means that you don't have our ScoutCamp fork in your local `./node_modules/` folder. You can fix that by running `npm install` or `npm update`, which will install/update any npm dependencies listed in the `./package.json` file (which asks for a specific revision of the `'@jankeromnes/camp'` module).

lib/routes.js
jankeromnes: You're welcome! :)

Coder206:
@jankeromnes I had the chance to look over the code and run it live in C9 and it was working as I would expect it for the most part. However there seemed to be no projects listed anymore:

page-shot-2017-4-3 projects the janitor

I would think that it is caused by the fact the projects are not on my C9 local environment. Does that make sense?

Unfortunately, I was not able to test on a real device but I was able to mock a phone and the result is great until inferior to 340px (probably not a big concern):

page-shot-2017-4-3 settings the janitor 1

Finally, after taking a closer look to the `Integrations` tab I noticed you are providing the website address as the name of the integration. What are your thoughts on using the project names (without the domain names), like this:

page-shot-2017-4-4 settings the janitor

jankeromnes:
Thanks a lot for taking the time to make such a great review! Much appreciated! :)

> However there seemed to be no projects listed anymore:
> I would think that it is caused by the fact the projects are not on my C9 local environment. Does that make sense?

Yes indeed, the list projects that your Janitor instance supports comes from your `db.json`, where by default the `"projects"` entry is empty (`{}`).

Note: There is a very basic UI in the Admin section to add new projects (at the bottom of `/admin/projects/`), by indicating a project ID (e.g. "servo"), a project name (e.g. "Servo"), and a project icon URL (e.g. "/img/servo.svg"). Once you've added a project, you can edit its various Dockerfiles to set up its Docker image and then keep it up-to-date, but that interface is currently very confusing and not intuitive. In the future, I would like to add the concept of "Project Owners" (or "Project Admins"), and add more intuitive UI (e.g. in Settings) to let you manage your own Janitor Hosts and Projects directly from your Janitor account (without needing the secret Admin section).

> Unfortunately, I was not able to test on a real device but I was able to mock a phone and the result is great until inferior to 340px (probably not a big concern):

Oh, indeed, thanks a lot for thinking of that! I will have to make sure that the interface looks and works great on mobile as well (Janitor should eventually work as a Progressive Web App on your phone, and there shouldn't be any annoying UI bugs such as the one you found). Maybe we'll decide that 340px is too narrow to use Janitor, but I'd really like Janitor to support as many screen sizes as possible.

> Finally, after taking a closer look to the `Integrations` tab I noticed you are providing the website address as the name of the integration. What are your thoughts on using the project names (without the domain names), like this:

Thanks for suggesting this variant! I really like the simplicity of showing the name of the integration instead of its URL.

The reason I went with URLs is because maybe we'll eventually support more services, and maybe there will be ambiguous names (e.g. other projects have a "Bugzilla", so at some point it might be important to specify that we're talking about https://bugzilla.mozilla.org and not https://bugzilla.kernel.org, https://bugs.kde.org/, https://bugzilla.redhat.com/ or some such).

However, this is very forward looking, and maybe for now it's simpler to show more readable service names. I'll switch to using your suggestion, and later we might switch back to URLs if we do encounter ambiguous integration names.

Coder206:
@jankeromnes I understand! Perhaps in the case of URL for bugzilla it could say Kernel Bugzilla?

Coder206:
@jankeromnes I was thinking of another integration, a little more project specific, connecting to discourses and forums.

jankeromnes:
> Perhaps in the case of URL for bugzilla it could say Kernel Bugzilla?

Agreed! We could indeed make sure to use a name that users of this service will recognize. For example, Firefox developers usually refer to their Bugzilla as "BMO" (from the initials in the bugzilla.mozilla.org URL).

> approved for the functionality of this PR

Thanks a lot! I'll implement your suggestions and then merge this pull request.

> I was thinking of another integration, a little more project specific, connecting to discourses and forums

Interesting suggestion! Do you have specific ideas of what sort of useful interaction there could be between Janitor and project-specific Forums? Maybe an IDE button that allows to get help by posting a question on the project's Forum?

Coder206:
@jankeromnes Sounds good! Yes, I was thinking of bringing the forums of projects into C9 while programming to consult when encountering issues and such. In my vision, this would unify a project resources in a way that is more convenient as sorting through documentation for methods of communication to a project to me. (and perhaps others)

jankeromnes:
Indeed, it's a good idea! But I think we need to make sure that such integrations are simple to implement, and stay up-to-date, because projects usually have a large number of various resources, which sometimes get created, deprecated, migrated, split, etc.

I think that adding an integration that allows to ask a simple question about a project is great, but maybe striving to deeply integrate all the docs and forums of every project will be too much work and get deprecated too quickly.

jankeromnes:
After doing some more checks on mobile, I decided to:
- [x] Implement your suggestion to use service names (e.g. "Bugzilla") instead of URL (e.g. "bugzilla.mozilla.org"), because it's both clearer and shorter,
- [x] Hide the words "Go to" and "Coming" from the service buttons on small screens (just like the "Edit in" words on the landing page's buttons).

PR #57, opened 2017-04-14
Change the port to appropriate number when using HTTP and HTTPS for the console

Coder206:
If you could take a look at the modification and test it for an HTTPS state that would be greatly appreciated. Unless this PR is breaking a feature ;)

**Problem**
I was running Janitor with `node app` and it says `http://localhost:8080` in the console. Port 8080 is the HTTPS port according to [db.json](https://github.com/JanitorTechnology/dockerfiles/blob/450ceb281b7ea460ec931edc9a3eaf0d4a18206f/janitor/db.json#L5). This would make it more challenging to simply click the link to navigate to the page.

**Solution**
I cloned the concept changing the link from `https://localhost:8080` to `http://localhost:8080` with regards to the port number. The expected result now is `https://localhost:8080` for HTTPS and `http://localhost:8081` for HTTP.

### DIFF
--- a/app.js
+++ b/app.js
@@ -37,7 +37,7 @@ boot.executeInParallel([
   });
 
   log('[ok] Janitor → http' + (security.forceHttp ? '' : 's') + '://' +
-    hostname + ':' + ports.https);
+    hostname + ':' + (security.forceHttp ? ports.http : ports.https));
 
   // Protect the server and its users with a security policies middleware.
   const enforceSecurityPolicies = (request, response, next) => {


Coder206:
@jankeromnes After more tests in a non Janitor in Janitor container it appears that this PR is detrimental to the project as it prevents the user to get to the page.

jankeromnes:
Hi @Coder206, thanks a lot for reporting this problem, and even submitting a pull request to try and fix it!

The exact purpose of the hostname and port names in `db.json` can be a little hard to guess, so I'll share a few more details on what they're for:

# `hostname`

The Janitor service expects to run on a given `hostname` (like "janitor.technology", or "moz1.janitor.technology"), and will embed the `hostname` string in many different URLs (e.g. the `[ok] Janitor → https://janitor.technology:1443` URL printed in the logs, the sign-in links sent via email, Cloud9 URL parameters to point the IDE to the right host and port of a container, etc.)

# `ports.http` & `ports.https`

The `ports.http` and `ports.https` ports could in theory be set to `80` and `443` respectively, allowing you do use the Janitor service via `http://hostname` and `https://hostame` directly without specifying non-standard ports, but running servers on ports below `1024` requires root privileges.

To work around this, we usually set them to higher values, e.g. `1080` and `1443`, or `8081` and `8080`, but since the service expects to work on standard `http://hostname` and `https://hostname` URLs, we need a separate mechanism to pipe the higher ports to `80` and `443`. This can be done with an `iptables` redirection (`make ports` can set this up for you), or by installing a reverse proxy like `nginx` to pipe all traffic (it is considered a good practice to run web services behind a reverse proxy in production, and can be used to perform additional optimizations and security checks on incoming and outgoing traffic).

`ports.http` is used by an HTTP server that simply [redirects most requests](https://github.com/JanitorTechnology/janitor/blob/bdb9fa73985144acd727de6ccdf419c96dd5d6d3/lib/boot.js#L52-L55) to an HTTPS URL (without a non-standard port). (An exception is made for Let's Encrypt [HTTP challenge tokens](https://github.com/JanitorTechnology/janitor/blob/bdb9fa73985144acd727de6ccdf419c96dd5d6d3/lib/boot.js#L43-L50), which need to be served over HTTP using a secret URL.)

`ports.https` is actually a bit of a misnomer, because it is used by the main Janitor web service (HTTPS by default), but setting `security.forceHttp` to `true` causes Janitor to serve HTTP content on `ports.https` (for example, you could be serving unencrypted HTTP on port 1443, but have an `nginx` reverse proxy encapsulate this traffic into HTTPS on port 443). Maybe renaming this port `ports.janitor` would be less confusing.

jankeromnes:
So more specifically for the problem you saw, if running `node app` tells you `[ok] Janitor → http://localhost:8080`, it means the server is actually serving HTTP content on port 8080, and clicking the link on your local computer should work directly.

However, if you're running this in Janitor or in Cloud9, you don't have an easy way to access the `http://localhost:8080` of your container (instead, this will be exposed via a proxy, on URLs like `https://moz1.janitor.technology/<containerID>/8080/` or `https://project-name.c9users.io/`, respectively).

You could set `hostname` in `db.json` to either `moz1.janitor.technology` (when in Janitor) or `project-name.c9users.io` (when in Cloud9). However, this will print a URL like `http://moz1.janitor.technology:8080` in the logs, which still won't work. In order to make these URLs easier to find, I think we should implement a button in https://janitor.technology/contributions/ to "open port 8080 of your container" and/or fix the "Preview Running Application" link in Cloud9 IDE to point to the right URL, e.g. with a Cloud9 plugin.

jankeromnes:
We should definitely address the two points from the previous comment, but I don't think we'll want to go forward with this PR's suggested change, so I'll close it for now. Thanks a lot for raising these important issues!

Issue #29, opened 2016-09-13
User Configurations Wishlist


jankeromnes:
Today, for every new contribution on the Janitor, there are a number of repetitive steps you need to perform in order to configure your environment properly. It would be great to automate these steps, e.g. by letting users add their desired configurations once and for all to their Janitor account, and automatically apply them to all your contributions.

Here are the configurations we'll aim to support:

- [x] Cloud9 username (to link your Cloud9 and Janitor accounts)
- [x] Cloud9 SSH public key (to authorize Cloud9 IDE to access to your contributions)
- [ ] Personal SSH public keys (to authorize SSH access to your contributions from anywhere)
- [ ] Dedicated SSH key pair (to allow pushing from your contribution environment e.g. to GitHub)
- [ ] Git/Mercurial preference (to prefer one over the other for contributions where both are supported)
- [ ] `.gitconfig` (to preconfigure e.g. `user.name`, `user.email`, `core.editor` and credentials like `bz.username` and `bz.apikey`)
- [ ] `.gitignore` (to make sure contribution internals are ignored, e.g. the `.c9/` folder in your workspace)
- [ ] `.hgrc` (to preconfigure `ui.username`, enable extensions like `mq`, specify the desired patch format)
- [ ] `.gdbinit` (e.g. to ignore `SIG38` and `SIG39` by default)
- [ ] `.vimrc`

If you feel that something important for your work is missing, please comment below and I'll update the list.

jankeromnes:
@beaufortfrancois @Coder206 @mantaroh @notriddle @rcatolino Please take a look: would this list cover all your configuration needs for your Janitor workflows?

Coder206:
@jankeromnes Based on my workflow and knowledge it looks like it would work well for Servo!

rcatolino:
A custom bashrc would be nice (for aliases, completion behavior etc.), and you could set your prompt there.

beaufortfrancois:
👍

notriddle:
This would cover all the *needs*, yeah, but could we add [the hub CLI app](https://hub.github.com/)? The hub command allows you to manage pull requests and stuff directly from the CLI.

If you did set up hub, since it requires you to get a personal access token anyway, you might as well use that for git, too, and skip the SSH stuff for GitHub users.

jankeromnes:
Very interesting, I didn't know about hub, thanks for the suggestion! I will look into adding it to all project images by default.

jankeromnes:
[...]

PR #58, opened 2017-04-14
Ignore the  folder when making git commands

Coder206:
@jankeromnes I was working on submitting the last PR when I encountered files from the `.c9/` added to my list of watched files. I heard that `.gitignore` can prevent files and folders from being pushed upstream so I made an addition to the `.gitignore` file. Let me know what you think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.c9/*
 .DS_Store
 Thumbs.db
 npm-debug.log
@@ -11,4 +12,3 @@ janitor.log
 janitor.pid
 backups
 tokens
-.c9/*
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-.c9/*
+.c9
 .DS_Store
 Thumbs.db
 npm-debug.log
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ janitor.log
 janitor.pid
 backups
 tokens
+.c9/*


.gitignore
```diff
@@ -11,3 +11,4 @@ janitor.log
 janitor.pid
 backups
 tokens
+.c9/*
```
jankeromnes: Nit: This `.gitignore` file lists generic development files & folders at the top, and Janitor-specific files at the bottom. Since `.c9` is not a Janitor-specific folder, please move it to the top of the file (e.g. to the first line).

jankeromnes:
Thanks a lot for this contribution! Several people have been complaining about the `.c9/` folder showing up in `git status`, because it's confusing and slightly annoying.

The next step will be to prevent it from showing up in any project on Janitor, by including it in a global `/home/user/.gitignore` file installed in every user container (this feature/file is also requested by #29).

.gitignore
jankeromnes: Nit: You can also omit the `/*` part, as ignoring `.c9` will also ignore everything it contains.

.gitignore
```diff
@@ -1,3 +1,4 @@
+.c9
```
jankeromnes: That's perfect, thanks a lot!

Coder206:
My pleasure!

jankeromnes:
Approved! Thanks a lot for this valuable contribution, and congratulations on your first commit landed in Janitor! 💯🎉

PR #61, opened 2017-04-24
Implement a server route to drop invalid requests and optionally report them.

jankeromnes:


jankeromnes:
The problem this pull request is trying to address is that sometimes, the server receives several hundred invalid requests at approximately the same time, which spams server logs with several hundred lines like:

    [2017-04-20T16:53:44.116Z] dropping request for 128.93.132.18
    [2017-04-20T16:53:46.272Z] dropping request for 128.93.132.18
    [2017-04-20T16:53:59.596Z] dropping request for 128.93.132.18
    [2017-04-20T16:54:14.495Z] dropping request for 128.93.132.18
    …

The goal is to unspam logs by summarizing these sessions with a single line like:

    [2017-04-17T19:35:58.112Z] dropped 350 requests for invalid hostname: 128.93.132.18

jankeromnes:
@bnjbvr This commit has been lying around my workspace for a long time, but should be a quick review. Could you please have a look whenever is convenient for you?

Not very important, but it would be good to merge this one day.

lib/routes.js
```diff
@@ -22,6 +23,29 @@ exports.redirect = function (response, url, permanently = false) {
...
+  response.end();
+
+  if (!reason) {
+    return;
+  }
+
+  // Count all attempts dropped for the same reason within a short time window.
+  let { count = 0, timeout = null } = consecutiveAttempts[reason] || {};
+  count++;
+  clearTimeout(timeout);
```
bnjbvr: `clearTimeout(null)` will just do nothing, I assume?

lib/routes.js
```diff
@@ -22,6 +23,29 @@ exports.redirect = function (response, url, permanently = false) {
...
+  timeout = setTimeout(() => {
+    log('[warning] dropped', count, 'request' + (count === 1 ? '' : 's'), 'for',
+      reason);
+    delete consecutiveAttempts[reason];
+  }, 1500);
```
bnjbvr: nit: no magic numbers :P

lib/routes.js
jankeromnes: Correct, it doesn't do anything. (I rewrote this entire piece of code many times, because I wanted something short yet readable, and I believe this is a good trade off.)

lib/routes.js
jankeromnes: Would you advise having a constant, or a configurable parameter? I don't think there is much value in the latter, but I can at least do the former for good measure.

jankeromnes:
> LGTM

Thanks again for an amazingly quick review! 👍

lib/routes.js
bnjbvr: Yeah just a constant at top of the file, in uppercase casing, would be nicer 👍

PR #62, opened 2017-04-24
Make the web proxy re-send any previously consumed data chunks

### LINKED ISSUE #54: Some HTTP methods don't get properly redirected in the service proxy
When using a service that's hosted in a container, e.g. a Web service, then the user can access the Web service, and all the requests get proxied through Janitor (which does a simple redirect). However, some methods aren't getting properly redirected: I have seen instances where `POST`, `PUT`, `DELETE` methods wouldn't ever reach the web service in the Docker instance.

jankeromnes:
This uses `saveRequestChunks`, a new ScoutCamp feature added in https://github.com/espadrine/sc/pull/60, and partially fixes #54.

### DIFF
--- a/app.js
+++ b/app.js
@@ -29,6 +29,7 @@ boot.executeInParallel([
   // The main Janitor server.
   const app = camp.start({
     documentRoot: process.cwd() + '/static',
+    saveRequestChunks: true,
     port: ports.https,
     secure: !security.forceHttp,
     key: https.key,
--- a/join.js
+++ b/join.js
@@ -39,6 +39,7 @@ boot.executeInParallel([
     // Start an authenticated Janitor proxy for Docker containers on this host.
     const proxy = camp.start({
       documentRoot: process.cwd() + '/static',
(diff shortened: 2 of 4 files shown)


jankeromnes:
Ben, I think this very small pull request will make you happy. 😄

jankeromnes:
>  I assume this is the new scoutcamp API for saving chunks in use here.

Yes indeed! This feature has not landed yet, but I updated our ScoutCamp fork to include https://github.com/espadrine/sc/pull/60/commits/8f8204aa850533a3f3bd621339e4849b4153fce1 so we can already use it.

> Does it work with Kresus?

It should at least fix part of the problem, but I didn't remember how to test Kresus. I'll do that now.

jankeromnes:
Rebased, and now using https://github.com/espadrine/sc/pull/60/commits/34a8855d4ceeb9c568e7571b30dea8f5717f1033 (which renames `request.chunks` to a more explicit `request.savedChunks`).

Issue #63, opened 2017-05-07
VNC won't open


notriddle:
I deleted the container, added it back, and it still won't open.

Coder206:
@notriddle Have you received an error message on your end?

notriddle:
No error message. It just hangs.

notriddle:
The project is Servo

Coder206:
@notriddle Like this:
![page-shot-2017-5-6 novnc](https://cloud.githubusercontent.com/assets/7758155/25776990/f0b1819c-329c-11e7-9cfd-a89d3e2e8721.png)?

notriddle:
Like that

Coder206:
Curiously, the one above is a new container created tonight. I am still able to open an older container of Servo in VNC.

Coder206:
I am receiving an error when navigating to https://moz1.janitor.technology/****************/8088/vnc.html

jankeromnes:
Thanks a lot for opening this issue! I will have a look in the backend to see what's happening here.

jankeromnes:
A few general notes:

- There is no need to hide container IDs from URLs. Janitor's OAuth2 proxy verifies container ownership (if you're not signed in to Janitor as the rightful owner of a container, accessing another user's URL will simply lead to a 404 error).
- Simply refreshing the noVNC page helps with many different issues (connection problems are often due to discrepancies between the noVNC resources loaded from cache and the WebSocket connection actually being established).

My current theory to explain this particular problem is that the latest Firefox, Servo, Thunderbird and Janitor images are based on a broken noVNC revision.

- The latest image update was on May 3, around 16:00 UTC.
- It looks like noVNC shows a similar quota-exceeded bug (in the JS Console)* on a Firefox image update from April 29, ~13:00 UTC, but not on an older update from March 23, ~10:00 UTC.

I didn't find a storage-quota-related bug in the [noVNC repository](https://github.com/novnc/noVNC/issues), but will trigger a new image update based on today's latest noVNC revision, and see if the problem is solved. If it's not, I'll file a noVNC bug, and maybe try to base another round of image updates on a last-known-good-revision of noVNC.

* `Failed to execute 'setItem' on 'Storage': Setting the value of 'https://moz1.janitor.technology/482bb14c21f81c5a/8088/<anon1>!raw' exceeded the quota.`

jankeromnes:
[...]

PR #65, opened 2017-05-16
Update the dockerfiles submodule and fix URLs for older and newer (production-ready) noVNC builds

jankeromnes:


jankeromnes:
Minor changes to fix #63, merging without review.

PR #66, opened 2017-05-17
Add an automatically generated API reference page to the website.

jankeromnes:
This simple change just adds a small template that includes `api.toHTML()`, which is a basic HTML export of Janitor's live [SelfAPI](https://github.com/JanitorTechnology/selfapi) instance. I also added a few basic CSS rules to make these docs more readable.

The resulting page looks like this:

janitor-api

Edit: I've also added another commit with a few minor fixes (coding style, API description).

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -81,6 +81,24 @@ exports.landingPage = function (response, user = null) {
   ]);
 };
 
+// Public API reference page.
+const apiSection = camp.template('./templates/reference-api.html');
+exports.apiPage = function (response, api, user = null) {
+  const title = 'API';
+  const htmlReference = api.toHTML();
+
+  response.template({
+    htmlReference,
+    title,
+    user,
+    scripts: []
+  }, [
+    appHeader,
+    apiSection,
+    appFooter
+  ]);
+};
+
 // Public blog page.
 const blogSection = camp.template('./templates/blog.html');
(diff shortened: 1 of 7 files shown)


jankeromnes:
@bnjbvr Could you please have a look at your convenience?

lib/routes.js
```diff
@@ -81,6 +81,24 @@ exports.landingPage = function (response, user = null) {
   ]);
 };
 
+// Public API reference page.
+const apiSection = camp.template('./templates/reference-api.html');
+exports.apiPage = function (response, api, user = null) {
```
bnjbvr: The only call site passes an explicit `user` value, should the default value be removed here?

lib/routes.js
```diff
@@ -81,6 +81,24 @@ exports.landingPage = function (response, user = null) {
...
+  const title = 'API';
+  const htmlReference = api.toHTML();
```
bnjbvr: Very nice!

jankeromnes:
> APPROVED! Nice patch, thanks.

Thank you! :D

lib/routes.js
jankeromnes: You're right, the default value doesn't do anything, and probably won't. I added the `= null` just to give programmers a hint that `user` can be `null` (when using the website without signing in), but I can remove it if you feel strongly about this.

PR #68, opened 2017-05-18
Add an API to manage third-party OAuth2 providers.

jankeromnes:
This pull requests adds two API endpoints (that will be used in a follow-up pull request to support authentication via GitHub):
- List all OAuth2 provider configurations (their URLs and client credentials)
- Update an OAuth2 provider (accepting a JSON body formatted as an RFC 6902 JSON Patch array)

They should only be usable by website administrators (note the `users.isAdmin(user)` checks), and should fail otherwise.

The SelfAPI auto-generated documentation looks like this:
admin-api-docs

### DIFF
--- a/api/admin-api.js
+++ b/api/admin-api.js
@@ -1,0 +1,107 @@
+// Copyright © 2017 Jan Keromnes. All rights reserved.
+// The following code is covered by the AGPL-3.0 license.
+
+const jsonpatch = require('fast-json-patch');
+const selfapi = require('selfapi');
+
+const db = require('../lib/db');
+const log = require('../lib/log');
+const users = require('../lib/users');
+
+// API resource to manage the Janitor instance itself.
+const adminAPI = module.exports = selfapi({
+  title: 'Admin'
+});
+
+// API sub-resource to manage OAuth2 providers.
+const oauth2providersAPI = adminAPI.api('/oauth2providers');
+
+oauth2providersAPI.get({
+  title: 'List OAuth2 providers',
+
+  handler: (request, response) => {
+    const { user } = request;
+    if (!user || !users.isAdmin(user)) {
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' }, null, 2);
+      return;
+    }
+
+    const providers = db.get('oauth2providers');
+    response.json(providers, null, 2);
+  },
+
+  examples: [{
+    response: {
+      body: JSON.stringify({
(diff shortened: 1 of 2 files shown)


jankeromnes:
@bnjbvr please have a look when you have a moment. :)

api/admin-api.js
```diff
@@ -0,0 +1,105 @@
...
+
+    let json = '';
+    request.on('data', chunk => {
+      json += String(chunk);
+    });
+    request.on('end', () => {
+      let operations = null;
+      try {
+        operations = JSON.parse(json);
+      } catch (error) {
```
bnjbvr: `error` is unused. Maybe log it for debugging?

jankeromnes:
Thanks a lot for this review in spite of you're extremely busy schedule! I really appreciate it 💯🥇

api/admin-api.js
```diff
@@ -0,0 +1,105 @@
...
+  handler: (request, response) => {
+    const { user } = request;
+    if (!user || !users.isAdmin(user)) {
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' }, null, 2);
+      return;
+    }
+
+    const providers = db.get('oauth2providers');
+    response.json(providers, null, 2);
```
jankeromnes: If not specified, responses should have a 200 status code.

api/admin-api.js
```diff
@@ -0,0 +1,105 @@
...
+const oauth2providersAPI = adminAPI.api('/oauth2providers');
+
+oauth2providersAPI.get({
+  title: 'List OAuth2 providers',
+
+  handler: (request, response) => {
+    const { user } = request;
+    if (!user || !users.isAdmin(user)) {
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' }, null, 2);
```
jankeromnes: I believe it's a good idea to make JSON output human readable by default, because it doesn't incur any significant cost (gzip will take care of the transfer size) and has significant advantages (readable by default without additional tooling, e.g. by just opening a `GET` URL in a browser tab, or by reading JSON responses in logs with mixed content, etc).

See also: https://github.com/espadrine/sc/issues/53 (I tried to make JSON replies human-readable by default in ScoutCamp, but failed to convince @espadrine).

api/admin-api.js
jankeromnes: Fair enough!

PR #69, opened 2017-06-01
Refactor Scout/Ajax forms to use `window.fetch()` with the new Janitor API

jankeromnes:
Currently, async actions triggered by buttons and forms on the Janitor website work approximately like this:

```html
<form id="my-form">
  <input data-submit-on="blur" name="key" value="value">
</form>

<script>
Scout('#my-form').on('submit', function (query) {
  query.action = 'ajax-action';
  query.data = getFormData(form);
  query.resp = onAjaxResponse;
});
</script>
```

```js
// Server-side (/app.js):
camp.ajax.on('ajax-action', onAjaxRequest);
```

We would like to replace `Scout` calls by `window.fetch()` on the website, and `camp.ajax.on()` handlers by the new [Janitor API](https://janitor.technology/reference/api) on the server (a REST/JSON web API implemented with [Self API](https://github.com/JanitorTechnology/selfapi)).

The new proposed way to implement async website actions (buttons and forms) works approximately as follows:

```html
<form id="my-form" action="/api/hosts/host.name/credentials" method="delete">
  <input type="submit" value="Reset host credentials">
</form>

<script>
var form = document.querySelector('#my-form');
  var data = getFormData(form);
  window.fetch(form.action, {
    method: 'DELETE',
    body: JSON.stringify(data, null, 2)
  })
  .then(onFetchResponse)
  .catch(onError);
});
</script>
```

```js
// Server-side (/api/hosts-api.js):
hostAPI.delete('/credentials', {
  title: 'Reset host credentials',
  handler: (request, response) => {}
});
```


jankeromnes:
@espadrine Hi! Since you're familiar with Scout, and since you suggested I switch to `window.fetch()` in order to use my new JSON API directly from the website, could you please take a look at this pull request?

I've already tested the new code and it works fine (for `DELETE` and `PATCH` calls), but I'd really like a second opinion on this, in case I'm doing something horribly wrong.

Thanks! 😄

espadrine:
lgtm!

You might want to include a polyfill for old browsers (incl. IE11). Although, if you are targeting evergreen, it's not necessary.

jankeromnes:
I squeezed in two commits to add `fetch` and `Promise` polyfills, and update the website's JS libraries.

jankeromnes:
As @nt1m pointed out, this change seems to break Firefox but not Chrome. Oh well.

PR #70, opened 2017-06-02
Allow renaming containers

bnjbvr:
@jankeromnes, can you take a look please? Cheers!

### DIFF
--- a/api/admin-api.js
+++ b/api/admin-api.js
@@ -92,18 +92,16 @@ oauth2providerAPI.patch({
       json += String(chunk);
     });
     request.on('end', () => {
-      let operations = null;
       try {
-        operations = JSON.parse(json);
+        const operations = JSON.parse(json);
+        jsonpatch.applyPatch(provider, operations, true);
       } catch (error) {
-        log('[fail] json patch', error);
+        log('[fail] patching oauth2 provider', error);
         response.statusCode = 400; // Bad Request
-        response.json({ error: 'Problems parsing JSON' }, null, 2);
+        response.json({ error: 'Invalid JSON Patch' }, null, 2);
         return;
       }
 
-      // Apply the requested changes to the provider.
-      jsonpatch.applyPatch(provider, operations);
       db.save();
       response.json(provider, null, 2);
     });
--- a/api/hosts-api.js
+++ b/api/hosts-api.js
@@ -1,6 +1,7 @@
 // Copyright © 2016 Jan Keromnes. All rights reserved.
 // The following code is covered by the AGPL-3.0 license.
 
+const jsonpatch = require('fast-json-patch');
 const selfapi = require('selfapi');
 
 const db = require('../lib/db');
@@ -336,13 +337,108 @@ hostAPI.get('/version', {
 });
 
 // API sub-resource to manage a single container on a cluster host.
-const containerAPI = hostAPI.api('/:container', {
-  title: 'Containers'
+const containerAPI = hostAPI.api('/:container', 'Containers');
+
+containerAPI.patch({
+  title: 'Update a container',
+  description: 'Update the properties of a given Docker container.',
+
+  handler: (request, response) => {
+    const { user } = request;
+    if (!user) {
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' }, null, 2);
+      return;
+    }
+
+    const { container } = request.query;
+    if (container.length < 16 || !/^[0-9a-f]+$/.test(container)) {
+      response.statusCode = 400; // Bad Request
+      response.json({ error: 'Invalid container ID' }, null, 2);
+      return;
+    }
+
+    const { hostname } = request.query;
+    const machine = machines.getMachineByContainer(user, hostname, container);
+    if (!machine) {
+      response.statusCode = 404;
+      response.json({ error: 'Container not found' }, null, 2);
+      return;
+    }
+
+    let json = '';
+    request.on('data', chunk => {
+      json += String(chunk);
+    });
+    request.on('end', () => {
+      try {
+        const operations = JSON.parse(json);
+        jsonpatch.applyPatch(machine.properties, operations, true);
+      } catch (error) {
+        log('[fail] patching machine.properties', error);
+        response.statusCode = 400; // Bad Request
+        response.json({ error: 'Invalid JSON Patch' }, null, 2);
+        return;
+      }
+
+      db.save();
+      response.json(machine.properties, null, 2);
+    });
+  },
+
+  examples: [{
+    request: {
+      urlParameters: {
+        hostname: 'example.com',
+        container: 'abcdef0123456789',
+      },
+      body: JSON.stringify([
+        { op: 'replace', path: '/name', value: 'Much container. Such wow.' }
+      ], null, 2)
+    },
+    response: {
+      body: JSON.stringify({ name: 'Much container. Such wow.' }, null, 2)
+    }
+  }, {
+    request: {
+      urlParameters: {
+        hostname: 'example.com',
+        container: 'abcdef0123456789',
+      },
+      body: 'This is not JSON Patch',
+    },
+    response: {
+      status: 400,
+      body: JSON.stringify({ error: 'Invalid JSON Patch' }, null, 2)
+    }
+  }, {
+    request: {
+      urlParameters: {
+        hostname: 'example.com',
+        container: 'invalid-container-id',
+      },
+    },
+    response: {
+      status: 400,
+      body: JSON.stringify({ error: 'Invalid container ID' }, null, 2)
+    }
+  }, {
+    request: {
+      urlParameters: {
+        hostname: 'example.com',
+        container: '0000000000000000',
+      },
+    },
+    response: {
+      status: 404,
+      body: JSON.stringify({ error: 'Container not found' }, null, 2)
+    }
+  }]
 });
 
 containerAPI.get('/:port', {
   title: 'Get a single container port',
-  description: 'Get information about a given Docker container port.',
+  description: 'Get the properties of a given Docker container port.',
 
   handler: (request, response) => {
     let { user, oauth2scope } = request;
--- a/api/user-api.js
+++ b/api/user-api.js
@@ -58,18 +58,17 @@ userAPI.patch({
       json += String(chunk);
     });
     request.on('end', () => {
-      let operations = null;
       try {
-        operations = JSON.parse(json);
+        const operations = JSON.parse(json);
+        jsonpatch.applyPatch(user.profile, operations, true);
       } catch (error) {
+        log('[fail] patching user.profile', error);
         response.statusCode = 400; // Bad Request
-        response.json({ error: 'Problems parsing JSON' }, null, 2);
+        response.json({ error: 'Invalid JSON Patch' }, null, 2);
         return;
       }
 
-      jsonpatch.applyPatch(user.profile, operations);
       db.save();
-
       response.json(user.profile, null, 2);
     });
   },
@@ -132,29 +131,27 @@ configurationsAPI.patch({
       json += String(chunk);
     });
     request.on('end', () => {
-      let operations;
       try {
-        operations = JSON.parse(json);
+        const operations = JSON.parse(json);
         const changedFiles = operations
           .map(operation => operation.path.replace(/^\//, ''));
 
         for (const file of changedFiles) {
           if (!configurations.allowed.includes(file)) {
             response.statusCode = 400; // Bad Request
-            response.json({
-              error: 'Updating ' + file + ' is forbidden'
-            }, null, 2);
+            response.json({ error: `Updating ${file} is forbidden` }, null, 2);
             return;
           }
         }
+
+        jsonpatch.applyPatch(user.configurations, operations, true);
       } catch (error) {
         log('[fail] patching user.configurations', error);
         response.statusCode = 400; // Bad Request
-        response.json({ error: 'Problems parsing JSON' }, null, 2);
+        response.json({ error: 'Invalid JSON Patch' }, null, 2);
         return;
       }
 
-      jsonpatch.applyPatch(user.configurations, operations);
       db.save();
       response.json(user.configurations, null, 2);
     });
--- a/lib/users.js
+++ b/lib/users.js
@@ -271,6 +271,17 @@ function getOrCreateUser (email) {
     user.profile = {
       name: ''
     };
+    delete user.name;
+  }
+
+  // Temporary migration code: Previous users didn't have machine properties.
+  for (const projectId in user.machines) {
+    for (const machine of user.machines[projectId]) {
+      if (!machine.properties) {
+        machine.properties = { name: machine.name };
+        delete machine.name;
+      }
+    }
   }
 
   // Temporary migration code: Previous users didn't have configuration files.
--- a/static/css/janitor.css
+++ b/static/css/janitor.css
@@ -280,6 +280,20 @@ h4 {
   color: #c9302c;
 }
 
+.editable-label:not(.editing) .editable-editor,
+.editable-label.editing .editable-value {
+  display: none;
+}
+
+.editable-editor {
+  width: 100%;
+}
+
+.editable-toggle {
+  margin-left: 10px;
+  cursor: pointer;
+}
+
 /* Lists */
 
 .list-group-item {
@@ -332,11 +346,41 @@ h4 {
 
 .project-title {
   flex-grow: 1;
-  padding: 0 10px;
-  margin: 0;
+  margin: 0 10px;
   display: flex;
   flex-wrap: wrap;
   align-items: center;
+}
+
+.project-title .editable-label {
+  width: 100%;
+}
+
+.project-title .editable-value {
+  display: flex;
+  flex-grow: 1;
+  align-items: center;
+}
+
+.project-title .editable-toggle {
+  color: #777;
+}
+
+.panel-project:not(:hover) .editable-toggle:not(:focus) {
+  visibility: hidden;
+}
+
+.project-title .editable-editor .form-control {
+  margin-bottom: 0;
+}
+
+.project-title h4 {
+  margin: 0;
+  hyphens: auto;
+}
+
+.project-title .form-control-feedback {
+  top: 0;
 }
 
 .project-status {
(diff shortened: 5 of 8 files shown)


jankeromnes:
Sure! Thanks a lot for this pull request! 😄

api/hosts-api.js
```diff
@@ -336,7 +336,98 @@ hostAPI.get('/version', {
   }]
 });
 
-hostAPI.get('/:container/:port', {
+const containerAPI = hostAPI.api('/:container');
+
+containerAPI.patch({
+  title: 'Update a given container',
```
jankeromnes: Nit: Please use a shorter title like "Update a container".

api/hosts-api.js
```diff
@@ -336,7 +336,98 @@ hostAPI.get('/version', {
...
+  description: 'Update a container\'s information.',
```
jankeromnes: Nit: Please use a more specific description like "Update the properties of a given Docker container.".

api/hosts-api.js
```diff
@@ -336,7 +336,98 @@ hostAPI.get('/version', {
...
+      body: JSON.stringify({
+        name: 'Container McContaineryFace'
+      }, null, 2)
+    }
+  }]
+});
+
+containerAPI.get('/:port', {
   title: 'Get a single container port',
   description: 'Get information about a given Docker container port.',
```
jankeromnes: Suggestion: If you'd like to make this pre-existing description consistent with my nit above, please drive-by fix it to something like "Get the properties of a given Docker container port.".

api/hosts-api.js
```diff
@@ -336,7 +336,98 @@ hostAPI.get('/version', {
...
+      try {
+        operations = JSON.parse(json);
+      } catch (error) {
+        response.statusCode = 400; // Bad Request
+        response.json({ error: 'Problems parsing JSON' }, null, 2);
+        return;
+      }
+
+      // FIXME: Allow real JSON Patch operations to be applied to a safe subset
+      // of the container's data (e.g. a `container.properties` object).
```
jankeromnes: Nit: Technically, we still call containers "machines" internally. It might make sense to reword this comment to "a safe subset of a machine's data (e.g. a `machine.properties` object)" for now.

templates/contributions.html
```diff
@@ -12,7 +12,9 @@ <h2 class="header">My Contributions</h2>{{
           <div class="panel-heading">
             <img class="project-icon" src="{{= project.icon in xmlattr}}" alt="{{= project.name in xmlattr}} Logo">
             <h4 class="project-title">
-              {{= project.name in html}} #{{= id in id}}
+              <form action="/api/hosts/{{= machine.docker.host}}/{{= machine.docker.container}}" method="patch">
```
jankeromnes: Nit: Even if they're not supposed to contain wacky characters, please escape the machine's `host` and `container` references, e.g. like so `{{= machine.docker.host in xmlattr}}/{{= machine.docker.container in xmlattr}}` just in case.

nt1m:
@bnjbvr @jankeromnes I've addressed all nits except the UX one (I need to figure out how to actually test this out in dev).

jankeromnes:
@nt1m Very cool, thanks!

Here is a hacky way to make a container show up in your contribution list. Simply kill the server, and add a `"firefox"` project and machine to your `db.json` like so:

```json
{
  "projects": {
    "firefox": {
      "id": "firefox",
      "name": "Firefox",
      "icon": "",
      "docker": {
        "host": "localhost",
        "image": "",
        "path": "",
        "ports": {
          "22": {
            "proxy": "none"
          },
          "8088": {
            "proxy": "https"
          }
        },
        "build": "",
        "update": "",
        "logs": ""
      },
      "data": {}
    }
  },
  "users": {
    "admin@localhost": {
      "machines": {
        "firefox": [
          {
            "id": 0,
            "name": "Firefox",
            "ports": {
              "ssh": 42024,
              "vnc": 42025
            },
            "docker": {
              "image": "firefox",
              "logs": "",
              "ports": {
                "22": {
                  "port": 42024,
                  "proxy": "none"
                },
                "8088": {
                  "port": 42025,
                  "proxy": "https"
                }
              },
              "container": "abcdef1234567890",
              "host": "localhost"
            },
            "status": "started",
            "data": {
              "updated": 1489079794563
            }
          }
        ]
      }
    }
  }
}
```

nt1m:
@jankeromnes @bnjbvr I've fixed the UX, the PR should now be ready for review.

static/css/janitor.css
```diff
@@ -280,6 +280,16 @@ h4 {
   color: #c9302c;
 }
 
+.editable-label:not(.editing) .editable-editor,
+.editable-label:not(.editing) .form-control-feedback,
+.editable-label.editing .editable-value {
```
jankeromnes: Nit: Maybe there is a way to capture elements less precisely, so that any new markup in this section will be also hidden/shown when useful? (E.g. `.editable-label:not(.editing) > *`).

templates/contributions.html
```diff
@@ -11,10 +11,15 @@ <h2 class="header">My Contributions</h2>{{
         <div class="panel panel-default panel-project">
           <div class="panel-heading">
             <img class="project-icon" src="{{= project.icon in xmlattr}}" alt="{{= project.name in xmlattr}} Logo">
-            <h4 class="project-title">
-              {{= project.name in html}} #{{= id in id}}
-              <small class="project-status" data-status="{{= machine.status in xmlattr}}"></small>
-            </h4>
+            <form class="project-title editable-label ajax-form has-feedback" action="/api/hosts/{{= machine.docker.host in xmlattr}}/{{= machine.docker.container in xmlattr}}" method="patch" data-refresh-after-success="true">
+              <div class="editable-editor">
```
jankeromnes: Nit: You can make this a form, instead of tweaking the `.project-title` thing (no need to include the `h4` in a form).

templates/contributions.html
```diff
@@ -11,10 +11,15 @@ <h2 class="header">My Contributions</h2>{{
...
+                <input class="form-control" type="text" name="/name" value="{{= machine.properties.name in xmlattr }}" data-submit-on="blur">
+              </div>
+              <h4 class="editable-value">
+                <span>{{= machine.properties.name in xmlattr }}</span>
+                <span class="glyphicon glyphicon-pencil editable-toggle"></span>
```
jankeromnes: Nit: This button feels pretty huge; a little too close to the text; and I think it should be hidden unless you hover over the `.project-title` element.

templates/contributions.html
```diff
@@ -27,6 +32,9 @@ <h4 class="project-title">
             </div>
           </div>
           <div class="panel-body">
+            <label>Status</label>
+            <small class="project-status" data-status="{{= machine.status in xmlattr}}"></small>
```
jankeromnes: Nit: Please align this better. (Also maybe it doesn't need to be a `small`, maybe a `span`?)

jankeromnes:
(Woops, sorry for accidentally closing this pull request!)

api/admin-api.js
```diff
@@ -103,6 +103,12 @@ oauth2providerAPI.patch({
       }
 
       // Apply the requested changes to the provider.
+      if (jsonpatch.validate(operations)) {
```
jankeromnes: I think the correct usage of `jsonpatch.validate` is to pass it `operations` as well as the target object (i.e. `provider` in this case). And it returns an array of errors, so we should check if `.length > 0` (and maybe even log the JSON patch errors, since they should be fairly rare).

Please refer to the [jsonpatch usage](https://github.com/Starcounter-Jack/JSON-Patch#usage) documentation.

api/admin-api.js
```diff
@@ -103,6 +103,12 @@ oauth2providerAPI.patch({
...
+        response.statusCode = 400; // Bad Request
+        response.json({ error: 'Problems parsing JSON patch' }, null, 2);
```
jankeromnes: Nit: It's not really a problem with parsing. Maybe something more like `Invalid JSON Patch operations`?

api/hosts-api.js
```diff
@@ -336,13 +337,120 @@ hostAPI.get('/version', {
...
+      let operations = null;
+      try {
+        operations = JSON.parse(json);
+      } catch (error) {
+        response.statusCode = 400; // Bad Request
+        response.json({ error: 'Problems parsing JSON' }, null, 2);
+        return;
+      }
+
+      if (jsonpatch.validate(operations)) {
```
jankeromnes: Nit: Improper use of `jsonpatch.validate` (see earlier comment).

api/hosts-api.js
```diff
@@ -336,13 +337,120 @@ hostAPI.get('/version', {
...
+    response: {
+      body: JSON.stringify({ name: 'Much container. Such wow.' }, null, 2)
+    }
+  },{
+    request: {
+      urlParameters: {
+        hostname: 'example.com',
+        container: 'abcdef0123456789',
+      },
+      body: JSON.stringify({ name: 'Container McContaineryFace' }, null, 2)
```
jankeromnes: Nit: The reason this example fails is quite subtle. Maybe make it more obvious with something like `body: 'This is not a valid JSON Patch.'`.

api/user-api.js
```diff
@@ -67,9 +67,14 @@ userAPI.patch({
         return;
       }
 
+      if (jsonpatch.validate(operations)) {
```
jankeromnes: Nit: Same as above.

api/user-api.js
```diff
@@ -153,6 +158,12 @@ configurationsAPI.patch({
         response.json({ error: 'Problems parsing JSON' }, null, 2);
         return;
       }
+        
+      if (jsonpatch.validate(operations)) {
```
jankeromnes: Nit: Ditto.

lib/users.js
```diff
@@ -269,6 +269,16 @@ function getOrCreateUser (email) {
...
 
+  // Temporary migration code: Previous users didn't have machine properties.
+  for (const projectId in user.machines) {
+    for (const machine of user.machines[projectId]) {
+      if (!machine.properties) {
+        machine.properties = { name: machine.name };
+        delete machine.name;
+      }
+    }
+  }
```
jankeromnes: Nit: In general, we should place this temporary migration snippets after one another (to prevent an error like what happened when we tried to generate configuration files for users that didn't have names yet). However, it looks like you're lucky here and there is no dependence on earlier migration scripts.

package.json
```diff
@@ -34,6 +34,7 @@
     "email-login": "~1.1.0",
     "fast-json-patch": "~2.0.3",
     "fleau": "~16.2.0",
+    "fsos": "^1.1.2",
```
jankeromnes: Please remove this import, as we don't use it in Janitor directly (it's already imported by email-login, for example).

jankeromnes:
Also, please rebase and re-push this branch, which should fix the Travis tests (or at least, fail because of your actual API examples instead of an unrelated dependency error).

api/admin-api.js
nt1m: I'm not sure what you mean: https://github.com/Starcounter-Jack/JSON-Patch#jsonpatchvalidatepatch-operation-document-any-validator-function-jsonpatcherror

"If there are no errors, returns undefined. If there is an errors, returns a JsonPatchError object with the following properties"

nt1m: Also, the target argument is optional. Not sure we want to put it ?

package.json
nt1m: Whoops, not sure how that slipped in.

lib/users.js
nt1m: I took the order in account, and I noticed there was no dependence as well :) Not sure what would be the preferable one for you ?

nt1m:
@nt1m: review?(@jankeromnes)

api/user-api.js
```diff
@@ -147,14 +145,15 @@ configurationsAPI.patch({
             return;
```
jankeromnes: Nit: Doesn't the above `response.json(...)` fit in one line? If so, please drive-by fix it.

api/user-api.js
```diff
@@ -58,18 +58,17 @@ userAPI.patch({
       json += String(chunk);
     });
     request.on('end', () => {
-      let operations = null;
       try {
-        operations = JSON.parse(json);
+        const operations = JSON.parse(json);
+        jsonpatch.applyPatch(user.profile, operations, true);
       } catch (error) {
+        log('[fail] json patch', error);
```
jankeromnes: Nit: Please make this logged message more explicit (e.g. `log('[fail] patching user.profile', error)`).

api/admin-api.js
```diff
@@ -92,18 +92,16 @@ oauth2providerAPI.patch({
       json += String(chunk);
     });
     request.on('end', () => {
-      let operations = null;
       try {
-        operations = JSON.parse(json);
+        const operations = JSON.parse(json);
+        jsonpatch.applyPatch(provider, operations, true);
       } catch (error) {
         log('[fail] json patch', error);
```
jankeromnes: Nit: Please drive-by fix this logged message to make it more explicit, e.g. `log('[fail] patching oauth2 provider', error)`.

lib/users.js
```diff
@@ -273,6 +273,16 @@ function getOrCreateUser (email) {
     };
```
jankeromnes: Nit: Please drive-by improve this migration code by adding a `delete user.name` line.

static/js/janitor.js
```diff
@@ -223,6 +223,13 @@ function getFormData (form) {
   }, {});
 }
 
+// Setup editable labels.
+Array.forEach(document.querySelectorAll('.editable-label'), function (label) {
+  label.querySelector('.editable-toggle').addEventListener('click', function () {
```
jankeromnes: Nit: You could make this code more robust by separating out the query selector, and ensuring it actually returned an element, e.g.

```js
const toggle = label.querySelector('.editable-toggle');
if (!toggle) {
  console.error('Editable label', label, 'has no ".editable-toggle" element!');
  return;
}
```

templates/contributions.html
```diff
@@ -11,10 +11,15 @@ <h2 class="header">My Contributions</h2>{{
         <div class="panel panel-default panel-project">
           <div class="panel-heading">
             <img class="project-icon" src="{{= project.icon in xmlattr}}" alt="{{= project.name in xmlattr}} Logo">
-            <h4 class="project-title">
-              {{= project.name in html}} #{{= id in id}}
-              <small class="project-status" data-status="{{= machine.status in xmlattr}}"></small>
-            </h4>
+            <div class="editable-label project-title">
+              <form action="/api/hosts/{{= machine.docker.host in xmlattr}}/{{= machine.docker.container in xmlattr}}" class="ajax-form editable-editor has-feedback" data-refresh-after-success="true" method="patch">
+                <input class="form-control" data-submit-on="blur" name="/name" type="text" value="{{= machine.properties.name in xmlattr }}">
```
jankeromnes: Nit: By playing with the renaming feature, I actually lost track of which container was which. It's my fault, but it made me think, maybe would could add a `placeholder="{{= project.name in xmlattr }} #{{= id in id}}"` attribute to this input?

templates/contributions.html
```diff
@@ -11,10 +11,15 @@ <h2 class="header">My Contributions</h2>{{
...
+              </form>
+              <div class="editable-value">
+                <h4>{{= machine.properties.name in xmlattr }}</h4>
```
jankeromnes: Nit: The escape function should be `in html` here.

Nit: If you accept my suggestion above about the `placeholder` attribute, I think it would make sense to change this to display the default machine name when `machine.properties.name` is an empty string, e.g. like so:

```html
<h4>{{if machine.properties.name
  then {{{{= machine.properties.name in html }}}}
  else {{{{= project.name in html}} #{{= id in id}}}}}}</h4>
```

static/css/janitor.css
```diff
@@ -332,13 +346,39 @@ h4 {
...
+}
+
+.project-title .editable-value {
+  display: flex;
+  flex-grow: 1;
+  align-items: center;
+}
+
+.panel-project:not(:hover) .editable-toggle {
+  opacity: 0;
```
jankeromnes: Not a nit, just a question: Why use `opacity: 0` instead of `visibility: hidden` or `display: none` on the `:hover` selector?

Nit: Please also do it for the `:focus` selector (e.g. for touch screens).

static/css/janitor.css
```diff
@@ -280,6 +280,20 @@ h4 {
...
+  display: none;
+}
+
+.editable-editor {
+  width: 100%;
+}
+
+.editable-toggle {
+  margin-left: 10px;
+  cursor: pointer;
```
jankeromnes: Nit: This `.editable-toggle` icon appears black instead of the expected `#777` color. Any idea why? If you'd like, you can show it black when it is itself `:hover`ed or `:focus`ed, but not when the `.panel-project` is focused (that should only control visibility of the `.editable-toggle`, not its focus-state).

api/hosts-api.js
```diff
@@ -336,13 +337,113 @@ hostAPI.get('/version', {
...
-const containerAPI = hostAPI.api('/:container', {
-  title: 'Containers'
+const containerAPI = hostAPI.api('/:container', 'Containers');
+
+containerAPI.patch({
+  title: 'Update a container',
+  description: 'Update the properties of a given Docker container.',
+
+  handler: (request, response) => {
+    let { user } = request;
```
nt1m: npm run lint didn't complain, so I guess not.

api/hosts-api.js
```diff
@@ -336,13 +337,113 @@ hostAPI.get('/version', {
...
+    }
+
+    let json = '';
+    request.on('data', chunk => {
+      json += String(chunk);
+    });
+    request.on('end', () => {
+      try {
+        const operations = JSON.parse(json);
+        jsonpatch.applyPatch(machine.properties, operations, jsonpatch.validate);
```
nt1m: good catch, accidentally left this test in.

lib/users.js
jankeromnes: Ideally always as the last migration script in this function (safest default behavior), but you can leave this one in place if you like, since there are no obvious problems.

api/user-api.js
nt1m: It does, if I put a template string instead (but not without)

static/css/janitor.css
nt1m: no specific preference :)

nt1m: For the :focus selector, I suppose you meant: `.panel-project:not(:hover) .editable-toggle:not(:focus)` ? Since you can't focus a <div>

PR #74, opened 2017-06-06
Add an API to manage users, and use it to edit real names

jankeromnes:
This pull request adds a `/api/user` endpoint which allows `GET` and `PATCH` operations to manage the current user's profile information.

Behind the scenes, it manages a newly introduced `user.profile = { name: '' }` object, which the user will be able to change at will. (In the future, it might contains things like avatar url, website, short bio, etc. if deemed useful).

The auto-generated User API docs look like this with this commit:

user-api-docs

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -12,4 +12,5 @@ const api = module.exports = selfapi({
 
 // Janitor API sub-resources.
 api.api('/hosts', require('./hosts-api'));
+api.api('/user', require('./user-api'));
 api.api('/admin', require('./admin-api'));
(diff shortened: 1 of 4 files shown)


jankeromnes:
@nt1m, I've changed the User API to address the `FIXME` I had previously. Could you please review this newest commit to see if it makes sense to you?

Details: I moved the `user.name` property into a dedicated `user.profile` object so that we can allow any sort of JSON Patch operations to be applied on it (not restricted to `/name` or `add`/`replace` patch operations).

Thanks!

nt1m:
Looks good to me ! :+1:

jankeromnes:
Thanks a lot for the quick review!

PR #76, opened 2017-06-08
Add .gitconfig configuration

nt1m:


lib/users.js
```diff
@@ -218,17 +219,20 @@ function getOrCreateUser (email) {
       profile: {
         name: ''
```
jankeromnes: Nit: Please add a trailing comma as a drive-by fix. Thanks!

lib/users.js
```diff
@@ -238,12 +242,20 @@ function getOrCreateUser (email) {
     db.save();
   }
 
+  // Temporary migration code: Previous users didn't have configuration
```
jankeromnes: Nit: Please add missing word `a` and finish this comment with a `.`.

lib/users.js
```diff
@@ -257,3 +269,5 @@ function getOrCreateUser (email) {
 
   return user;
 }
+
+getOrCreateUser("admin@localhost");
```
jankeromnes: We probably don't want to ship this hack.

static/js/janitor.js
```diff
@@ -8,12 +8,12 @@
...
 
-// Automatically set up delete button forms (when their 'method' is 'delete').
-Array.forEach(document.querySelectorAll('form[method=delete]'), function (form) {
+// Automatically set up delete and apply button forms (when their 'method' is 'delete' or 'put').
+Array.forEach(document.querySelectorAll('form[method=delete], form[method=put]'), function (form) {
   setupAsyncForm(form);
   form.addEventListener('submit', function (event) {
     var data = getFormData(form);
-    fetchAPI('DELETE', form.action, data, function (error, data) {
+    fetchAPI(form.getAttribute('method').toUpperCase(), form.action, data, function (error, data) {
```
jankeromnes: Nit: This line is pretty long. Please use an intermediate variable `var method = form.getAttribute('method').toUpperCase();`.

static/js/janitor.js
```diff
@@ -8,12 +8,12 @@
   };
 });
 
-// Automatically set up delete button forms (when their 'method' is 'delete').
-Array.forEach(document.querySelectorAll('form[method=delete]'), function (form) {
+// Automatically set up delete and apply button forms (when their 'method' is 'delete' or 'put').
```
jankeromnes: Nit: Please shorten this comment, e.g. as `// Automatically set up button forms (when their 'method' is 'delete' or 'put')`.

static/js/janitor.js
```diff
@@ -176,6 +177,12 @@ function updateFormStatus (form, status, message) {
     feedback.dataset.message = message;
     feedback.focus();
   }
+
+  if (form.dataset.refreshAfterSuccess && status == 'success') {
+    setTimeout(() => {
```
jankeromnes: Nit: Please don't use arrow functions on the client side, because older browsers don't support them yet and they are hard to polyfill.

api/user-api.js
```diff
@@ -76,3 +79,107 @@ userAPI.patch({
...
+configAPI.delete('/:configuration', {
+  handler(request, response) {
+    let { user, query } = request;
+    if (!user) {
+      response.statusCode = 403;
+      response.json('User not found');
+      return;
+    }
+
+    configurations.resetConfiguration(user, query.configuration);
```
jankeromnes: Nit: No need to repeat the word "configurations", please rename this method to something like `configurations.resetToDefault()`.

lib/users.js
```diff
@@ -238,12 +242,20 @@ function getOrCreateUser (email) {
     db.save();
   }
 
+  // Temporary migration code: Previous users didn't have configuration
+  if (!user.configurations) {
+    user.configurations = {};
+    for (let configuration in configurations.defaultConfigurations) {
```
jankeromnes: Nit: No need to repeat the word "configurations", please rename this export to something like `configurations.defaults`.

nt1m:
Thanks for the review! Will push a new revision soon.

api/user-api.js
```diff
@@ -76,3 +79,107 @@ userAPI.patch({
...
+configAPI.put('/:configuration', {
+  handler(request, response) {
+    let { user, query } = request;
+    if (!user) {
+      response.statusCode = 403;
+      response.json('User not found');
+      return;
+    }
+
+    let userMachines = Object.values(user.machines)
```
nt1m: It doesn't work in this case, because machines is already used by:

const machines = require('../lib/machines');

lib/machines.js
```diff
@@ -273,6 +281,30 @@ exports.destroy = function (user, projectId, machineId, callback) {
...
+      [file]: user.configurations[file],
+    }
+  }, error => {
+    if (error) {
+      callback(error);
+    }
+
+    // FIXME: Remove this workaround when the following Docker bug is fixed:
+    // https://github.com/docker/docker/issues/21651.
+    const command = `sudo chown user:user /home/user/.ssh/${file}`;
```
nt1m: What would the correct command be ?

nt1m:
@jankeromnes Pushed new revision!

lib/users.js
```diff
@@ -238,12 +242,20 @@ function getOrCreateUser (email) {
     db.save();
   }
 
+  // Temporary migration code: Previous users didn't have configuration files.
+  if (!user.configurations) {
+    user.configurations = {};
+    for (let configuration in configurations.defaultConfigurations) {
+      configurations.resetConfiguration(user, configuration);
```
jankeromnes: You probably mean `configurations.resetToDefault(user, configuration)`.

lib/users.js
```diff
@@ -238,12 +242,20 @@ function getOrCreateUser (email) {
     db.save();
   }
 
+  // Temporary migration code: Previous users didn't have configuration files.
+  if (!user.configurations) {
+    user.configurations = {};
+    for (let configuration in configurations.defaultConfigurations) {
```
jankeromnes: Nit: You probably mean `configurations.defaults`.

api/user-api.js
```diff
@@ -76,3 +79,122 @@ userAPI.patch({
...
+  },
+  examples: [{
+    response: {
+      body: JSON.stringify({ '.gitconfig': '' }, null, 2)
+    }
+  }]
+});
+
+configurationsAPI.patch({
+  title: 'Update user configuration file(s)',
```
jankeromnes: Nit: Please use a shorter title, e.g. `Update configuration files`.

api/user-api.js
```diff
@@ -76,3 +79,122 @@ userAPI.patch({
...
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' });
+      return;
+    }
+
+    const userMachines = Object.values(user.machines)
+      .reduce((acc, val) => acc.concat(val), []);
+
+    const promises = Promise.all(userMachines.map((machine) =>
+      new Promise((resolve) =>
```
jankeromnes: Nit: No need for parentheses around single arguments.

api/user-api.js
```diff
@@ -76,3 +79,122 @@ userAPI.patch({
...
+        machines.resetConfigurationFile(user, query.configuration, machine, resolve))
+    ));
+
+    promises.then(() => {
+      response.statusCode = 204; // No Content
+      response.end();
+    }).catch((error) => {
```
jankeromnes: Nit: No need for parentheses around single arguments.

api/user-api.js
```diff
@@ -76,3 +79,122 @@ userAPI.patch({
...
+      response.json({ error: 'Unauthorized' });
+      return;
+    }
+
+    const userMachines = Object.values(user.machines)
+      .reduce((acc, val) => acc.concat(val), []);
+
+    const promises = Promise.all(userMachines.map((machine) =>
+      new Promise((resolve) =>
+        machines.resetConfigurationFile(user, query.configuration, machine, resolve))
```
jankeromnes: What happens if one of these calls fails? Wouldn't `machines.resetConfigurationFile()` call `resolve(error)` instead of `reject()`/`throw()`, wrongly considering errors as successes?

api/user-api.js
```diff
@@ -76,3 +79,122 @@ userAPI.patch({
...
+    ));
+
+    promises.then(() => {
+      response.statusCode = 204; // No Content
+      response.end();
+    }).catch((error) => {
+      response.statusCode = 500; // Internal Server Error
+      response.json({ error: String(error) });
```
jankeromnes: Nit: In general, we try to `log()` Error objects, and only return generic error messages from the API (in order to prevent leaking sensitive information inside errors, which could be exploited by attacks similar to error-based SQL injections).

lib/configurations.js
```diff
@@ -0,0 +1,39 @@
+// Copyright © 2016 Jan Keromnes, Tim Nguyen. All rights reserved.
+// The following code is covered by the AGPL-3.0 license.
+
+'use strict';
+
+const fs = require('fs');
+const { create: createTemplate } = require('fleau');
```
jankeromnes: Nit: We try to keep imports ordered alphabetically. Also, we try to keep them as simple as possible: would you be open to do `const fleau = require('fleau')` and use `fleau.create()` below?

lib/configurations.js
```diff
@@ -0,0 +1,39 @@
...
+  fs.readdir(path, (error, files) => {
+    files.forEach(configuration => {
+      fs.readFile(path + configuration, "utf8", (error, template) => {
+        exports.defaults[configuration] = template;
+      });
+    });
+  });
+}
+
+loadDefaults();
```
jankeromnes: Nit: The convention in other files is to name this function `load()`, and to call it right below the collection(s) that it's trying to fill (if any). E.g. here, you could move it right below `exports.defaults = {};`, and use JS's hoisting mechanism to call the function before its actual definition.

lib/configurations.js
```diff
@@ -0,0 +1,39 @@
...
+
+exports.resetToDefault = function (user, configuration) {
+  if (!exports.defaults[configuration]) {
+    log('[fail] configuration does not exist:', configuration);
```
jankeromnes: Nit: This could happen to custom configurations that users have created via the JSON Patch API, but for which there is no default template. Please use a more specific message, e.g. `[fail] default configuration not found:`.

lib/configurations.js
```diff
@@ -0,0 +1,39 @@
...
+    return;
+  }
+  const stream = createTemplate(exports.defaults[configuration])({ user });
+  let str = '';
+  stream.on('data', c => str += c);
```
jankeromnes: Nit: Please use full, explicit variable names, e.g. `chunk` and `content`.

lib/routes.js
```diff
@@ -205,6 +205,7 @@ exports.contributionsPage = function (response, user = null) {
 const settingsHeader = camp.template('./templates/settings-header.html');
 const settingsSections = {
   account: camp.template('./templates/settings-account.html'),
+  configurations: camp.template('./templates/settings-configurations.html'),
```
jankeromnes: I think you forgot to `git add` this new template.

lib/configurations.js
```diff
@@ -0,0 +1,39 @@
...
+const log = require('./log');
+
+exports.defaults = {};
+
+function loadDefaults() {
+  const path = './templates/configurations/';
+  fs.readdir(path, (error, files) => {
+    files.forEach(configuration => {
+      fs.readFile(path + configuration, "utf8", (error, template) => {
+        exports.defaults[configuration] = template;
```
jankeromnes: I think you also forgot to `git add templates/user/.gitconfig` (it might be worth checking which local files are still untracked after you commited, e.g. with `git status`).

api/user-api.js
```diff
@@ -76,3 +80,132 @@ userAPI.patch({
...
+  description: 'Overwrite a configuration file in all the user\'s containers (any local changes will be lost!)',
+  handler ({ user, query }, response) {
+    if (!user) {
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' });
+      return;
+    }
+
+    const userMachines = Object.values(user.machines)
+      .reduce((acc, val) => acc.concat(val), []);
```
jankeromnes: I hit a bug with this code because the `user` I was testing with had some machines with an empty `machine.docker.container` (this is expected, e.g. when you have machines with `machine.status === 'new'`, which are basically placeholders not actually tied to any real container).

You might want to add a filter to eliminate any machines where `machine.status` is `'new'` or `'start-failed'` (you'd only want `'started'`), or even better you could filter out any machines that don't have a valid `machine.docker.container` (e.g. with `container && container.length >= 16 && /^[0-9a-f]+$/.test(container)`).

However, since this is touching a lot of `machine` internals that could change over time, please move this whole code into a new `machines` module function (e.g. `machines.deployConfigurationFile(user, configuration, callback)`). This way the User API is kept simple and readable, and any machine-internal-aware code is kept inside the machines module.

api/user-api.js
```diff
@@ -76,3 +80,132 @@ userAPI.patch({
...
+
+    const promises = Promise.all(userMachines.map(machine =>
+      new Promise((resolve, reject) =>
+        machines.resetConfigurationFile(user, query.configuration, machine, error => {
+          if (!error) {
+            return;
+          }
+          reject(error);
```
jankeromnes: Nit: This could be simplified as:

```js
if (error) {
  reject(error);
}
```

api/user-api.js
```diff
@@ -76,3 +80,132 @@ userAPI.patch({
...
+        })
+      )
+    ));
+
+    promises.then(() => {
+      response.statusCode = 204; // No Content
+      response.end();
+    }).catch(error => {
+      response.statusCode = 500; // Internal Server Error
+      log('PUT', query.configuration, error);
```
jankeromnes: Nit: Please log something more descriptive, e.g. `log('[fail] could not deploy configuration:', query.configuration, error)`.

lib/configurations.js
```diff
@@ -0,0 +1,39 @@
+// Copyright © 2016 Jan Keromnes, Tim Nguyen. All rights reserved.
```
jankeromnes: Nit: Not very important, but it's 2017 :)

lib/configurations.js
```diff
@@ -0,0 +1,39 @@
...
+// The following code is covered by the AGPL-3.0 license.
+
+'use strict';
+
+const fs = require('fs');
+const fleau = require('fleau');
```
jankeromnes: Nit: Please alpha-sort imports (`fleau` comes before `fs`).

lib/configurations.js
```diff
@@ -0,0 +1,39 @@
...
+exports.defaults = {};
+
+load();
+
+function load() {
+  const path = './templates/configurations/';
+  fs.readdir(path, (error, files) => {
+    files.forEach(configuration => {
+      fs.readFile(path + configuration, "utf8", (error, template) => {
+        exports.defaults[configuration] = template;
```
jankeromnes: Nit: Please pre-compile all templates ahead of time by assigning this to `fleau.create(template)` instead of just `template`.

lib/configurations.js
```diff
@@ -0,0 +1,39 @@
...
+    });
+  });
+}
+
+exports.resetToDefault = function (user, configuration) {
+  if (!exports.defaults[configuration]) {
+    log('[fail] default configuration for', configuration, 'does not exist.');
+    return;
+  }
+  const stream = fleau.create(exports.defaults[configuration])({ user });
```
jankeromnes: Nit: Following the nit above, this would become `const stream = exports.defaults[configuration]({ user })`.

lib/configurations.js
```diff
@@ -0,0 +1,39 @@
...
+      fs.readFile(path + configuration, "utf8", (error, template) => {
+        exports.defaults[configuration] = template;
+      });
+    });
+  });
+}
+
+exports.resetToDefault = function (user, configuration) {
+  if (!exports.defaults[configuration]) {
+    log('[fail] default configuration for', configuration, 'does not exist.');
```
jankeromnes: Nit: If you make `exports.resetToDefault` take a `callback` parameter, you could replace this `log(...)` call with `callback(new Error('Default configuration not found: ' + configuration))`.

lib/configurations.js
```diff
@@ -0,0 +1,39 @@
...
+    return;
+  }
+  const stream = fleau.create(exports.defaults[configuration])({ user });
+  let content = '';
+  stream.on('data', chunk => content += chunk);
+  stream.on('end', () => {
+    user.configurations[configuration] = content;
+    db.save();
```
jankeromnes: If you add a `callback` parameter, you could call it below this line in order to indicate success: `callback();` (with no `error` parameter because everything went well).

static/js/janitor.js
```diff
@@ -86,12 +87,13 @@ function ajaxForm (selector, action, callback) {
...
     method: method.toUpperCase(),
-    headers: {
+    headers: new Headers({
       'Accept': 'application/json',
       'Content-Type': 'application/json'
-    },
+    }),
     credentials: 'same-origin',
     body: JSON.stringify(data, null, 2)
   }).then(function (response) {
```
jankeromnes: Nit: Since you are using "204 No Content" replies in your API, please add support for empty replies to this helper (e.g. by returning `null` below instead of `response.json()` when `response.status === 204`.

templates/settings-configurations.html
```diff
@@ -0,0 +1,16 @@
+    <section class="settings">
+      <div class="container">
+        <div class="row">
+          <form action="/api/user/configurations" class="account-form ajax-form has-feedback col-sm-6" method="patch">
```
jankeromnes: Nit: This form can take the full width regardless of screen size. Please change `col-sm-6` to `col-sm-12`.

templates/settings-configurations.html
```diff
@@ -0,0 +1,16 @@
...
+            <label>.gitconfig</label>
+            <textarea class="form-control" data-submit-on="blur" name="/.gitconfig" placeholder=".gitconfig file">{{= user.configurations['.gitconfig'] in html}}</textarea>
```
jankeromnes: Nit: Please make this textarea prettier, e.g. by using Ace or CodeMirror (making sure it works great on mobile too), or at least add a `rows="7"` attribute.

templates/settings-configurations.html
```diff
@@ -0,0 +1,16 @@
...
+          </form>
+          <form action="/api/user/configurations/.gitconfig" class="account-form ajax-form has-feedback col-sm6" method="delete" data-refresh-after-success="true">
+            <button class="btn btn-danger">Reset configuration to default</button>
```
jankeromnes: Nit: Please use shorter texts in buttons, e.g. `Reset to Default`.

templates/settings-configurations.html
```diff
@@ -0,0 +1,16 @@
+    <section class="settings">
+      <div class="container">
+        <div class="row">
+          <form action="/api/user/configurations" class="account-form ajax-form has-feedback col-sm-6" method="patch">
+            <label>.gitconfig</label>
+            <textarea class="form-control" data-submit-on="blur" name="/.gitconfig" placeholder=".gitconfig file">{{= user.configurations['.gitconfig'] in html}}</textarea>
+          </form>
+          <form action="/api/user/configurations/.gitconfig" class="account-form ajax-form has-feedback col-sm6" method="delete" data-refresh-after-success="true">
```
jankeromnes: Nit: You probably mean `col-sm-6`.

jankeromnes: Nit: Please order HTML attributes alphabetically (`data-refresh-after-success` before `method`).

templates/settings-configurations.html
```diff
@@ -0,0 +1,16 @@
...
+            <button class="btn btn-danger">Reset configuration to default</button>
+          </form>
+          <form action="/api/user/configurations/.gitconfig" class="account-form ajax-form has-feedback col-sm6" method="put">
```
jankeromnes: Nit: You probably mean `col-sm-6`.

templates/settings-configurations.html
```diff
@@ -0,0 +1,16 @@
...
+            <button class="btn btn-danger">Overwrite .gitconfig on existing machines</button>s
```
jankeromnes: Nit: Please transform this big button into a text with button or link like you suggested earlier (something like `Any changes will only apply to new containers.<br>To overwrite .gitconfig in existing containers, please click **here**
 (careful, you might lose data!)`).

static/js/janitor.js
```diff
@@ -86,18 +87,19 @@ function ajaxForm (selector, action, callback) {
...
     // The server is responding!
     responseStatus = response.status;
-    return response.json();
+    return responseStatus == 204 ? null : response.json();
```
jankeromnes: Nit: When you return `null`, the `if (data.error)` check below fails. Please change that check to `if (data && data.error)`.

templates/settings-configurations.html
```diff
@@ -0,0 +1,19 @@
...
+            <textarea class="form-control" data-submit-on="blur" name="/.gitconfig" placeholder=".gitconfig file" rows="7">{{= user.configurations['.gitconfig'] in html}}</textarea>
+          </form>
+          <form action="/api/user/configurations/.gitconfig" class="account-form ajax-form has-feedback col-sm-12" data-refresh-after-success="true" method="delete">
+            <button class="btn btn-danger">Reset to Default</button>
+          </form>
+          <form action="/api/user/configurations/.gitconfig" class="account-form ajax-form has-feedback col-sm-12" method="put">
+            <p>
+            Any changes will only apply to new containers.<br/>
+            To overwrite .gitconfig in existing containers, please click <button class="btn btn-danger btn-xs">here</button> (be careful, you might lose data!).
+            </p>
```
jankeromnes: Nit: Please move the `<p>` explanation right below the `<textarea>`, and call the two buttons `Reset to Default` and `Overwrite in All Containers`.

Nit: Maybe it would also be nice not to use `data-submit-on="blur"` but have a green `Save` button?

nt1m: Test

jankeromnes: It works! 🎉

PR #79, opened 2017-06-15
Slightly improve the Configurations settings panel.

jankeromnes:
In this change I tweak the API and the front-end code to slightly improve the UX of the new "Configurations" panel. This changes:
- The "Configurations" panel now has a short description, and the inputs & buttons are placed more elegantly.
- "Deploy configuration" API now returns the number of modified containers (e.g. "Successfully deployed to 3 containers").
- On the front-end, any `data.message` attribute will be displayed (if there is no error to display instead).

### DIFF
--- a/api/user-api.js
+++ b/api/user-api.js
@@ -85,7 +85,7 @@ userAPI.patch({
 const configurationsAPI = userAPI.api('/configurations');
 
 configurationsAPI.get({
-  title: 'Get all configuration files',
+  title: 'Get all user configurations',
 
   handler ({ user }, response) {
     if (!user) {
@@ -105,9 +105,8 @@ configurationsAPI.get({
 });
 
 configurationsAPI.patch({
-  title: 'Update configuration files',
-  description:
-    'Update any of the user\'s configuration file(s) (using JSON Patch).',
+  title: 'Update user configurations',
+  description: 'Update any user configuration file(s) (using JSON Patch).',
 
   handler (request, response) {
     const { user } = request;
@@ -151,11 +150,11 @@ configurationsAPI.patch({
 });
 
 // API sub-resource to manage a single configuration file.
-const configurationAPI = configurationsAPI.api('/:configuration');
+const configurationAPI = configurationsAPI.api('/:file');
 
 configurationAPI.delete({
-  title: 'Reset a configuration file',
-  description: 'Reset a configuration file to its default template value.',
+  title: 'Reset a user configuration',
+  description: 'Reset a user configuration file to its default template value.',
 
   handler ({ user, query }, response) {
     if (!user) {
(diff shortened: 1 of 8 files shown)


jankeromnes:
@nt1m Could you please have a look at these changes?

static/js/janitor.js
```diff
@@ -182,7 +182,7 @@ function updateFormStatus (form, status, message) {
   if (form.dataset.refreshAfterSuccess && status == 'success') {
     setTimeout(function () {
       location.reload();
-    }, 1000);
+    }, 400);
```
jankeromnes: 400ms is already pretty slow, and 300ms would probably be better (for a quick success confirmation and subsequent page refresh). As indicated in [this article](http://blog.teamtreehouse.com/perceived-performance), 100ms is generally perceived as instantaneous, while 1s is perceived as sluggish (users feel the pause).

And for my personal taste, waiting for 1s after a successful change is frustrating, because I feel compelled to wait for it to happen, but 1s is at the limit of breaking my user flow.

static/css/janitor.css
```diff
@@ -240,6 +257,25 @@ h4 {
   margin-right: 10px;
 }
 
+.top-right-controls {
+  display: flex;
+  flex-direction: row-reverse;
+  flex-wrap: wrap;
```
jankeromnes: Fair enough, I will address this and re-push my commit.

api/user-api.js
```diff
@@ -196,25 +195,27 @@ configurationAPI.put({
...
     }
 
-    const { configuration } = query;
-    machines.deployConfigurationFileInAllContainers(user, configuration)
-      .then(() => {
-        response.statusCode = 204; // No Content
-        response.end();
+    const { file } = query;
+    machines.deployConfigurationInAllContainers(user, file)
+      .then(count => {
```
jankeromnes: Fair enough. The reason I prefer:

```js
foo
  .then(count => {
    // ...
  })
  .catch(error => {
    // ...
  })
```

over:
```js
foo.then(count => {
  // ...
}).catch(error => {
  // ...
})
```

is that even if the indentation is slightly off, it's really easy to visually isolate `then` and `catch` blocks. Additionally, it makes things easier to read when you have a `veryLongVariableNameHoldingAPromiseThatMightSucceedOrFail.then(count => {})` (putting `.then` on the next line avoid having to look for it all the way at the end of a line, sometimes breaking the 80-char or 100-char limits).

static/js/janitor.js
nt1m: Sounds good, I probably didn't notice the 1s because I focused on other errors.

jankeromnes:
Cheers!

PR #81, opened 2017-06-20
Add default .gitignore template to user configurations.

jankeromnes:
This commit also makes the "Configurations" panel more generic (the template now supports any number of configurations, although the UX won't scale to many more configurations and we'll probably have to tweak it at some point).

@nt1m please have a look.


lib/machines.js
```diff
@@ -212,11 +212,13 @@ exports.spawn = function (user, projectId, callback) {
...
 
-    // Install all the user's configuration files into this container.
+    // Install all non-empty user configuration files into this container.
     Object.keys(user.configurations).forEach(file => {
+      if (!user.configurations[file]) {
+        return;
+      }
       exports.deployConfiguration(user, machine, file).then(() => {
         log('spawn-config', file, container.id.slice(0, 16), 'success');
-        db.save();
```
nt1m: Good catch! Thanks for fixing this

lib/users.js
```diff
@@ -245,8 +246,15 @@ function getOrCreateUser (email) {
   // Temporary migration code: Previous users didn't have configuration files.
   if (!user.configurations) {
     user.configurations = {};
-    for (let file in configurations.defaults) {
-      configurations.resetToDefault(user, file);
+  }
+  for (let configuration in configurations.defaults) {
+    const file = configuration;
+    if (!user.configurations[file]) {
+      configurations.resetToDefault(user, file, error => {
```
nt1m: You can avoid this file variable by doing this (which you've suggested to me last time !):

```js
Object.keys(configurations.defaults).forEach(file => {
  ...
});
```

PR #83, opened 2017-06-23
Minor dependency green-keeping.

jankeromnes:


jankeromnes:
Trivial change, merging without review.

PR #84, opened 2017-06-23
Fix several fleau templating bugs.

jankeromnes:


jankeromnes:
Important hot fix, merging without review.

PR #85, opened 2017-06-23
Fix `jsonpatch.apply()` calls to `jsonpatch.applyPatch()`.

jankeromnes:
The exported method name recently changed to be more explicit.


jankeromnes:
Another quick hot fix, merging without review.

PR #86, opened 2017-06-23
Fix deploying configuration files to all containers.

jankeromnes:
The status `filter` was wrongly applied to the Array of project machines *before* they were `reduce`d to a single Array of machines. The structure looks like this:

```js
user.machines = {
  "firefox": [ machine0, machine1 ],
  "chromium": [ machine0 ],
}
```

The existing code was wrongly checking `[ machine0, machine1 ].status === 'started'`, which doesn't work because an Array has no "status" property.

The fix is to make multiple `machine0.status === 'started'` checks on all machines.


jankeromnes:
Small hot fix, merging without review.

PR #91, opened 2017-06-30
Add .hgrc configuration

nt1m:


jankeromnes:
@padenot and @kewisch, could you please have a look at the proposed `.hgrc` template, and verify that it will be useful for both Firefox and Thunderbird developers? (Please suggest any tweaks that would make it more useful for you or other developers).

(Also, thanks @padenot for your personal `.hgrc` file from a few months ago, as I believe this template is based on it.)

Thanks!

kewisch:
This shares a lot with my hgrc, so lgtm. You could add:

```
[defaults]
qnew = -D -U
qrefresh = -D
```

for those still using mq to add the current date and user to the patch when creating and refreshing mq patches.

padenot:
Looks good.

PR #94, opened 2017-07-05
Use a random cookie name

### LINKED ISSUE #93: Use a random cookie name for Janitor
At the moment, when you want to contribute to janitor (through janitor), the docker host proxy will get confused, and use the same cookie and thus break the auth. This means that no static file will be served.

The current solution is to edit `lib/sessions.js` to change the cookie name on the following lines:
- https://github.com/JanitorTechnology/janitor/blob/135c06adc3a64fabba84c21eac9b62c76b24a2ec/lib/sessions.js#L49
- https://github.com/JanitorTechnology/janitor/blob/135c06adc3a64fabba84c21eac9b62c76b24a2ec/lib/sessions.js#L65

Doing this will allow you to run the janitor app using `node app` in the container's terminal, and access it on `https://moz1.janitor.technology/<container>/8080/`

If we were to use a random cookie name in those two lines, and store that cookie name in the DB for future uses, that should solve the issue.

magopian:
Fixes #93 

I didn't use [proper really random values](https://developer.mozilla.org/en/docs/Web/API/RandomSource/getRandomValues) because it seemed overkill, I can change that if needed ;)

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -67,16 +67,3 @@ function get (key, defaultValue) {
 }
 
 exports.get = get;
-
-
-// Set an entry in the datastore.
-
-function set (key, value) {
-
-  store[key] = value;
-
-  save();
-
-}
-
-exports.set = set;
--- a/lib/db.js
+++ b/lib/db.js
@@ -67,3 +67,16 @@ function get (key, defaultValue) {
 }
 
 exports.get = get;
+
+
+// Set an entry in the datastore.
+
+function set (key, value) {
+
+  store[key] = value;
+
+  save();
+
+}
+
+exports.set = set;
--- a/lib/sessions.js
+++ b/lib/sessions.js
@@ -11,6 +11,18 @@ const login = new emaillogin({
 });
 const useSecureCookies = !db.get('security').forceInsecure;
 
+// Get the cookiename saved from a previous session if any, or set a random one.
+const cookieInfo = db.get('cookieInfo');
+let cookieName;
+if (!cookieInfo['name']) {
+  // Generate a random cookie name made of letters 'a' to 'z'/
+  cookieName = Math.random().toString(36).replace(/[^a-z]+/g, 'a');
+  // Set the value in the db by providing it as default value to the `get` helper.
+  db.set('cookieInfo', {name: cookieName});
+} else {
+  cookieName = cookieInfo.name;
+}
+
 // Create a new session with a unique token.
 exports.create = function (callback) {
   login.login((error, token, session) => {
@@ -46,7 +58,7 @@ exports.get = function (request, callback) {
       }
 
       if (request.cookies) {
-        request.cookies.set('token', token, {
+        request.cookies.set(cookieName, token, {
           expires: new Date('2038-01-19T03:14:07Z'),
           secure: useSecureCookies
         });
@@ -62,7 +74,7 @@ exports.destroy = function (request, callback) {
   exports.get(request, (error, session, token) => {
     if (request.cookies) {
       // Destroy the cookie.
-      request.cookies.set('token', '', {
+      request.cookies.set(cookieName, '', {
         overwrite: true,
         secure: useSecureCookies
       });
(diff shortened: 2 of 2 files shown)


lib/sessions.js
```diff
@@ -11,6 +11,19 @@ const login = new emaillogin({
...
+let cookieName;
+if (!cookieInfo['name']) {
+  // Generate a unique cookie name so that it doesn't clash when loading janitor
+  // inside a janitor container (see #93).
+  cookieName = String(Date.now());
+  // Set the value in the db by providing it as default value to the `get` helper.
+  db.set('cookieInfo', {name: cookieName});
+} else {
+  cookieName = cookieInfo.name;
+}
```
jankeromnes: Nit: You can simplify this if/else block like so:

```js
if (!cookieNames.token) {
  cookieNames.token = 'token-' + Date.now();
}
```

This will correctly modify the in-memory DB, so you don't need to do anything else. 😃 (Note: I won't modify the on-disk DB copy `./db.json` until you call `db.save()`, but you don't need to do that here).

lib/sessions.js
```diff
@@ -46,7 +59,7 @@ exports.get = function (request, callback) {
       }
 
       if (request.cookies) {
-        request.cookies.set('token', token, {
+        request.cookies.set(cookieName, token, {
```
jankeromnes: Nit: Please use the object you retrieved from the DB directly, e.g. `cookieNames.token`.

lib/sessions.js
```diff
@@ -62,7 +75,7 @@ exports.destroy = function (request, callback) {
   exports.get(request, (error, session, token) => {
     if (request.cookies) {
       // Destroy the cookie.
-      request.cookies.set('token', '', {
+      request.cookies.set(cookieName, '', {
```
jankeromnes: Nit: Same as above.

lib/sessions.js
```diff
@@ -11,6 +11,19 @@ const login = new emaillogin({
 });
 const useSecureCookies = !db.get('security').forceInsecure;
 
+// Get the cookiename saved from a previous session if any, or set a random one.
+const cookieInfo = db.get('cookieInfo');
+let cookieName;
+if (!cookieInfo['name']) {
+  // Generate a unique cookie name so that it doesn't clash when loading janitor
+  // inside a janitor container (see #93).
+  cookieName = String(Date.now());
```
jankeromnes: Nit: Actually, what about prefixing this value with something more human-readable, e.g. `'token-' + Date.now()`, to make debugging our cookies more transparent? Otherwise we might end up with many cookie names that are just numbers, and it will be hard to distinguish which is which.

lib/db.js
```diff
@@ -67,3 +67,16 @@ function get (key, defaultValue) {
...
+
+function set (key, value) {
+
+  store[key] = value;
+
+  save();
+
+}
+
+exports.set = set;
```
jankeromnes: Thanks a lot for implementing this new helper! I don't think you need to use it though, as modifying what you get from `db.get(key)` already modifies the DB. I'm not against this new helper, but we've never felt the need for it until now.

lib/sessions.js
```diff
@@ -11,6 +11,19 @@ const login = new emaillogin({
...
+  // Set the value in the db by providing it as default value to the `get` helper.
+  db.set('cookieInfo', {name: cookieName});
```
magopian: Yeah, I didn't realize I was changing the object itself, I guess I've spent too much time in functional programming/immutable languages ;)

Where should I do the `db.save()` then? Or will it be done automatically at some point?

jankeromnes: Haha yes, immutable languages will do that to your brain, and JS is very mutable, some would even say too mutable :P

Yes, `db.save()` will be called later at some point (in general, there is no need to `db.save()` on startup, because everything was either just loaded from disk or auto-generated by the server, and there aren't any user-made state changes to save yet.)

magopian:
I believe I've implemented your suggestions, thanks for taking the time to review this PR ;)

PR #96, opened 2017-07-05
Use setCustomValidity for form error messages

### LINKED ISSUE #71: Use HTML 5 form error message
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Constraint_validation#Constraint_API%27s_element.setCustomValidity()

magopian:
Fixes #71

I chose to set the error message on the first form input, and I'm not sure if that's a good idea. I've only tried that on two forms: the "join the alpha" and the "sign in" ones which are both made of just one input text and one submit button.

### DIFF
--- a/static/css/janitor.css
+++ b/static/css/janitor.css
@@ -179,8 +179,17 @@ h4 {
 }
 
 #invite-form > .form-control-feedback {
+  pointer-events: none;
   top: 0;
   right: 60px;
+}
+
+.form-control-feedback .glyphicon-ok {
+  color: #3c763d !important;
+}
+
+.form-control-feedback .glyphicon-remove {
+  color: #d9534f !important;
 }
 
 #invite-form input[type='email'] {
--- a/static/css/janitor.css
+++ b/static/css/janitor.css
@@ -175,27 +175,9 @@ h4 {
 /* Forms */
 
 .form-control-feedback {
-  background-color: transparent;
-  background-image: none;
-  border-color: transparent;
-  border-radius: 0;
-  box-shadow: none;
-  color: #337ab7;
-  cursor: pointer;
-  display: inline-block;
-  font-size: 14px;
-  font-weight: 400;
-  line-height: 1.42857143;
-  margin: 0;
-  padding: 0;
-  text-align: center;
-  touch-action: manipulation;
-  -webkit-user-select: none;
-  -moz-user-select: none;
-  -ms-user-select: none;
-  user-select: none;
-  vertical-align: baseline;
-  white-space: nowrap;
+  background: none;
+  border: none;
+  pointer-events: none;
 }
 
 #signup-form {
@@ -203,7 +185,6 @@ h4 {
 }
 
 #invite-form > .form-control-feedback {
-  pointer-events: none;
   top: 0;
   right: 60px;
 }
@@ -223,7 +204,6 @@ h4 {
 .ajax-form > .form-control-feedback {
   outline: none;
   padding: 1px;
-  pointer-events: none;
   top: 5px;
 }
 
@@ -267,12 +247,6 @@ h4 {
 
 .ajax-form:not(.has-error) > .form-control-feedback > .glyphicon-remove {
   display: none;
-}
-
-.ajax-form .btn-link {
-  margin: 0;
-  padding: 0;
-  vertical-align: baseline;
 }
 
 .ajax-form :invalid {
--- a/static/css/janitor.css
+++ b/static/css/janitor.css
@@ -198,6 +198,7 @@ h4 {
 .ajax-form > .form-control-feedback {
   outline: none;
   padding: 1px;
+  pointer-events: none;
   top: 5px;
 }
 
--- a/static/css/janitor.css
+++ b/static/css/janitor.css
@@ -249,9 +249,7 @@ h4 {
   vertical-align: baseline;
 }
 
-.ajax-form :invalid,
-.ajax-form :-moz-submit-invalid,
-.ajax-form :-moz-ui-invalid {
+.ajax-form :invalid {
   box-shadow: none;
 }
 
--- a/static/css/janitor.css
+++ b/static/css/janitor.css
@@ -174,14 +174,28 @@ h4 {
 
 /* Forms */
 
-#signup-form {
+.form-control-feedback {
+  background-color: transparent;
+  background-image: none;
+  border-color: transparent;
+  border-radius: 0;
+  box-shadow: none;
+  color: #337ab7;
+  cursor: pointer;
+  display: inline-block;
+  font-size: 14px;
+  font-weight: 400;
+  line-height: 1.42857143;
+  margin: 0;
   padding: 0;
-}
-
-#invite-form > .form-control-feedback {
-  pointer-events: none;
-  top: 0;
-  right: 60px;
+  text-align: center;
+  touch-action: manipulation;
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+  vertical-align: baseline;
+  white-space: nowrap;
 }
 
 .form-control-feedback .glyphicon-ok {
@@ -190,6 +204,16 @@ h4 {
 
 .form-control-feedback .glyphicon-remove {
   color: #d9534f !important;
+}
+
+#signup-form {
+  padding: 0;
+}
+
+#invite-form > .form-control-feedback {
+  pointer-events: none;
+  top: 0;
+  right: 60px;
 }
 
 #invite-form input[type='email'] {
--- a/static/css/janitor.css
+++ b/static/css/janitor.css
@@ -243,6 +243,12 @@ h4 {
   display: none;
 }
 
+.ajax-form .btn-link {
+  margin: 0;
+  padding: 0;
+  vertical-align: baseline;
+}
+
 .login-with {
   margin-bottom: 10px;
   display: flex;
--- a/static/css/janitor.css
+++ b/static/css/janitor.css
@@ -249,6 +249,12 @@ h4 {
   vertical-align: baseline;
 }
 
+.ajax-form :invalid,
+.ajax-form :-moz-submit-invalid,
+.ajax-form :-moz-ui-invalid {
+  box-shadow: none;
+}
+
 .login-with {
   margin-bottom: 10px;
   display: flex;
--- a/static/css/janitor.css
+++ b/static/css/janitor.css
@@ -174,19 +174,8 @@ h4 {
 
 /* Forms */
 
-.form-control-feedback {
-  background: none;
-  border: none;
-  pointer-events: none;
-}
-
 #signup-form {
   padding: 0;
-}
-
-#invite-form > .form-control-feedback {
-  top: 0;
-  right: 60px;
 }
 
 #invite-form input[type='email'] {
@@ -202,9 +191,16 @@ h4 {
 }
 
 .ajax-form > .form-control-feedback {
+  background: none;
+  border: none;
   outline: none;
+  pointer-events: none;
   padding: 1px;
   top: 5px;
+}
+
+.ajax-form > .form-control-feedback:invalid {
+  box-shadow: none;
 }
 
 .ajax-form.has-button > .form-control-feedback {
@@ -233,6 +229,11 @@ h4 {
   left: -30px;
 }
 
+#invite-form > .form-control-feedback {
+  top: 0;
+  right: 60px;
+}
+
 .row > .ajax-form > .form-control-feedback {
   right: 15px;
 }
@@ -247,10 +248,6 @@ h4 {
 
 .ajax-form:not(.has-error) > .form-control-feedback > .glyphicon-remove {
   display: none;
-}
-
-.ajax-form :invalid {
-  box-shadow: none;
 }
 
 .login-with {
--- a/static/css/janitor.css
+++ b/static/css/janitor.css
@@ -198,14 +198,6 @@ h4 {
   white-space: nowrap;
 }
 
-.form-control-feedback .glyphicon-ok {
-  color: #3c763d !important;
-}
-
-.form-control-feedback .glyphicon-remove {
-  color: #d9534f !important;
-}
-
 #signup-form {
   padding: 0;
 }
--- a/static/js/janitor.js
+++ b/static/js/janitor.js
@@ -158,7 +158,14 @@ function setupAsyncForm (form) {
 
 // Update the visual feedback of a <form>'s status.
 function updateFormStatus (form, status, message) {
+  // Get the targetted form element which should receive the form feedback.
+  const targetElement = form.querySelector('.form-feedback-target');
+
   form.classList.remove('has-success', 'has-error');
+  // Reset the custom validity message so the element isn't invalid anymore,
+  // and the form can be submitted.
+  targetElement.setCustomValidity('');
+
   switch (status) {
     case 'success':
       form.classList.add('has-success');
@@ -176,7 +183,6 @@ function updateFormStatus (form, status, message) {
   var feedback = form.querySelector('.form-control-feedback');
   if (message && feedback) {
     // Set a custom validation message on the target form element.
-    const targetElement = form.querySelector('.form-feedback-target');
     targetElement.setCustomValidity(message);
     // Force the display of thecustom validity message.
     form.reportValidity();
--- a/static/js/janitor.js
+++ b/static/js/janitor.js
@@ -198,7 +198,7 @@ function updateFormStatus (form, status, message) {
 // Add visual feedback elements to a given <form>.
 function addFormFeedback (form) {
   var feedback = document.createElement('button');
-  feedback.classList.add('form-control-feedback', 'btn','btn-link');
+  feedback.classList.add('form-control-feedback');
   feedback.setAttribute('tabindex', '99');
 
   // Append icons for 'success' and 'error' states.
--- a/static/js/janitor.js
+++ b/static/js/janitor.js
@@ -175,10 +175,11 @@ function updateFormStatus (form, status, message) {
 
   var feedback = form.querySelector('.form-control-feedback');
   if (message && feedback) {
-    // Set a custom validation message on the first input element of the form.
-    form.elements[0].setCustomValidity(message);
-    // Fire a submit event so the custom validation messages display.
-    form.querySelectorAll("input[type=submit]")[0].click();
+    // Set a custom validation message on the target form element.
+    const targetElement = form.querySelector('.form-feedback-target');
+    targetElement.setCustomValidity(message);
+    // Force the display of thecustom validity message.
+    form.reportValidity();
   }
 
   if (form.dataset.refreshAfterSuccess && status == 'success') {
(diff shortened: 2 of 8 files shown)


jankeromnes:
Thank you for the pull request! 😃 I will take a look.

static/js/janitor.js
```diff
@@ -201,16 +202,6 @@ function addFormFeedback (form) {
...
 
-  // Set-up the feedback message box (a bootstrap popover).
-  $(feedback).popover({
-    content: function () {
-      return this.dataset.message;
-    },
-    container: 'body',
-    placement: 'bottom',
-    trigger: 'focus'
-  });
```
jankeromnes: Removing this would be quite nice! It's one of the very few jQuery dependencies we have (the other one is for `timeago`, a jQuery plugin that shows our fuzzy timestamps like "3 days ago" or "a few seconds ago").

jankeromnes:
@magopian If you'd like to try this change on forms with more input fields, you can try the following steps:
- Run `node app` in the Terminal
- Access the Janitor web preview (a URL like `https://moz1.janitor.technology/abcdef123456/8080/`)
- Try signing in as `admin@localhost`
- In the terminal, a link is printed with a login key like `?key=[...]` (because you're in `insecure` mode)
- Copy this login key (beware, the Terminal may display it on multiple lines because it's quite big)
- Paste it at the end of your Janitor web preview URL (like so: `https://moz1[...]/8080/?key=[...]`) and hit `Enter` to log in as a website admin
- You can now try to use any form in `https://moz1[...]/8080/settings/configurations/` (newer, `window.fetch`-based forms) and in `https://moz1[...]/8080/settings/account/` (mostly older, `Scout/Ajax/XHR`-based forms, like the Cloud9 SSH key form) to see what happens with respect to feedback messages.

static/js/janitor.js
magopian: I'm all for getting rid of jquery ;) Also, there's libraries like https://www.npmjs.com/package/time-ago that might do the trick.

static/js/janitor.js
```diff
@@ -175,8 +175,10 @@ function updateFormStatus (form, status, message) {
 
   var feedback = form.querySelector('.form-control-feedback');
   if (message && feedback) {
-    feedback.dataset.message = message;
-    feedback.focus();
+    // Set a custom validation message on the first input element of the form.
+    form.elements[0].setCustomValidity(message);
```
magopian: Indeed!

static/js/janitor.js
```diff
@@ -175,8 +175,10 @@ function updateFormStatus (form, status, message) {
...
+    // Fire a submit event so the custom validation messages display.
+    form.querySelectorAll("input[type=submit]")[0].click();
```
magopian: More and more it seems to me the `setCustomValidity` solution isn't the right tool for the job :/ We might be able to fix all those issues, but I worry that the final implementation will be contrived. Let's see how it goes ;)

magopian:
Ok, I'm trying to gather a few thoughts here:
1/ the current form "validation" (feedback) implementation only works for a single feedback per form, and the feedback message isn't attached to any one form element. `setCustomValidity` on the other hand has to be set on a given form element
2/ setting a custom validity message on an element will make it invalid straight away (giving it a red border by default on FF), but won't display the message until the user hovers it or the form submit button is clicked (manually or via javascript)
3/ setting a validity message will make the element invalid, and will thus prevent any further form submission (unless the message is set to the empty string)

The use case for the current popover are:
1/ error messages like "this address already exist" or "signing in requires an invitation"
2/ feedback messages like reporting back that everything went well
The message is then displayed on focus (of that feedback div).


I believe most (all?) of the issues can be fixed using workarounds of some sorts, I'm just not sure that it's a good idea. Thoughts @jankeromnes @nt1m ?

nt1m:
>  1/ the current form "validation" (feedback) implementation only works for a single feedback per form, and the feedback message isn't attached to any one form element. setCustomValidity on the other hand has to be set on a given form element

This allows getting rid of the custom bootstrap form feedback icon, and instead have something as simple as:

```css
input:invalid {
  background-image: url(cross.svg);
}
```
> 2/ setting a custom validity message on an element will make it invalid straight away (giving it a red border by default on FF), but won't display the message until the user hovers it or the form submit button is clicked (manually or via javascript)

I don't mind triggering a second submit event as long as it does nothing but add the message.

>  3/ setting a validity message will make the element invalid, and will thus prevent any further form submission (unless the message is set to the empty string)

That shouldn't be too much extra code ?

> I'm just not sure that it's a good idea.

I still think it's a good idea, since we can finally use standard form feedback indicators (including the input:invalid selector) and get rid of some Jquery. Also allows us to move away from bootstrap which is something I'd like to do.

magopian:
@nt1m I forgot to add another issue I'm currently struggling with:
- some forms only have a button: I need to make sure the button can have a custom validity message
- some forms don't even have a button, but just a custom link: what should we do in this case? I don't think you can set a custom validity message on something else than a form element ((HTMLInputElement, HTMLSelectElement, HTMLButtonElement or HTMLTextAreaElement)

Also, setting a custom validity message will display the element as invalid, even if the element isn't really invalid (eg we're just trying to display a success feedback message). In that case we'd have the green check glyphicon but a red border.

What do you think @nt1m ?

nt1m:
@magopian hmm, I guess you could try styling a read-only input as a link or a button, or overlay an invisible input on top. Not sure.

magopian:
Ok, let's try this, even if I really think this is a bad idea:
- [ ] add a `form-feedback` class to the form element that should have the custom validity message
- [ ] use `setCustomValidity` on the form element that has the `form-feedback` class
- [ ] fake a form submission to display the custom validity message
- [ ] restyle the `:invalid` pseudo-class so it doesn't have the red box-shadow (or make it green if the feedback message is a success)
- [ ] set the custom validity message back to the empty string (disable the `:invalid` state) on focus so the form can be resubmitted
- [ ] change custom links to be submit buttons styled as simple links so they benefit from the same treatment as the other forms

magopian:
Ok, I guess this is now ready for review, I hope I haven't missed any form ;)

The caveat is that now each and every new form that needs feedback needs a form element (one of HTMLInputElement, HTMLSelectElement, HTMLButtonElement or HTMLTextAreaElement) to have the class `form-feedback-target` so the code knows where to `setCustomValidity`.

nt1m:
I was convinced we could remove more code than this. Right now, there's some extra code that handles the error status icon, I think we can now remove that and rely on the ":invalid" pseudo-selector.

static/css/janitor.css
```diff
@@ -243,6 +243,18 @@ h4 {
...
 
+.ajax-form .btn-link {
+  margin: 0;
+  padding: 0;
+  vertical-align: baseline;
+}
+
+.ajax-form :invalid,
+.ajax-form :-moz-submit-invalid,
+.ajax-form :-moz-ui-invalid {
```
nt1m: just :invalid should be enough

static/js/janitor.js
```diff
@@ -190,7 +199,6 @@ function updateFormStatus (form, status, message) {
 function addFormFeedback (form) {
   var feedback = document.createElement('div');
   feedback.classList.add('form-control-feedback');
-  feedback.dataset.message = '';
   feedback.setAttribute('tabindex', '99');
```
nt1m: I'm thinking we simplify/remove the code that adds the feedback icon by using some simple CSS:

input:invalid {
  background-image: url(cross.svg);
}

magopian: we would need a bit more than that sadly: the feedback div can't have the `setCustomValidity` message (it only applies to form elements). And the custom validity message can be on various elements, depending on the forms. Sometimes it's inside a textarea, sometimes it's at the right of the submit button that's styled as a link, sometimes it's before a submit button (styled as a standard button), and I might miss other cases.

I'll try adding the necessary CSS rules to do that, and will report back.

static/js/janitor.js
```diff
@@ -158,7 +158,14 @@ function setupAsyncForm (form) {
 
 // Update the visual feedback of a <form>'s status.
 function updateFormStatus (form, status, message) {
+  // Get the targetted form element which should receive the form feedback.
+  const targetElement = form.querySelector('.form-feedback-target');
```
jankeromnes: Nit: Why force every `form` element to have a `input.form-feedback-target` element, when we already inject a `'div.form-control-feedback'`? Couldn't we simply insert a customValidity target there? (If it has to be an input, maybe we can add a hidden input, or transform the current `span.glyphicon` elements input inputs?) Just asking, because it would make the form markup simpler and more robust.

static/js/janitor.js
```diff
@@ -175,8 +182,10 @@ function updateFormStatus (form, status, message) {
 
   var feedback = form.querySelector('.form-control-feedback');
   if (message && feedback) {
-    feedback.dataset.message = message;
-    feedback.focus();
+    // Set a custom validation message on the target form element.
+    targetElement.setCustomValidity(message);
+    // Force the display of thecustom validity message.
+    form.reportValidity();
```
jankeromnes: This is much better than a fake follow-up submit! 👍

static/js/janitor.js
nt1m: You could do something like:

input:invalid + .feedback-icon {
 ...
}

to handle all possible elements.

jankeromnes: Note: Some forms don't show these icons in input elements (for example, button forms like "Reset to Default" show it on the left of the button, and other forms like "deploy configuration" show it inside the bootstrap alert). Keep in mind that some forms don't have inputs at all.

magopian: What i meant is that if we remove the feedback divs, we need to replace them by css rules (putting the feedback icon before, after, inside...) for each case. It's not a big deal, i'll try doing that now.

static/js/janitor.js
magopian: It is... somewhat: the support isn't very good (only chrome and ff): https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/reportValidity

static/js/janitor.js
magopian: the `.form-feedback-target` is set on one of the form elements (could be an input, a button, a textarea). I can try adding a "fake" input to the the feedback div, but if I understand correctly, @nt1m would like to get rid of it entirely. What should I do?

jankeromnes:
Thanks a lot for working on this!

I think that if we decide to tweak the feedback mechanism, we should do it right:
- Either find a standard, drop-in replacement for the current mechanism that simplifies code but doesn't tweak too many things (e.g. finding a way to show `.customValidity(message)` on the existing `.form-control-feedback` div that we inject everywhere)
- Or we go all the way and come up with a much better way to show feedback on async behavior (e.g. like MailChimp buttons, that animate when a call is pending, and show feedback in smart ways). But this option no longer qualifies as a `Good First Issue` as it implies rewriting big parts of the web UI.

I don't think it's worth doing anything in between (as in, solutions that go half-way by tweaking things in complex ways but not producing optimal results).

static/js/janitor.js
jankeromnes: That's unfortunate 😐 we try to be as compatible as possible without doing anything too expensive. In this case, if we find a simple and lean polyfill that Just Works™️, we could use that, otherwise I'd prefer to shy away from anything that's not too well supported across all browsers.

static/js/janitor.js
jankeromnes: I think removing the `.form-control-feedback` element entirely corresponds to point 2 of https://github.com/JanitorTechnology/janitor/pull/96#issuecomment-313671183.

If you're really up for modernizing the feedback UX for the whole website, I think a better approach would be to work on a prototype UI where we can iterate faster on the UX of feedback for async buttons, forms, etc.

Otherwise, I'd prefer going with point 1, with minimal change and impact to the existing mechanism (except using something more standard), but nothing in between please. Yoda would say "do, or do not". 😛

magopian:
I tried another implementation: I changed the form feedback div to be a button instead (so it's a form element), and set the custom validity message on it.

According to my tests, it seems to be working well, not sure if this is looking better to you @jankeromnes ?

static/js/janitor.js
```diff
@@ -201,14 +206,9 @@ function addFormFeedback (form) {
...
 
-  // Set-up the feedback message box (a bootstrap popover).
-  $(feedback).popover({
-    content: function () {
-      return this.dataset.message;
-    },
-    container: 'body',
-    placement: 'bottom',
-    trigger: 'focus'
+  // Clicking on the form feedback "button" shouldn't submit the form.
```
nt1m: You could also add pointer-events: none; on the button to prevent clicks on the button

magopian: done in https://github.com/JanitorTechnology/janitor/pull/96/commits/0350af689d7bfbdda8f547005b481980203c1a7a (thanks for the "pointer" (haha), I didn't know about this one)

nt1m: :D You can also remove the event listener that only does e.preventDefault(); now

static/js/janitor.js
```diff
@@ -174,9 +174,15 @@ function updateFormStatus (form, status, message) {
   }
 
   var feedback = form.querySelector('.form-control-feedback');
+
+  // Reset the custom validity message so the element isn't invalid anymore,
+  // and the form can be submitted.
+  feedback.setCustomValidity('');
```
nt1m: You need to check if feedback exists before calling this.

static/js/janitor.js
magopian: Oh, right, didn't understand it would make it redundant, thanks.

nt1m:
Looks pretty good to me!

Some things I've noticed:
- The "x" icon turns blue when the popup is opened
- The panel is a bit off, but I think that's a bug in Firefox, since the button is perfectly positioned and sized

image

magopian:
I never had a blue icon, how did you do that?

Regarding the panel positioning, I have no clue how to change it or have any effect on it, do you?

nt1m:
@magopian 
> I never had a blue icon, how did you do that?

It just turns blue every time the popup shows (I'm using Firefox on Mac). Adding an !important statement of each color rule did prevent that from happening.

> Regarding the panel positioning, I have no clue how to change it or have any effect on it, do you?

This looks like a Firefox bug, but would just ignore it for now.

magopian:
I'm also using Firefox on Mac, never had this icon appear blue :/
What did you set in the textarea to have this error message? I can't seem to have a cross, only the (always green) checkmark.

magopian:
Ok @nt1m I gave it a try, can you please confirm that it's now using the correct color?

nt1m:
@magopian 

> What did you set in the textarea to have this error message? 

Just open that page, then kill the server, and try to save any field.

> Oh, wait, just got it, but no clue how to change it, I believe it's from bootstrap's styles? I'll try styling with :invalid

So showing the validity seems to cause :focus styles to apply. And `.btn-link:focus {}` has a special color.
I prefer that you apply your own rules on .form-control-feedback instead of using .btn-link .

templates/settings-configurations.html
```diff
@@ -19,7 +19,7 @@
           </form>
           <form action="/api/user/configurations/{{= file in uri}}" class="account-form ajax-form has-feedback col-sm-12" method="put">
             <p class="alert alert-warning">
-              <strong>Note:</strong> This configuration will only apply to new containers. Overwrite <strong>~/{{= file in html}}</strong> in all existing containers by clicking <a data-submit-on="click" href="javascript:void(0)">here</a>.
+              <strong>Note:</strong> This configuration will only apply to new containers. Overwrite <strong>~/{{= file in html}}</strong> in all existing containers by clicking <button class="btn btn-link" data-submit-on="click" type="submit">here</button>.
```
jankeromnes: Nit: Why change this? Also, if you really wanted to do this, it's not necessary to use `data-submit-on="click"` because you already set `type="submit"`.

magopian: ah, you're right, it's not needed anymore, at some point i implemented the `setCustomValidity` on one of the existing form elements, so it needed a form element (and a link isn't one). I can change that back to what it was now ;)

magopian:
> I prefer that you apply your own rules on .form-control-feedback instead of using .btn-link .

I don't understand why, as it would be duplicating the CSS rules, but I'm not going to argue. I've added a new commit doing just that.

jankeromnes:
Thanks a lot for fixing this up! It seems to work almost perfectly now.

I was about to complain about a regression on the signup form on the index page: The feedback icon is not vertically centered (but @nt1m pointed out that this problem was already present before your change, so you can simply ignore it).

> So showing the validity seems to cause :focus styles to apply. And .btn-link:focus {} has a special color.

That's correct, we hide the `.form-control-feedback` element by default, and only show it when it's focused. That's how we manage to dismiss the feedback message when the user clicks somewhere else. (And it's impossible to focus it by mistake, except if you `Tab` through all the elements of the page, or focus it via Devtools).

> I prefer that you apply your own rules on `.form-control-feedback` instead of using `.btn-link`.

I would prefer this too, if it's not too much trouble, because it's cleaner than to re-use a somewhat unrelated bootstrap class only to re-tweak it afterwards.

static/css/janitor.css
```diff
@@ -174,11 +174,36 @@ h4 {
...
+  margin: 0;
+  padding: 0;
+  text-align: center;
+  touch-action: manipulation;
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+  vertical-align: baseline;
+  white-space: nowrap;
```
nt1m: I've reduced this enormous list of rules to:
```css
.form-control-feedback {
  background: none;
  border: none;
  pointer-events: none;
}
```

static/css/janitor.css
```diff
@@ -174,11 +174,36 @@ h4 {
...
+}
+
 #signup-form {
   padding: 0;
 }
 
 #invite-form > .form-control-feedback {
+  pointer-events: none;
```
nt1m: You can remove this if it's set in the global `.form-control-feedback` rule above.

static/css/janitor.css
```diff
@@ -198,6 +223,7 @@ h4 {
 .ajax-form > .form-control-feedback {
   outline: none;
   padding: 1px;
+  pointer-events: none;
```
nt1m: Same here

static/css/janitor.css
```diff
@@ -243,6 +269,16 @@ h4 {
   display: none;
 }
 
+.ajax-form .btn-link {
```
nt1m: This ruleset isn't used anywhere, please remove it.

nt1m:
@magopian I've addressed my own nits.

static/css/janitor.css
```diff
@@ -174,6 +174,12 @@ h4 {
 
 /* Forms */
 
+.form-control-feedback {
+  background: none;
+  border: none;
+  pointer-events: none;
```
jankeromnes: Wow, that's much shorter than previously. Thanks for simplifying these rules!

Nit: However, please add them with the other `.form-control-feedback` rules below (see `.ajax-form > .form-control-feedback`).

static/css/janitor.css
```diff
@@ -243,6 +249,10 @@ h4 {
   display: none;
 }
 
+.ajax-form :invalid {
+  box-shadow: none;
```
jankeromnes: Nit: Why remove the `box-shadow` of every `:invalid` element inside an `.ajax-form`? Wouldn't it make more sense to do this only for `.ajax-form > .form-control-feedback:invalid`?

PR #97, opened 2017-07-05
Use a random cookie name 2/2.

jankeromnes:
Fix a forgotten hard-coded "token" cookie name that was left over from #94.


jankeromnes:
Pretty straight-forward fix, merging without review.

PR #98, opened 2017-07-06
Accept '-' in project IDs for project page URLs.

jankeromnes:
This would allow accessing https://janitor.technology/projects/firefox-hg/ (currently 404).

### DIFF
--- a/app.js
+++ b/app.js
@@ -111,7 +111,7 @@ boot.executeInParallel([
   });
 
   // Public project pages.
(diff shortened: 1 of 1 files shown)


magopian:
This would also accept IDs starting and/or ending with a `-`. If you're ok with that, it looks good to me 👍

jankeromnes:
Thanks! IDs starting/ending with a `-` should be OK. 😃

PR #101, opened 2017-07-07
Enforce a style guide

### LINKED ISSUE #77: Enforce a consistent coding style by adding a linter
In order to guarantee a consistent coding style, and prevent some types of coding errors to enter the code base, we should:
- [x] Automatically lint the server code (Node.js)
- [ ] Automatically lint the website code (ES5 JavaScript, HTML, CSS)

Here are a few guidelines that I think should be respected:
- We should write very short but very simple code (no unreasonably convoluted code and no black magic)
- We should use explicit, full variable names (e.g. not `c` but `chunk`, not `req` but `request`, not `values` but `durations` etc.)
- We should sort most things alphabetically (requires, variables, parameters, HTML attributes, etc) unless it makes things more confusing

Here are a few tools we could use:
- [eslint](https://github.com/eslint/eslint): A great JS linter, similar to JSLint and JSHint (we would need to write a coding style configuration for Janitor, or auto-detect it from our code)
- [standard](https://github.com/feross/standard): Basically ESLint with powerful, pre-configured rules and no customization possible (it is opinionated but powerful and convenient. Maybe the most controversial rule it enforces is "no semicolons")
- [flow](https://github.com/facebook/flow): A static JS type checker made by Facebook.

I would be OK to enforce `standard` for Janitor, even if it removes all semicolons. @bnjbvr @nt1m @Coder206 @notriddle or @espadrine, do you have any thoughts or preferences about linting our code?

magopian:
Fixes #77

To run eslint:
- for node: `npm run lint-node` (use `npm run lint-node-fix` to auto fix some issues)
- for the browser: `npm run lint-browser` (use `npm run lint-browser-fix` to auto fix some issues)
- for both: `npm run lint` (use `npm run lint-fix` to auto fix some issues)


magopian:
Here's a run for the node files: https://dpaste.de/cay1

.eslintrc-browser.js
```diff
@@ -0,0 +1,69 @@
+module.exports = {
+  "env": {
+    "browser": true,
+    "es6": false
+  },
+  "extends": [
+    "eslint:recommended",
+    "plugin:node/recommended",
+    "standard"
```
jankeromnes: Interesting, what are these `eslint:recommended` and `plugin:node/recommended` rules? Also, are they not already included in `standard`?

.eslintrc-browser.js
```diff
@@ -0,0 +1,69 @@
...
+  ],
+  "parserOptions": {
+    "ecmaVersion": "2017",
+    "ecmaFeatures": {
+      "experimentalObjectRestSpread": true,
+      "jsx": true
+    },
+    "sourceType": "module"
```
jankeromnes: I don't get these parser options for the browser environment. Since we don't use ES6, why add `"ecmaVersion": "2017"`? Also, I don't believe we do any experimental object rest spread or JSX here (which is React's non-standard JS format).

.eslintrc-browser.js
```diff
@@ -0,0 +1,69 @@
...
+  },
+  "plugins": [
+    "node"
+  ],
+  "rules": {
+    // Standard js rules
```
jankeromnes: Nit: These are actually rules that contradict standard.

.eslintrc-browser.js
```diff
@@ -0,0 +1,69 @@
...
+    "semi": ["error", "always"],
+    "comma-dangle": ["error", "never"],
```
jankeromnes: Nit: We try to use dangling commas on multiline objects and arrays. Please use something like `"comma-dangle": ["error", "only-multiline"]` (or you may want to explicitly list all [options](http://eslint.org/docs/rules/comma-dangle#options) in order to special-case `"functions": "never"`, because it may break older browsers).

.eslintrc-browser.js
```diff
@@ -0,0 +1,69 @@
...
+    
+    // Node require rules
+    "node/exports-style": ["error", "module.exports"],
```
jankeromnes: We don't use `exports` or `module.exports` in the browser.

.eslintrc-browser.js
```diff
@@ -0,0 +1,69 @@
...
+
+    // overriding recommended rules
+    "no-constant-condition": ["error", { checkLoops: false }],
+    "no-console": [ "error", { allow: ["log", "warn", "error"] } ],
+    "no-unused-vars": ["error", { "argsIgnorePattern": "^_" }],
```
jankeromnes: Interesting, why did you add these?

.eslintrc-browser.js
```diff
@@ -0,0 +1,69 @@
...
+    "no-useless-constructor": "error",
+    "no-useless-rename": "error",
+    "no-useless-return": "error",
+    "no-var": "off",
+    "no-void": "error",
+    "no-with": "error",
+    "prefer-const": "error",
+    "prefer-promise-reject-errors": "error",
+    "prefer-rest-params": "error",
+    "prefer-spread": "error",
```
jankeromnes: Why did you add all these? Aren't they examples already present in `"standard"`? Also a few of them are ES6+ and don't make sense in the browser.

.eslintrc-node.js
```diff
@@ -0,0 +1,78 @@
+module.exports = {
+  "env": {
+    "browser": true,
```
jankeromnes: Why set `"browser": true` in server files?

package.json
```diff
@@ -37,7 +37,15 @@
     "selfapi": "~0.0.16",
     "tar-stream": "~1.5.4"
   },
-  "engines" : {
-    "node" : ">=6.0.0"
+  "engines": {
+    "node": ">=6.0.0"
```
jankeromnes: Thank you for fixing this bogus whitespace!

.eslintrc-browser.js
magopian: I believe those are base rules that are "cascading"? Here's the `eslint:recommended`: http://eslint.org/docs/rules/

I'm not sure which is where, I just feel it's sensible to do it like that, but we could very well just keep `standard` I guess?

.eslintrc-browser.js
magopian: not in the browser, as discussed on irc. Or did I misunderstand?

.eslintrc-browser.js
magopian: mmmm, copy pasted from the html profiler devtools, i'll remove them.

.eslintrc-browser.js
magopian: same as above

magopian:
I added some automatic git hooks that will run the linting before each commit and push. You'll need to run `npm install` to activate them.

When commiting or pushing you should see something like this (check the little spinner next the "lint-browser-fix" line):

screen shot 2017-07-07 at 16 55 54

If the checks fail, you'll be prompted with the following message:

```husky > pre-commit hook failed (add --no-verify to bypass)```

So if you wish to bypass the checks, simply use `git commit --no-verify ....`

File: lib/configurations.js [hunk 10]
```diff
@@ -10,23 +10,23 @@ const db = require('./db');
...
+  '.emacs',
+  '.eslintrc',
+  '.gdbinit',
+  '.gitconfig',
+  '.gitignore',
+  '.hgrc',
+  '.nanorc',
+  '.netrc',
+  '.ssh/config',
+  '.vimrc',
```
jankeromnes: Heh, @nt1m is not going to like this 😛

nt1m: Noooooooooo

static/js/bootstrap-3.3.7.min.js
```diff
@@ -3,5 +3,8 @@
...
...&&e[c]()})}b.VERSION="3.3.7",b.DEFAULTS={offset:10},b.prototype.getScrollHeight=function(){return this.$scrollElement[0].scrollHeight||Math.max(this.$body[0].scrollHeight,document.documentElement.scrollHeight)},b.prototype.refresh=function(){var b=this,c="offset",d=0;this.offsets=[],this.targets=[],this.scrollHeight=this.getScrollHeight(),a.isWindow(this.$scrollElement[0])||(c="position",d=this.$scrollElement.scrollTop()),this.$body.find(this.selector).map(function(){var b=a(this),e=b.data("target")||b.attr("href"),f=/^#./.test(e)&&a(e);return f&&f.length&&f.is(":visible")&&[[f[c]().top+d,e]]||null}).sort(function(a,b){return a[0]-b[0]}).each(function(){b.offsets.push(this[0]),b.targets.push(this[1])})},b.prototype.process=function(){var a,b=this.$scrollElement.scrollTop()+this.options.offset,c=this.getScrollHeight(),d=this.options.offset+c-this.$scrollElement.height(),e=this.offsets,f=this.targets,g=this.activeTarget;if(this.scrollHeight!=c&&this.refresh(),b>=d)return g!=(a=f[f.length-1])&&this.activate(a);if(g&&b<e[0])return this.activeTarget=null,this.clear();for(a=e.length;a--;)g!=f[a]&&b>=e[a]&&(void 0===e[a+1]||b<e[a+1])&&this.activate(f[a])},b.prototype.activate=function(b){
```
nt1m: Can you leave libs unchanged and add a .eslintignore file including common library patterns ?

magopian: What kind of "common library patterns"? Keep in mind I barely know any JS ;)

magopian:
Added an `.eslintignore` file, and ignored all files ending with `.min.js`. There aren't many errors left, and the linting step before commiting/pushing is not really short (less than a second on my machine), which is imho totally acceptable.

.eslintrc-node.js
```diff
@@ -0,0 +1,28 @@
...
+  "rules": {
+    // Override some of standard js rules
+    "semi": ["error", "always"],
+    "comma-dangle": [
+      "error", {
+        "arrays": "only-multiline",
+        "objects": "only-multiline",
+        "imports": "never",
+        "exports": "never",
+        "functions": "never",
```
jankeromnes: Nit: These rules should be in the browser rules, not in the Node.js/server rules. In the node rules, it should be `'command-dangle': ['error', 'only-multiline']`.

api/hosts-api.js
```diff
@@ -149,7 +149,6 @@ hostAPI.post({
     // No authentication.
     response.statusCode = 403; // Forbidden
     response.json({ error: 'Unauthorized' }, null, 2);
-    return;
```
jankeromnes: Interesting. This `return` was indeed unnecessary, but I added it as an indication that there should be no further instructions in this block (below, there are some inline functions, but there shouldn't be any other code or logic that risks calling `response.json()` or `response.end()` a second time). This is just a random remark, I'm fine with this change.

api/user-api.js
```diff
@@ -154,7 +154,7 @@ configurationsAPI.patch({
       ], null, 2)
     },
     response: {
-      body: JSON.stringify({ '.gitconfig': '[user]\nname = Janitor'  }, null, 2)
+      body: JSON.stringify({ '.gitconfig': '[user]\nname = Janitor' }, null, 2)
```
jankeromnes: Nice catch! 😄

lib/certificates.js
```diff
@@ -485,5 +446,4 @@ function requestLetsEncryptCertificate (parameters, callback) {
   };
 
   client.getCertificate(options, callback);
-
```
jankeromnes: Huge 👍! I used to do all these changes by hand... You're saving me so much time!

package.json
```diff
@@ -22,7 +22,13 @@
...
     "start": "if [ -z \"$SCRIPT\" ] ; then printf \"Run which Janitor script? [join/app]:\" && read SCRIPT ; fi ; node \"$SCRIPT\" >> janitor.log 2>&1 & printf \"$!\\n\" > janitor.pid",
     "poststart": "printf \"[$(date -uIs)] Background process started (PID $(cat janitor.pid), LOGS $(pwd)/janitor.log).\\n\"",
-    "stop": "if [ -e janitor.pid -a -n \"$(ps h $(cat janitor.pid))\" ] ; then kill $(cat janitor.pid) && printf \"[$(date -uIs)] Background process stopped (PID $(cat janitor.pid)).\\n\" ; fi ; rm -f janitor.pid"
+    "stop": "if [ -e janitor.pid -a -n \"$(ps h $(cat janitor.pid))\" ] ; then kill $(cat janitor.pid) && printf \"[$(date -uIs)] Background process stopped (PID $(cat janitor.pid)).\\n\" ; fi ; rm -f janitor.pid",
+    "lint-node": "eslint -c .eslintrc-node.js *.js api/ lib/",
+    "lint-node-fix": "npm run lint-node -- --fix",
+    "lint-browser": "eslint -c .eslintrc-browser.js static/",
+    "lint-browser-fix": "npm run lint-browser -- --fix",
+    "precommit": "lint-staged",
+    "prepush": "lint-staged"
```
jankeromnes: Actually, I'm not fond of commit hooks, and @nt1m is worried about them too. Do you mind removing them from this pull request, and maybe suggest them in a follow-up pull request if you really think they're really valuable?

static/js/graphs.js
```diff
@@ -23,16 +23,13 @@ Array.map(document.querySelectorAll('*[data-data]'), function (div) {
         includeZero: true
       }
     },
-    labelsUTC: true,
+    labelsUTC: true
```
jankeromnes: Nit: I believe we should allow dangling commas for multi-line objects and arrays in browser code.

static/js/bootstrap-3.3.7.min.js
nt1m: static/js/*.min.js as you figured out :)

static/js/graphs.js
```diff
@@ -50,23 +47,4 @@ function formatTime (milliseconds) {
...
-  var prefix = ['', 'K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y'];
-  var p = 0;
-  var value = Number(bytes);
-
-  while (value > 1024 && p < prefix.length) {
-    value /= 1024;
-    p++;
-  }
-
-  return (Math.round(value * 100) / 100) + ' ' + prefix[p] + 'B';
```
jankeromnes: Please don't delete this function! I know it's not used yet, but we might want to use it in the future.

static/js/janitor.js
```diff
@@ -99,7 +101,7 @@ function fetchAPI (method, url, data, callback) {
   }).then(function (response) {
     // The server is responding!
     responseStatus = response.status;
-    return responseStatus == 204 ? null : response.json();
+    return responseStatus === 204 ? null : response.json();
```
jankeromnes: 👍!

static/js/bootstrap-3.3.7.min.js
magopian: ah, that, sorry, didn't understand your statement at first ;)

static/js/settings.js
```diff
@@ -1,11 +1,11 @@
 // Copyright © 2015 Jan Keromnes. All rights reserved.
 // The following code is covered by the AGPL-3.0 license.
 
+/* global ajaxForm, updateFormStatus */
```
nt1m: I think you can use something like import-globals-from or define something in static/.eslintrc.js

.eslintrc-node.js
magopian: iiuc the browsers are not all on board with allowing dangling commas, that's why i have https://github.com/JanitorTechnology/janitor/pull/101/files#diff-a75257962af8dc9ccbfdb1d2c43be2feR13 which disables those checks. But then maybe I should rewrite this to `"comma-dangle": "off"` instead, it would be clearer I guess.

Regarding node, it does allow dangling commas, right? So that's why I have `"only-multiline"`, but as we discussed on irc, only for objects and arrays, nothing else, or did I misunderstand? So appart from rewriting the browser rule to be clearly `"off"`, what else do you want me to modify?

api/user-api.js
magopian: It was an auto-fix <3

lib/certificates.js
magopian: Same here, just an auto-fix

static/js/graphs.js
magopian: Wow, so I did completely misunderstand our discussion on IRC. My understanding was that browsers don't all allow them, but node does. Can you please confirm that it's the opposite?

static/js/graphs.js
magopian: Ok I'll revert this commit ;)

magopian:
i believe I've addressed all your comments (but https://github.com/JanitorTechnology/janitor/pull/101#discussion_r126185661, should I do something about that?).

Please let me know if there's anything else you wish me to modify ;)

join.js
```diff
@@ -16,9 +16,10 @@ const sessions = require('./lib/sessions');
 const hostname = db.get('hostname', 'localhost');
 
 if (!hostname || hostname === 'localhost') {
-  log('[fail] cannot join cluster as [hostname = ' + hostname + ']: ' +
-    'please fix the hostname in ./db.json and try again');
-  return;
+  const error = '[fail] cannot join cluster as [hostname = ' + hostname + ']: ' +
+    'please fix the hostname in ./db.json and try again';
+  log(error);
+  throw new Error(error);
```
jankeromnes: Nit: Please either `log('[fail] something something')` or `throw new Error('Something something')`, but not both. (In this case, I believe you want to just `throw`, and thus there is no need for the `[fail]` log prefix).

join.js
```diff
@@ -258,16 +259,16 @@ function routeRequest (proxyParameters, request, response) {
       routes.webProxy(request, response, { port, path });
       return;
 
-    case 'none':
+    case 'none': {
       const url = 'https://' + hostname + ':' + port + path;
       routes.redirect(response, url);
       return;
+    }
```
jankeromnes: Nit: What? Why?

lib/boot.js
```diff
@@ -263,7 +263,7 @@ exports.ensureDockerTlsCertificates = function (next) {
             return;
           }
 
-          fs.chmod(path, 0600 /* read + write by owner */, (error) => {
+          fs.chmod(path, '0600' /* read + write by owner */, (error) => {
```
jankeromnes: Why change this to a String? The Node.js [fs.chmod documentation](https://nodejs.org/api/fs.html#fs_fs_chmod_path_mode_callback) says that the `mode` argument should be an Integer (even though it does unofficially support strings). There is no example, but the default value is said to be `0o666` in [other places](https://nodejs.org/api/fs.html#fs_fs_appendfile_file_data_options_callback).

lib/certificates.js
```diff
@@ -1,7 +1,7 @@
 // Copyright © 2016 Jan Keromnes. All rights reserved.
 // The following code is covered by the AGPL-3.0 license.
 
-let child_process = require('child_process');
+let childProcess = require('child_process');
```
jankeromnes: Grmbl... fair enough. But we might as well make these const, by enabling the [prefer-const](http://eslint.org/docs/rules/prefer-const) ESLint rule.

lib/docker.js
```diff
@@ -1,14 +1,15 @@
 // Copyright © 2016 Jan Keromnes. All rights reserved.
 // The following code is covered by the AGPL-3.0 license.
 
-const dockerode = require('dockerode');
+const Dockerode = require('dockerode');
```
jankeromnes: Nit: What? Why?

Edit: Hm, because of the usage with `new`? ... Well, ok then, if it must be.

lib/docker.js
```diff
@@ -206,12 +207,14 @@ exports.removeContainer = function (parameters, callback) {
...
 };
 
-
 // Get the Docker version of a given host.
 exports.version = function (parameters, callback) {
   const { host } = parameters;
 
   getDocker(host, (error, docker) => {
+    if (error) {
+      log('[fail] Error while getting docker', error);
```
jankeromnes: Nit: In general, log messages are lower-case, while error messages in `new Error()`s are capitalized.

Nit 2: Please add the word `client` as in `while getting docker client`.

lib/hosts.js
```diff
@@ -33,8 +33,8 @@ exports.get = function (hostname) {
 
 // Find a Docker hostname that fully matches the request's OAuth2 credentials.
 exports.authenticate = function (request) {
-  let { client_id = null, client_secret = null } = request.query;
-  if (!client_id || !client_secret) {
+  let { clientId = null, clientSecret = null } = request.query;
```
jankeromnes: ⚠️ This change breaks the destructuring (`request.query` specifically contains the keys `client_id` and `client_secret`, but not `clientId` nor `clientSecret`).

Hint: The correct fix would be:

```js
let { client_id: clientId = null, client_secret: clientSecret = null } = request.query;
```

But that's very long and unreadable. I'll let you decide whether to continue using `client_id` everywhere, or fix the destructuring. (Maybe it might become more readable multi-line?)

Hint 2: @nt1m suggested not using destructuring in this case, i.e.:

```js
let clientId = request.query.client_id || null;
let clientSecret = request.query.client_secret || null;
```

lib/hosts.js
```diff
@@ -155,17 +155,17 @@ exports.resetOAuth2ClientSecret = function (host, callback) {
 
 // Pre-authorize a host to make OAuth2 requests on behalf of a user.
 exports.issueOAuth2AuthorizationCode = function (request, callback) {
-  let { client_id, scope, state, redirect_url = null } = request.query;
-  const hostname = exports.identify(client_id);
+  let { clientId, scope, state, redirectUrl = null } = request.query;
```
jankeromnes: ⚠️ Same as above, broken destructuring.

lib/hosts.js
```diff
@@ -188,21 +188,21 @@ exports.issueOAuth2AuthorizationCode = function (request, callback) {
...
 
-  oauth2provider.generateAuthorizationCode(client_id, grant, state, onCode);
+  oauth2provider.generateAuthorizationCode(clientId, grant, state, onCode);
 };
 
 // Effectively allow a host to make OAuth2 requests on behalf of a user.
 exports.issueOAuth2AccessToken = function (request, callback) {
-  const { client_id, code, state } = request.query;
-  const hostname = exports.identify(client_id);
+  const { clientId, code, state } = request.query;
```
jankeromnes: ⚠️ Broken destructuring.

lib/hosts.js
```diff
@@ -188,21 +188,21 @@ exports.issueOAuth2AuthorizationCode = function (request, callback) {
       return;
     }
     const { code } = data;
-    redirect_url += (redirect_url.includes('?') ? '&' : '?') +
+    redirectUrl += (redirectUrl.includes('?') ? '&' : '?') +
       'code=' + encodeURIComponent(code) +
       '&state=' + encodeURIComponent(state);
-    callback(null, { code, redirect_url });
+    callback(null, { code, redirectUrl });
```
jankeromnes: ⚠️ Caution: This changes the object format sent with the callback. It will no longer include a key called `redirect_url`, and instead include `redirectUrl`. Are you sure this is ok? Otherwise, you can use `{ code, redirect_url: redirectUrl }` to preserve the output format.

lib/machines.js
```diff
@@ -87,7 +87,7 @@ exports.rebuild = function (projectId, callback) {
 
   docker.buildImage({ host, tag, dockerfile }, (error, stream) => {
     if (error) {
-      log ('rebuild', image, error);
+      log('rebuild', image, error);
```
jankeromnes: Please drive-by fix this log message to include the prefix `[fail]`.

lib/machines.js
```diff
@@ -288,7 +288,7 @@ exports.destroy = function (user, projectId, machineId, callback) {
 // Install or overwrite a configuration file in all the user's containers.
 exports.deployConfigurationInAllContainers = function (user, file) {
   let count = 0;
-  const machines = Object.values(user.machines)
+  const machines = Object.values(user.machines) // eslint-disable-line node/no-unsupported-features
```
jankeromnes: Nit: Doesn't this line overshoot the 80-char limit? Also, we didn't have a debate about the whole 80-char vs 100-char vs no-limit story 😄

lib/oauth2.js
```diff
@@ -106,6 +107,10 @@ exports.request = function (parameters, callback) {
     const body = data ? JSON.stringify(data, null, 2) : null;
     let url = 'https://' + api + path;
 
+    if (error) {
+      log('[fail] Error while getting the client', error);
+    }
```
jankeromnes: Nit: I prefer `error` parameters to be the first thing handled in a function. Please move it to the function's first line, i.e. above `const api = ...`.

jankeromnes: Nit 2: Please use lower-case log messages (as in `error` instead of `Error`). Additionally, logging the variable `error` will already print `Error: Something something` in the console. Maybe it's not necessary to repeat the word "error" multiple times? I.e.

```
[2017-07-10T16:00:43.733Z] [fail] Error while getting the client Error: Couldn't reach host 'example.com'
```

lib/routes.js
```diff
@@ -364,6 +364,10 @@ exports.webProxy = function (request, response, parameters) {
   });
 
   proxy.on('error', error => {
+    if (error) {
+      log('[fail] Error while processing the request', error);
```
jankeromnes: Nit: Cool fix! But please use lower-case error messages (or remove the word "Error").

lib/sessions.js
```diff
@@ -69,6 +70,10 @@ exports.get = function (request, callback) {
 // Destroy the session associated to the given request.
 exports.destroy = function (request, callback) {
   exports.get(request, (error, session, token) => {
+    if (error) {
+      log('[fail] Error while destroying the session', error);
```
jankeromnes: Nit: Same as earlier nits (word "Error" is superfluous).

lib/streams.js
```diff
@@ -3,14 +3,14 @@
 
 var stream = require('stream');
```
jankeromnes: Whoa, this file still uses `var` 😄 Maybe it can be auto-fixed by ESLint to use `const`, or `let` if it really needs to be mutable?

lib/streams.js
```diff
@@ -52,23 +49,24 @@ function set (object, key, readable) {
 
   // Remove the stream if an error occurs.
   readable.on('error', function (error) {
+    if (error) {
+      log('[fail] Error reading the stream', error);
```
jankeromnes: Nit: Please remove the word `Error`, e.g. like so `'[fail] problem while reading stream'`.

lib/users.js
```diff
@@ -95,6 +95,10 @@ exports.resetSSHKeyPair = function (user) {
 // Send a single-use login link to the user's email address.
 exports.sendLoginEmail = function (email, request, callback) {
   sessions.get(request, (error, session, token) => {
+    if (error) {
+      log('[fail] Error sending the login email', error);
```
jankeromnes: Nit: -Error, -the.

lib/users.js
```diff
@@ -145,6 +149,10 @@ exports.sendLoginEmail = function (email, request, callback) {
 exports.sendInviteEmail = function (email, callback) {
   // Generate a dummy token for that invite.
   sessions.create((error, token, session) => {
+    if (error) {
+      log('[fail] Error creating the session', error);
```
jankeromnes: Nit: Maybe `'[fail] could not create session'`?

package.json
```diff
@@ -22,7 +22,13 @@
...
     "start": "if [ -z \"$SCRIPT\" ] ; then printf \"Run which Janitor script? [join/app]:\" && read SCRIPT ; fi ; node \"$SCRIPT\" >> janitor.log 2>&1 & printf \"$!\\n\" > janitor.pid",
     "poststart": "printf \"[$(date -uIs)] Background process started (PID $(cat janitor.pid), LOGS $(pwd)/janitor.log).\\n\"",
-    "stop": "if [ -e janitor.pid -a -n \"$(ps h $(cat janitor.pid))\" ] ; then kill $(cat janitor.pid) && printf \"[$(date -uIs)] Background process stopped (PID $(cat janitor.pid)).\\n\" ; fi ; rm -f janitor.pid"
+    "stop": "if [ -e janitor.pid -a -n \"$(ps h $(cat janitor.pid))\" ] ; then kill $(cat janitor.pid) && printf \"[$(date -uIs)] Background process stopped (PID $(cat janitor.pid)).\\n\" ; fi ; rm -f janitor.pid",
+    "lint-node": "eslint -c .eslintrc-node.js *.js api/ lib/",
+    "lint-node-fix": "npm run lint-node -- --fix",
+    "lint-browser": "eslint -c .eslintrc-browser.js static/",
+    "lint-browser-fix": "npm run lint-browser -- --fix",
+    "lint": "npm run lint-node && npm run lint-browser",
+    "lint-fix": "npm run lint-node-fix && npm run lint-browser-fix"
```
jankeromnes: Nit: Is it really worth it to have separate `lint-node{,-fix}` and `lint-browser{,-fix}` commands? Why not just have `lint` and `lint-fix`?

static/js/graphs.js
jankeromnes: So, most browsers support:

```js
var a = {
  b: 4,
  c: 5,
};
```

And:

```js
var a = [
  c,
  d,
];
```

But only very recent browsers support:

```js
function bla(
  a,
  b,
  c,
) { }
```

So in browser code, we want to forbid dangling spaces in function arguments (as well as in imports/exports since they don't make sense in a browser env anyway).

In Node.js code, we can add dangling commas everywhere (i.e. on every multi-line thing), even function parameters.

magopian: I believe I changed it to do just that, if I'm wrong, please let me know ;)

lib/boot.js
magopian: The octal syntax seems to be deprecated, see http://eslint.org/docs/rules/no-octal (first paragraph). In strict mode, eslint will complain about a parsing error.

join.js
magopian: It's my understanding that those are needed to fix http://eslint.org/docs/rules/no-case-declarations

lib/streams.js
magopian: I will add a commit with such a rule

lib/hosts.js
magopian: I'm reverting all those changes and allowing non-camelcase names instead. Thanks for spotting this!

package.json
magopian: You're right, I'll remove the others ;)

magopian:
I believe I've addressed your comments, please let me know if I've missed anything.

I added the `no-var` rule in https://github.com/JanitorTechnology/janitor/pull/101/commits/af4933dd1ed1a6aa05313a2101d83c3d331aeae0, please have a deep look at my manual changes there, to make sure I haven't set something as `const` while it should really be a `let` ;)

I also rebased the PR, so I hope I didn't break anything in the process :/

magopian:
Rebased following latest PR merges

.eslintrc-browser.js
```diff
@@ -0,0 +1,36 @@
+module.exports = {
+  "env": {
+    "browser": true,
+    "es6": false
```
nt1m: Quite funny that the eslintrc.js files don't actually respect the eslint rules :D 

Property names should either have no quotes or single quotes.

join.js
nt1m: Can you simply move the contents of `url`  inside routes.redirect and remove the extra variable and the brackets?

join.js
```diff
@@ -258,16 +257,16 @@ function routeRequest (proxyParameters, request, response) {
...
 
     default:
       log('[fail] unsupported proxy type:', proxy);
       response.statusCode = 500; // Internal Server Error
       response.end();
-      return;
```
nt1m: It's inconsistent that we have `return;` in the above `case` statement, but not here. Can you instead use `break;` here and above to remove this inconsistency?

lib/boot.js
nt1m: I prefer that we use `0o666` rather than `'0600'`.

lib/db.js
```diff
@@ -25,38 +24,31 @@ function load () {
...
-
-  var json = JSON.stringify(store, null, 2);
+  const json = JSON.stringify(store, null, 2);
 
   fs.writeFile(file, json + '\n', function (error) {
     if (error) {
       console.error('Can\'t write DB file!', error.stack);
     }
-    fs.chmod(file, 0600 /* read + write by owner */, function (error) {
+    fs.chmod(file, '0600' /* read + write by owner */, function (error) {
```
nt1m: Same here.

lib/machines.js
nt1m: You can use:
```js
// eslint-disable-next-line node/no-unsupported-features
const machines = Object.values(user.machines)
```

lib/metrics.js
```diff
@@ -132,23 +116,19 @@ exports.getContributionData = function () {
...
 
 exports.getHostData = function () {
-
   let data = {
     docker: []
   };
   let hosts = db.get('hosts');
 
-  for (let hostname in hosts) {
+  for (let hostname in hosts) { // eslint-disable-line no-unused-vars
```
nt1m: Nit: please use `// eslint-disable-next-line` instead.

static/js/graphs.js
```diff
@@ -50,14 +45,11 @@ function formatTime (milliseconds) {
...
   return (Math.round(value * 10) / 10) + ' ' + unit.code;
-
 }
 
-
 // Format bytes into human readable text.
 
-function formatMemory (bytes) {
-
+function formatMemory (bytes) { // eslint-disable-line no-unused-vars
```
nt1m: nit: I prefer you use `/* exported formatMemory */` on top of the file

lib/boot.js
```diff
@@ -292,7 +292,7 @@ exports.ensureDockerTlsCertificates = function (next) {
       return;
     }
 
-    done = null;
+    done = null; // eslint-disable-line no-func-assign
```
nt1m: nit: please use 
```js
// eslint-disable-next-line no-func-assign
done = null;
```

I think it's cleaner as it doesn't clutter up the line length.

.eslintrc-browser.js
magopian: I believe it's because you can either use a js file, or plain json? I don't have an opinion on this matter. Also for some reason those configuration files are auto-ignored by eslint.

static/js/graphs.js
magopian: that doesn't seem to change eslint's mind about warning on this line :/

Probably because of https://github.com/eslint/eslint/issues/1200#issuecomment-156464573

lib/certificates.js
```diff
@@ -369,7 +341,7 @@ exports.createHTTPSCertificate = function (parameters, callback) {
     }
 
     // All required tasks are now finished, destroy the `done` callback.
-    done = null;
+    done = null; // eslint-disable-line no-func-assign
```
nt1m: nit: use eslint-disable-next-line

static/js/graphs.js
nt1m: nit: please use // eslint-disable-next-line

magopian:
@nt1m and @jankeromnes i'm not sure I've addressed all your comments, this PR is getting so long that it's hard to tell. Please let me know if I've missed anything.

On a side note, I did change the eslint exceptions to use `// eslint-disable-next-line` but I'm not sure it's a good idea: it's the kind of comments which will
- have a much bigger importance (at the same level as code comments explaining pieces of the implementation, and adding "noise", instead of just being an addition far to the right of the line
- be easily forgotten if the line that is being ignored is moved around, or fixed, or removed

nt1m:
> have a much bigger importance (at the same level as code comments explaining pieces of the implementation, and adding "noise", instead of just being an addition far to the right of the line

I think eslint-disable-next-line looks cleaner as it doesn't make the line too long. I do think having a line that's suddenly wider than the other ones looks odd. I would even be favorable to forbidding long inline comments.

> be easily forgotten if the line that is being ignored is moved around, or fixed, or removed

Either eslint will catch the mistake, either the comment is no longer relevant :)

join.js
```diff
@@ -253,21 +252,21 @@ function routeRequest (proxyParameters, request, response) {
   }
 
   const { port, proxy } = proxyParameters;
+  const url = 'https://' + hostname + ':' + port + path;
```
jankeromnes: Nit: This `url` variable is not used by anyone else than the `routes.redirect()` below. Please inline its value directly into the redirect call (no need for an intermediary variable).

lib/configurations.js
```diff
@@ -70,7 +70,7 @@ exports.resetToDefault = function (user, file, callback) {
   }
 
   let content = '';
-  stream.on('data', chunk => content += String(chunk));
+  stream.on('data', chunk => (content += String(chunk)));
```
jankeromnes: Nit: I believe you meant `{ content += String(chunk); }` instead of `(content += String(chunk))` (we didn't actually want to return the results of the `+=`).

package.json
```diff
@@ -24,10 +24,12 @@
     "poststart": "printf \"[$(date -uIs)] Background process started (PID $(cat janitor.pid), LOGS $(pwd)/janitor.log).\\n\"",
     "stop": "if [ -e janitor.pid -a -n \"$(ps h $(cat janitor.pid))\" ] ; then kill $(cat janitor.pid) && printf \"[$(date -uIs)] Background process stopped (PID $(cat janitor.pid)).\\n\" ; fi ; rm -f janitor.pid",
     "test": "cd tests && node tests.js",
-    "watch": "watch-run --initial --pattern 'app.js,package.json,api/**,lib/**,templates/**' --stop-on-error npm run app & tailf janitor.log -n 0"
+    "watch": "watch-run --initial --pattern 'app.js,package.json,api/**,lib/**,templates/**' --stop-on-error npm run app & tailf janitor.log -n 0",
+    "lint": "eslint -c .eslintrc-node.js *.js api/ lib/ && eslint -c .eslintrc-browser.js static/",
+    "lint-fix": "eslint -c .eslintrc-node.js *.js api/ lib/ --fix && eslint -c .eslintrc-browser.js static/ --fix"
```
jankeromnes: Nit: Please alpha-sort NPM scripts (except for `pre*` and `post*` scripts).

api/user-api.js
```diff
@@ -168,7 +167,7 @@ configurationsAPI.patch({
       ], null, 2)
     },
     response: {
-      body: JSON.stringify({ '.gitconfig': '[user]\nname = Sally' }, null, 2)
+      body: JSON.stringify({ '.gitconfig': '[user]\nname = Janitor' }, null, 2)
```
magopian: Yeah, I don't know where this change comes from, I believe it was from a rebase? I personally didn't make this change, do you want me to revert it?

nt1m: Yeah, I think it would be preferable to revert it

PR #102, opened 2017-07-10
Automatically test the Janitor API against its own request/response examples.

jankeromnes:
This uses a semi-real Janitor server to mount the API, and runs all request/response examples against their respective handlers.

The only fake things are:
- A pre-registered `admin@example.com` admin user,
- A few simplified values (e.g. `.gitconfig` just contains `"[user]\nname = User"`),
- Some methods from `lib/docker.js` (because it's hard to test against a real Docker Remote API daemon or mock server)

### DIFF
--- a/api/admin-api.js
+++ b/api/admin-api.js
@@ -14,7 +14,24 @@ const adminAPI = module.exports = selfapi({
 });
 
 // API sub-resource to manage OAuth2 providers.
-const oauth2providersAPI = adminAPI.api('/oauth2providers');
+const oauth2providersAPI = adminAPI.api('/oauth2providers', {
+  beforeEachTest: next => {
+    const providers = db.get('oauth2providers');
+    if (!providers.github) {
+      // FIXME Remove this if block when the GitHub pull request is merged:
+      // https://github.com/JanitorTechnology/janitor/pull/80
+      providers.github = {
+        id: '',
+        secret: '',
+        hostname: 'github.com',
+        api: 'api.github.com'
+      };
+    }
+    providers.github.id = '1234';
+    providers.github.secret = '123456';
+    next();
+  }
+});
 
 oauth2providersAPI.get({
   title: 'List OAuth2 providers',
@@ -34,10 +51,11 @@ oauth2providersAPI.get({
   examples: [{
     response: {
       body: JSON.stringify({
-        provider: {
+        github: {
           id: '1234',
           secret: '123456',
-          hostname: 'host.name'
+          hostname: 'github.com',
+          api: 'api.github.com'
         }
       }, null, 2)
     }
@@ -45,7 +63,9 @@ oauth2providersAPI.get({
 });
 
 // API sub-resource to manage a single OAuth2 provider.
-const oauth2providerAPI = oauth2providersAPI.api('/:provider');
+const oauth2providerAPI = oauth2providersAPI.api('/:provider', {
+  beforeEachTest: oauth2providersAPI.beforeEachTest
+});
 
 oauth2providerAPI.patch({
   title: 'Update an OAuth2 provider',
@@ -91,17 +111,18 @@ oauth2providerAPI.patch({
 
   examples: [{
     request: {
-      urlParameters: { provider: 'provider' },
+      urlParameters: { provider: 'github' },
       body: JSON.stringify([
-        { op: 'add', path: '/secret', value: '654321' }
-      ], null, 2)
+        { op: 'add', path: '/secret', value: '654321' },
+      ], null, 2),
     },
     response: {
       body: JSON.stringify({
         id: '1234',
         secret: '654321',
-        hostname: 'host.name'
-      }, null, 2)
+        hostname: 'github.com',
+        api: 'api.github.com',
+      }, null, 2),
     }
   }]
 });
--- a/tests/tests.js
+++ b/tests/tests.js
@@ -1,0 +1,229 @@
+// Copyright © 2017 Jan Keromnes. All rights reserved.
+// The following code is covered by the AGPL-3.0 license.
+
+'use strict';
+
+const fs = require('fs');
+const net = require('net');
+const path = require('path');
+const selfapi = require('selfapi');
+
+if (path.basename(process.cwd()) !== 'tests') {
+  console.error('Warning: Tests need to run inside the tests/ folder.\n' +
+    '  This is to prevent interference between tests and production.\n' +
+    '  Teleporting to tests/ now.');
+  process.chdir('tests');
+}
+
+try {
+  fs.unlinkSync('./db.json');
+} catch (error) {
+  if (error.code !== 'ENOENT') {
+    console.error(error.stack);
+    process.exit(1);
+  }
+}
+
+try {
+  fs.symlinkSync('../templates', './templates', 'dir');
+} catch (error) {
+  if (error.code !== 'EEXIST') {
+    console.error(error.stack);
+    process.exit(1);
+  }
+}
+
+let tests = [];
+
+tests.push({
+  title: 'Janitor API self-tests',
+
+  test: (port, callback) => {
+    const db = require('../lib/db');
+    // Tell our fake Janitor app that it runs on the fake host "example.com":
+    db.get('hostname', 'example.com');
+    // Disable sending any emails (for invites or signing in):
+    db.get('mailer').block = true;
+    // Disable Let's Encrypt HTTPS certificate generation and verification:
+    db.get('security').forceHttp = true;
+    // Disable any security policies that could get in the way of testing:
+    db.get('security').forceInsecure = true;
+
+    const boot = require('../lib/boot');
+    const docker = require('../lib/docker');
+    const machines = require('../lib/machines');
+    const users = require('../lib/users');
+
+    // Fake several Docker methods for testing.
+    // TODO Maybe use a real Docker Remote API server (or a full mock) for more
+    // realistic tests?
+    docker.runContainer = function (parameters, callback) {
+      callback(null, { id: 'abcdef0123456789' }, '');
+    };
+    docker.copyIntoContainer = docker.execInContainer =
+      function (parameters, callback) {
+        callback();
+      };
+    docker.version = function (parameters, callback) {
+      callback(null, { Version: '17.06.0-ce' });
+    };
+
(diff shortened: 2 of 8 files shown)


jankeromnes:
All Self API tests are now passing! 💯🥇🎆

jankeromnes:
@magopian Could you please have a look to double-check I haven't made any obvious mistakes?

Also, I'd be interested in your opinion of the minimal test framework I implemented (it currently has only one test, which itself runs all 18 API tests). Do you like it as is, or would you prefer something more well-established like mocha, or chai, or some other test system/module/framework?

Thanks! 😃

tests/tests.js
```diff
@@ -0,0 +1,220 @@
+// Copyright © 2017 Jan Keromnes. All rights reserved.
+// The following code is covered by the AGPL-3.0 license.
+
+'use strict';
+
+require('promise/lib/rejection-tracking').enable({ allRejections: true });
```
jankeromnes: Self-nit: Actually maybe this line is pointless. It didn't help me find out whichever error was being trapped by the `promise` module before I updated `email-login`. Oh well. Hopefully we'll switch all promises to Node.js' native Promise implementation soon. 🙂

tests/tests.js
```diff
@@ -0,0 +1,220 @@
...
+  }
+}
+
+let tests = [];
+
+tests.push({
+  title: 'Janitor API self-tests',
+
+  test: (port, callback) => {
+    const db = require('../lib/db');
```
magopian: I understand the need to run the tests from the `tests` folder now :P

I wonder if it would be possible to have a "db name" in the db file. This way the worse you could have is a `test` db inside your production `db.json` file, if for some reason you mess things up and you end up running the tests with the production DB.

tests/tests.js
```diff
@@ -0,0 +1,220 @@
...
+    docker.copyIntoContainer = docker.execInContainer =
+      function (parameters, callback) {
+        callback();
+      };
+    docker.version = function (parameters, callback) {
+      callback(null, { Version: '17.06.0-ce' });
+    };
+
+    function registerTestUser (next) {
+      db.get('admins')['admin@example.com'] = true;
```
magopian: It's still a bit confusing to me when I read such lines: I believe there's no user in the db at this point, right? It's just automatically created when using get on a missing key. I'm not very comfortable with that.

I think I'd rather have something like get_or_create which would then return a tuple (the value, and a boolean telling if the key was created). It's a bit more verbose, but far more explicit.

What do you think?

tests/tests.js
```diff
@@ -0,0 +1,220 @@
...
+
+/*
+tests.push({
+  title: '',
+  test: (port, callback) => {
+    // test some things
+    // callback(error);
+  }
+});
+*/
```
magopian: Do you want to keep those commented blocks in the PR?

tests/tests.js
```diff
@@ -0,0 +1,220 @@
...
+      test.test(port, error => reportTest(test, error));
+    } catch (error) {
+      reportTest(test, error);
+    }
+  });
+}
+
+while (tests.length > 0) {
+  const i = Math.floor(Math.random() * tests.length);
+  const test = tests.splice(i, 1)[0];
```
magopian: What is that for? To run the tests in random order on each execution? Maybe add a comment?

Also, would something like

```js
const test = tests[Math.floor(Math.random() * tests.length)];
```

be more concise?

jankeromnes:
Thanks a lot for the helpful review! You make excellent points.

> This is an impressive piece of work, congratz!

Yay, thanks! 😄

> I try to shy away from the NIH syndrome as a general rule

Good point! I try to do the same, but may not be very successful at avoiding it (what's out there often doesn't fit my exact needs, or seems of poor quality, so I sometimes re-implement stuff when it seems quick... e.g. Janitor's OAuth2 scheme, the Self API framework (but I really think there were no viable Node.js alternatives for these, and they work great!))

> I like to maintain as few lines of code as possible (not less), because that means it's more time that I can focus on useful stuff

Me too! 👍 (Although for small things I do prefer verbose copy/pastes that are dead simple to delete over shorter abstractions that might need time-consuming refactoring later.)

> more well-established frameworks usually have more people contributing to it, and usually have very useful features (I'm thinking about code coverage, ability to re-run a single test, cover more corner cases, have helpers and utilities, faster test runners, mock helpers, logging, verbose mode...)

I think that's the main point to win me over. For example, [Self API](https://github.com/JanitorTechnology/selfapi) is super helpful and simplifies/solidifies the Janitor API implementation a great deal, but sometimes I wish there were some similar project in Node.js with a big community like Python's Flask, RESTPlus, etc.

> So I would personally advise going with whatever test framework other teams at Mozilla are using (like the html devtools team, or the addons team).

The problem here is that we're special, because we use Node.js (noone else at Mozilla does this I think).

DevTools switched to React/Redux, and so they use the [Jest](https://facebook.github.io/jest/docs/tutorial-react.html) test framework which is excellent for React-debugging. We don't use anything like that in Janitor (and don't plan to), so using Jest probably wouldn't make. They also use Marionette, which I don't want to get close to again in my life 😛 (because of all the convoluted, mind-bending DevTools tests, and all the various flakiness, intermittents, leaks and bloats (Note: My bad experience with it was certainly due to my own inexperience rather than Marionette itself)).

I don't know what the Addons team uses, but I'm not sure how well it applies to Node.js apps.

But in general, I agree that we should try to use an existing test framework instead of awkwardly growing our `tests/tests.js` file, especially if we feel the need for more advanced helpers, code coverage, etc. I think a great widely-used Node.js test framework nowadays is [mocha](https://mochajs.org/). I even tried using it in `tests/tests.js`, but couldn't wrap my head around it in time so I gave up. Let's file an issue and try that again! 😃

> I'm not sure about the `beforeEachTest` spread in the code though. It looks like fixtures that would probably be better placed in the tests file? But then maybe it makes sense to have them at the same place as the examples used by the selfapi auto-testing feature.

Good point. However, the whole point of Self API is to place code examples right next to your actual API handlers, and naturally write and refactor them when you work on the API handlers, "automatically" keeping your tests and documentation up-to-date with very limited effort. There are some test fixtures in `tests/tests.js` already (because they're sort of common to all handlers).

The `{before,after}EachTest` functions are supposed to be really small fixtures to make your examples work, and should be too distracting from the handlers. Do you think that they're small enough to be kept next to their handlers without making things confusing? Please let me know if you think some fixtures should really be moved elsewhere.

> On a side note, have you tried using CircleCI? I've been told by some mozillians that it's the way to go nowadays?

I've compared the CircleCI and Travis CI services for our needs, and Travis CI seemed better for our `selfapi` and `janitor` repositories (because it's free and has some multi-platform web testing support), while CircleCI seemed better for our `dockerfiles` repository (because it can be used to build Dockerfiles into images, and push them to production automatically).

api/admin-api.js
```diff
@@ -14,7 +14,24 @@ const adminAPI = module.exports = selfapi({
...
-const oauth2providersAPI = adminAPI.api('/oauth2providers');
+const oauth2providersAPI = adminAPI.api('/oauth2providers', {
+  beforeEachTest: next => {
+    const providers = db.get('oauth2providers');
+    if (!providers.github) {
+      // FIXME Remove this if block when the GitHub pull request is merged:
+      // https://github.com/JanitorTechnology/janitor/pull/80
+      providers.github = {
+        id: '1234',
+        secret: '123456',
```
jankeromnes: True. I could avoid setting the `id` and `secret` keys here altogether. Let's do that 😄

tests/tests.js
jankeromnes: Interesting idea! I think the main reason for running all this in a separate folder was to avoid the need for being "smart" about such file names. (E.g. I believe we may also touch files like `./janitor.pid` and `./janitor.log` if we use any NPM scripts, or `./docker.crt`, `./docker.key` and `./docker.ca` if we verify/generate Docker TLS certificates, etc).

I even wonder if we'll need many more test sub-folders, e.g. when we'll test running a Janitor app (`node app.js`) and one or more Janitor cluster hosts (`node join.js`) at the same time.

tests/tests.js
jankeromnes: Ah, I get why this line is confusing, but maybe it's not because of `db.get()`'s behavior:
- `db.get("name")`'s only job is to give you a editable key/value store called `"name"` that lives in the DB. That key/value store is an Object, so if it didn't previously exist it's going to be `{}` always.
- In [db.json](https://discourse.janitor.technology/t/explaining-db-json/12), there is a special key/value store called `"admins"` that looks like this: `{ "email@example.com": true, "other@example.com": true }`. It specifies whose emails should be granted admin privileges on the website.
- In the code you commented on, we simply take the `"admins"` key/value store, and add an entry for `admin@example.com`.
- Additionally, due to this change, the next line won't complain about the email `admin@example.com` not being invited to the Alpha, and will simply create it.

I should definitely add a comment to explain these "magical" fixture lines. Sorry about causing you these headaches!

tests/tests.js
jankeromnes: I did it on purpose, because this block allows anyone to easily copy/paste it and add a new test (without having to read through the current, very long tests).

The code above is another test that I'd like to add. I usually keep these comments local to my environment, but up to now they either linger around (causing me rebase trouble), and other developers have independently identified and fixed them. So I thought, since this is a test file, why not add my `TODO` thoughts?

Would you prefer I remove them to make this file cleaner? I would't mind :)

tests/tests.js
jankeromnes: Agreed! Thanks for catching this! I should definitely add a comment, and use your shorter suggestion.

I believe the `i` variable is a left-over from several refactors prior to this pull request! I agree it doesn't serve much purpose anymore.

jankeromnes: P.S.: Oh, hi there 😄 
https://github.com/JanitorTechnology/selfapi/blob/103e12c311e21aa7249a5662cae8e861b0375d70/selfapi.js#L186-L187

tests/tests.js
magopian: I usually don't keep "dead code" around, but then if it's for a good reason, why not ;)

magopian:
Thanks for taking the time to answer my very long review/comment, it shows me we share most (if not all) values about code, and that's awesome 💯

Issue #95, opened 2017-07-05
Add live-reloading to the project


magopian:
It would be very convenient to have the app restart automatically whenever some files change. I believe there's many different solutions that would do that out there.

jankeromnes:
Very cool suggestion! Looks like many developers expect to have this nowadays, because you're not the first to request this feature 😃. See also https://github.com/JanitorTechnology/janitor/projects/2#card-3209302.

Also, @nt1m might be interested in addressing this (except if you've already started @magopian?)

magopian:
Haven't started, not yet ;)

jankeromnes:
This issue is being tackled in pull request #103, so I'm closing it in order to keep the list of open issues small and readable.

PR #103, opened 2017-07-10
Add watch command (#95)

nt1m:


package.json
```diff
@@ -16,10 +16,11 @@
     "url": "https://github.com/janitortechnology/janitor"
   },
   "scripts": {
+    "watch": "watch \"npm run app\" api lib templates & tailf janitor.log",
```
jankeromnes: Nit: It would be great to also restart on changes to the file `app.js`. Not sure if that's possible though.

jankeromnes: Nit: Please add a `--wait=4` option, in order not to restart unreasonably often (e.g. in case of startup failures).

package.json
```diff
@@ -37,6 +38,9 @@
     "selfapi": "~0.0.16",
     "tar-stream": "~1.5.4"
   },
+  "devDependencies": {
+    "watch": "1.0.2"
```
jankeromnes: Nit: I tried this without doing `npm install` first, and it seems to break in strange ways because it's using the unix `watch` command. Is it possible to explicitly tell NPM to use the module, and not the unix command, e.g. by explicitly using `./node_modules/watch/cli.js` instead of just `watch` in the `"watch"` NPM script?

nt1m:
@jankeromnes Updated

package.json
```diff
@@ -16,10 +16,11 @@
     "url": "https://github.com/janitortechnology/janitor"
   },
   "scripts": {
+    "watch": "watch-run -i -s -p \"api/**,templates/**,lib/**,app.js\" npm run app & tailf janitor.log -n 50",
```
jankeromnes: Nit: Please use `-n 0` instead of `-n 50` for the `tailf` command, because it's confusing to see (parts of) any previous errors when you start a new server.

jankeromnes: Nit: Please also sort the pattern in alphabetical order, grouped by files first and folders next, i.e. `"app.js,api/**,lib/**,templates/**"`. Also, maybe we should also watch the `package.json` file?

nt1m:
Thanks ! :+1:

PR #120, opened 2017-07-26
Tag internal Docker images with ':janitor-production'.

jankeromnes:
This change makes Janitor use the internal Docker tag `:janitor-production` for image updates and container spawns.

Previously, the default tag `:latest` was used for that, which caused the need for base images to use a different tag (e.g. `:base`).

As previously discussed on IRC, this wasn't convenient for most Docker Hub users out there wishing to add their projects to Janitor. So this change replaces the internal tag of `:latest` with an explicit `:janitor-production`, and allows using any base image tag instead of `:base` (i.e. you can now call base images `:latest`, or `:nightly`, or `:stable`, or `:1.0.5`, or anything you like except `:janitor-production`).

Note: The reason we need an external tag and an internal tag that's different is for building incremental image updates (we wish to build all incremental updates on top of the same base image, otherwise update layers would be chained indefinitely, causing very large Docker images).

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -74,9 +74,48 @@ exports.buildImage = function (parameters, callback) {
   });
 };
 
+// Pull a Docker image into a given host.
+exports.pullImage = function (parameters, callback) {
+  const { host, image: imageId } = parameters;
+
+  getDocker(host, (error, docker) => {
+    if (error) {
+      callback(error);
+      return;
+    }
+
+    docker.pull(imageId, function (error, stream) {
+      if (error) {
+        callback(error);
+        return;
+      }
+
+      callback(null, stream);
+    });
+  });
+};
+
+// Tag a Docker image in a given host.
+exports.tagImage = function (parameters, callback) {
+  const { host, image: imageId, tag: tagId } = parameters;
+
+  getDocker(host, (error, docker) => {
+    if (error) {
+      callback(error);
+      return;
+    }
+
+    const image = docker.getImage(imageId);
+    const [ repo, tag = 'latest' ] = tagId.split(':');
+    image.tag({ repo, tag }, (error, data) => {
+      callback(error);
+    });
+  });
+};
+
 // Delete a Docker image from a given host.
 exports.removeImage = function (parameters, callback) {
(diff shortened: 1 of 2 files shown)


jankeromnes:
@bnjbvr This change is quite small, but I'm afraid of forgetting a spot where the new tag `:janitor-production` should be used. Could you please have a quick read-through to see if anything looks suspicious?

lib/docker.js
```diff
@@ -74,9 +74,48 @@ exports.buildImage = function (parameters, callback) {
...
+  const { host, image: imageId, tag: tagId } = parameters;
+
+  getDocker(host, (error, docker) => {
+    if (error) {
+      callback(error);
+      return;
+    }
+
+    const image = docker.getImage(imageId);
+    const [ repo, tag = 'latest' ] = tagId.split(':');
```
bnjbvr: Can an image name have more than 2 `:`?

lib/docker.js
```diff
@@ -74,9 +74,48 @@ exports.buildImage = function (parameters, callback) {
...
+exports.pullImage = function (parameters, callback) {
+  const { host, image: imageId } = parameters;
+
+  getDocker(host, (error, docker) => {
+    if (error) {
+      callback(error);
+      return;
+    }
+
+    docker.pull(imageId, function (error, stream) {
```
bnjbvr: (you might as well do `docker.pull(imageId, callback);`, but again it's less explicit :))

jankeromnes: Agreed! Thanks for catching this, but I think I prefer the explicit, "dead code" version where it's easy to check what you're getting back from this function. 😄

lib/docker.js
jankeromnes: I certainly hope not! To verify, I tried doing this:

```
$ sudo docker pull ubuntu:16.04:latest
Error parsing reference: "ubuntu:16.04:latest" is not a valid repository/tag
```

It's currently possible to submit pretty much any value as `project.docker.image`, so I guess we might want to add validation at some point.

jankeromnes:
Thanks! 👍

PR #121, opened 2017-07-26
Add an API to manage projects, and use it pull their Docker images.

jankeromnes:


jankeromnes:
Potential problem with this implementation: Pull API is public, and can be spammed by anyone. Maybe we should rate limite?

api/projects-api.js
```diff
@@ -0,0 +1,65 @@
...
+
+// API sub-resource to manage a single software project.
+const projectAPI = projectsAPI.api('/:project');
+
+projectAPI.post('pull', {
+  title: 'Pull a project',
+  description: 'Trigger a Docker image pull for a given software project.',
+
+  handler: (request, response) => {
+    const { project: projectId } = request.query;
```
nt1m: nit: Probably more simple to do:
```
const projectId = request.query.project;
```

api/projects-api.js
```diff
@@ -0,0 +1,65 @@
...
+    const projects = db.get('projects');
+    if (!projects[projectId]) {
+      response.statusCode = 404;
```
nt1m: nit: `response.statusCode = 404; // Not found`

api/projects-api.js
```diff
@@ -0,0 +1,65 @@
...
+      response.json({ error: 'Project not found' }, null, 2);
+      return;
+    }
+
+    machines.pull(projectId, error => {
+      if (error) {
+        log('[fail] pulling project', projectId, error);
+        response.statusCode = 500;
```
nt1m: nit: `response.statusCode = 500; // Internal Error`

api/projects-api.js
```diff
@@ -0,0 +1,65 @@
...
+});
+
+// API sub-resource to manage a single software project.
+const projectAPI = projectsAPI.api('/:project');
+
+projectAPI.post('pull', {
+  title: 'Pull a project',
+  description: 'Trigger a Docker image pull for a given software project.',
+
+  handler: (request, response) => {
```
nt1m: This API shouldn't be public.

Maybe limit this to logged in users only, or even just admins (`users.isAdmin(user)`)

jankeromnes: The thing is that I'd also like to support Docker Hub web hooks. Maybe there is a way to authenticate that as well, eg with a shared secret.

nt1m: So, maybe using authentication would be a good choice. Or maybe just rate limit the API.

jankeromnes: Hm, I'd love to make this API public somehow, or at least very easily accessible to Docker Hub web hooks, but let's do that in another pull request. I'll make this handler admin-only to begin with.

jankeromnes:
I added a `docker.inspectImage()` method in order to verify the creation date of the image.
- If the image is **more recent** than our current production image, we tag it with `:janitor-production` (and set `project.data.updated` to the creation date, not the pull date)
- If the image is **as old as, or older than** our current production image (e.g. we already pulled this image, or we have an incremental update image that is more recent), then we still keep it, but we don't tag it with `:janitor-production` (and we don't change `project.data.updated`).

api/projects-api.js
```diff
@@ -0,0 +1,75 @@
...
+    const { user } = request;
+    // FIXME: Make this API handler accessible to Docker Hub web hooks, so that
+    // it's easy to automatically deploy new project images.
+    if (!users.isAdmin(user)) {
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' }, null, 2);
+      return;
+    }
+
+    const projectId = request.query.project;
```
nt1m: nit: above, use `const { user, query } = request;` and here: `const projectId = query.project;`

api/projects-api.js
```diff
@@ -0,0 +1,75 @@
...
+
+    machines.pull(projectId, error => {
+      if (error) {
+        log('[fail] pulling project', projectId, error);
+        response.statusCode = 500; // Internal Server Error
+        response.json({ error: 'Could not pull project' }, null, 2);
+        return;
+      }
+
+      response.json({}, null, 2);
```
nt1m: Not sure what's the point of this response. Why not just response.end() ?

jankeromnes:
Oh, so clicking on "Update branch" creates a merge commit... 

giphy

PR #123, opened 2017-07-28
Create API instead of using Scout for spawning and deleting containers

nt1m:


api/hosts-api.js
```diff
@@ -336,9 +336,81 @@ hostAPI.get('/version', {
   }]
 });
 
+hostAPI.post('/containers', {
+  title: 'Create a new container',
+  description: 'Create a new container',
```
jankeromnes: Nit: This description is useless (it doesn't add any information that the title doesn't). Please remove it.

api/hosts-api.js
```diff
@@ -336,9 +336,81 @@ hostAPI.get('/version', {
...
+
+  handler ({ user, query }, response) {
+    if (!user) {
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' }, null, 2);
+      return;
+    }
+
+    const projectId = query.project;
+    if (!db.get('projects').hasOwnProperty(projectId)) {
```
jankeromnes: Nit: In general, we check this like so: `if (!(projectId in db.get('projects'))`.

api/hosts-api.js
```diff
@@ -336,9 +336,81 @@ hostAPI.get('/version', {
...
+      response.statusCode = 400; // Invalid request
```
jankeromnes: Nit: The status is actually `Bad Request`.

api/hosts-api.js
```diff
@@ -336,9 +336,81 @@ hostAPI.get('/version', {
...
+      response.json({ error: 'Project does not exist.' });
```
jankeromnes: Nit: For things that don't exist or can't be found, we should use the response status `404; // Not Found`.

api/hosts-api.js
```diff
@@ -336,9 +336,81 @@ hostAPI.get('/version', {
...
+      return;
+    }
+
+    machines.spawn(user, projectId, (error, machine) => {
+      if (error) {
+        log('[fail] creating a new container', error);
```
jankeromnes: Nit: What failed is actually `machines.spawn`, not `docker.createContainer`. Please use more precise log messages like e.g. `[fail] could not spawn machine`.

api/hosts-api.js
```diff
@@ -336,9 +336,81 @@ hostAPI.get('/version', {
...
+        response.statusCode = 500; // Internal Error
+        response.json({ error: 'Could not create new container.' });
+        return;
+      }
+
+      response.json({
+        machineId: machine.id,
```
jankeromnes: Nit: We suffix variable names with `Id` internally to differentiate between a `machine` rich object, and a `machineId` value. However, since we never send rich objects over the API, there is no need to expose that suffix externally to API users. It's find to return things like:

```js
{
  "machine": 4,
  "container": "abcde12345678"
}
```

from our API (and that's actually what users expect).

api/hosts-api.js
```diff
@@ -336,9 +336,81 @@ hostAPI.get('/version', {
...
+    });
+  }
+});
+
 // API sub-resource to manage a single container on a cluster host.
 const containerAPI = hostAPI.api('/:container', 'Containers');
 
+containerAPI.delete({
+  title: 'Delete a specified container',
+  description: 'Delete a specified container',
```
jankeromnes: Nit: This description is useless.

api/hosts-api.js
```diff
@@ -441,8 +513,7 @@ containerAPI.get('/:port', {
   description: 'Get the properties of a given Docker container port.',
 
   handler: (request, response) => {
-    let { user } = request;
-    const { oauth2scope } = request;
+    let { user, oauth2scope } = request;
```
jankeromnes: Nit: You reverted a lint fix I made, which breaks lint again. Please don't do that.

nt1m:
@jankeromnes Addressed your comments!

jankeromnes:
Nit: You have handlers without `examples`, which breaks Travis. Please add `examples: []` to every handler you implement (or if you want, you can even add a simple example to each).

api/hosts-api.js
```diff
@@ -336,9 +336,79 @@ hostAPI.get('/version', {
...
+    if (!user) {
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' }, null, 2);
+      return;
+    }
+
+    const projectId = query.project;
+    if (!(projectId in db.get('projects'))) {
+      response.statusCode = 404; // Not Found
+      response.json({ error: 'Project does not exist.' });
```
jankeromnes: Nit: The convention in this code base appears to be `error: 'Project not found'` (it's shorter, and doesn't have a full stop `.`).

jankeromnes: Nit: `null, 2`.

api/hosts-api.js
```diff
@@ -336,9 +336,79 @@ hostAPI.get('/version', {
...
+      return;
+    }
+
+    machines.spawn(user, projectId, (error, machine) => {
+      if (error) {
+        log('[fail] could not spawn machine', error);
+        response.statusCode = 500; // Internal Error
```
jankeromnes: Nit: The correct status message is `Internal Server Error`: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_Server_errors

api/hosts-api.js
```diff
@@ -336,9 +336,79 @@ hostAPI.get('/version', {
...
+        response.json({ error: 'Could not create new container.' });
```
jankeromnes: Nit: Please remove `.` and add `null, 2` for human-readable JSON.

api/hosts-api.js
```diff
@@ -336,9 +336,79 @@ hostAPI.get('/version', {
   }]
 });
 
+hostAPI.post('/containers', {
+  title: 'Create a new container',
```
jankeromnes: Nit: Word `new` is useless (you can't create an old container).

api/hosts-api.js
```diff
@@ -336,9 +336,79 @@ hostAPI.get('/version', {
...
+      }, null, 2);
+    });
+  }
+});
+
 // API sub-resource to manage a single container on a cluster host.
 const containerAPI = hostAPI.api('/:container', 'Containers');
 
+containerAPI.delete({
+  title: 'Delete a specified container',
```
jankeromnes: Nit: Word `specified` container is useless (it would be pointless to delete a random container).

api/hosts-api.js
```diff
@@ -336,9 +336,79 @@ hostAPI.get('/version', {
   }]
 });
 
+hostAPI.post('/containers', {
+  title: 'Create a new container',
+
+  handler ({ user, query }, response) {
```
jankeromnes: I know that you like destructuring a lot 🙂 I personally prefer having consistent code throughout the code base, which makes it easier to understand how a handler works.

For consistency, I would prefer `handler: (request, response) => {`. Please at least make it an arrow function like everywhere else.

api/hosts-api.js
```diff
@@ -336,9 +336,79 @@ hostAPI.get('/version', {
...
+  }
+});
+
 // API sub-resource to manage a single container on a cluster host.
 const containerAPI = hostAPI.api('/:container', 'Containers');
 
+containerAPI.delete({
+  title: 'Delete a specified container',
+
+  handler ({ user, query }, response) {
```
jankeromnes: Same nit as above. Please consider destructuring `user` and `query` on a line below, and at least make it an arrow function.

Also, the reason I tend not to destructure `request.query` is because `query` is a very generic and thus ambiguous word (for example, it's easily confused with the `query` argument you get in `camp.route()` handlers. I feel like using `request.query` removes the ambiguity).

api/hosts-api.js
```diff
@@ -336,9 +336,79 @@ hostAPI.get('/version', {
...
+    if (!machine) {
+      response.statusCode = 404;
+      response.json({ error: 'Container not found' }, null, 2);
+      return;
+    }
+
+    machines.destroy(user, machine.project, machine, error => {
+      if (error) {
+        log('[fail] Destroying container', error);
+        response.statusCode = 500;
```
jankeromnes: Nit: `// Internal Server Error`.

api/hosts-api.js
```diff
@@ -336,9 +336,79 @@ hostAPI.get('/version', {
...
+    const machine = machines.getMachineByContainer(user, hostname, container);
+    if (!machine) {
+      response.statusCode = 404;
+      response.json({ error: 'Container not found' }, null, 2);
+      return;
+    }
+
+    machines.destroy(user, machine.project, machine, error => {
+      if (error) {
+        log('[fail] Destroying container', error);
```
jankeromnes: It's not very important, but we usually don't capitalize log messages. But if you prefer capitalization, I'm ok for gradually introducing this usage.

lib/users.js
```diff
@@ -281,6 +281,9 @@ function getOrCreateUser (email) {
         machine.properties = { name: machine.name };
         delete machine.name;
       }
+      if (!machine.project) {
```
jankeromnes: Nit: Please move the above comment about machine properties next to the `if (!machine.properties)` code, and add a comment above `if (!machine.project)` e.g.:

```js
// Temporary migration code: Previous machines didn't keep track of their project.
```

templates/projects.html
```diff
@@ -10,10 +10,14 @@
...
             <div class="project-actions">
-              <a class="btn btn-primary" {{if user then
-                {{data-action="spawn" data-project="{{= id in id}}" href="https://c9.io/open/ssh?name={{= id in id}}-{{= machines[id].id in integer}}&description={{= project.name in uri}}%20on%20janitor.technology&host={{= project.docker.host in uri}}&port={{= machines[id].docker.ports['22'].port in integer}}&user=user&workspaceDir={{= project.docker.path in uri}}&nodePath=%2Fhome%2Fuser%2F.c9%2Fnode%2Fbin%2Fnode" target="_blank"}}
-                else {{href="/login"}}
-              }}><span class="hidden-xs">Edit in </span>Cloud9</a>
+              {{if user then {{
+                <form action="/api/hosts/{{= project.docker.host in uri}}/containers?project={{= id in id}}" class="ajax-form" data-redirect-after-success="https://c9.io/open/ssh?name={{= id in id}}-{{= machines[id].id in integer}}&description={{= project.name in uri}}%20on%20janitor.technology&host={{= project.docker.host in uri}}&port={{= machines[id].docker.ports['22'].port in integer}}&user=user&workspaceDir={{= project.docker.path in uri}}&nodePath=%2Fhome%2Fuser%2F.c9%2Fnode%2Fbin%2Fnode" method="post">
```
jankeromnes: Waiting for container spawn (~1.7s) *before* redirecting to Cloud9 makes it even slower to access a working IDE (and it's not guaranteed that Cloud9 won't still show SSH connection problems). This significantly degrades user experience. I would prefer redirecting to a view of the container, and let the user click on Cloud9 or VNC or anything else once the container is created. (You could rename the creation button to something like `New Container`).

Additionally, the previous link used to open a separate Cloud9 tab, keeping your current tab open (which then redirects to the view of all containers once spawn is successful). You're changing this behavior, by completely navigating away from Janitor. I would prefer all external links (e.g. Cloud9, VNC, etc) to have `target="_blank"` by default.

jankeromnes: Follow-up note: In your case maybe it's easiest to simply rename the button, and use `data-redirect-after-success="/contributions/"`.

static/js/janitor.js
```diff
@@ -236,6 +242,14 @@ Array.forEach(document.querySelectorAll('.editable-label'), function (label) {
   });
 });
 
+// Setup modals
+$('.modal-form').on('show.bs.modal', function (event) {
+  var menuItem = $(event.relatedTarget);
+  $(this).find('.modal-title').text(menuItem.data('confirm'));
+  $(this).find('.modal-details').text(menuItem.data('details'));
+  $(this).attr('action', menuItem.data('form-action'));
+});
```
jankeromnes: Nit: What about the existing modal set-up code in https://github.com/JanitorTechnology/janitor/blob/89942ab8f34ec1c552cb0f667b1eaeb43ab93859/static/js/projects.js#L56-L100 ?

templates/contributions.html
```diff
@@ -49,21 +49,21 @@ <h2 class="header">My Contributions</h2>{{
...
         }}
       </div>
-      <div class="modal fade" id="confirm" tabindex="-1" role="dialog" aria-labelledby="confirm-title">
+      <form class="ajax-form modal-form modal fade" data-refresh-after-success="true" id="confirm" method="delete" tabindex="-1" role="dialog" aria-labelledby="confirm-title">
         <div class="modal-dialog" role="document">
           <div class="modal-content">
             <div class="modal-header">
-              <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
-              <h4 class="modal-title" id="confirm-title"></h4>
+              <a href="#" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></a>
```
jankeromnes: Was this change really necessary?

templates/contributions.html
```diff
@@ -49,21 +49,21 @@ <h2 class="header">My Contributions</h2>{{
...
+              <h4 class="modal-title"></h4>
             </div>
             <div class="modal-body">
-              <p id="confirm-details"></p>
+              <p class="modal-details"></p>
             </div>
             <div class="modal-footer">
-              <button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>
-              <button type="button" class="btn btn-danger" id="confirm-button"></button>
+              <a class="btn btn-default" data-dismiss="modal" href="#">Cancel</a>
```
jankeromnes: Was this change really necessary?

templates/contributions.html
```diff
@@ -30,7 +30,7 @@ <h2 class="header">My Contributions</h2>{{
                 <span class="glyphicon glyphicon-cog" aria-hidden="true"></span>
               </a>
               <ul class="dropdown-menu">
-                <li><a data-action="destroy" data-confirm="Delete {{= project.name in html}} #{{= id in id}}?" data-details="This action can't be undone, please make sure you've backed up everything valuable." data-machine={{= id in json}} data-project={{= projectId in json}} data-toggle="modal" data-target="#confirm" href="#">Delete</a></li>
```
jankeromnes: Yay for removing a `data-action` attribute! 💯

Edit: Sadly you're adding a `data-form-action`, which is only slightly less terrible. I'm not sure if there is a clean way to do this. Actually, I think we should re-implement Delete without a modal (e.g. in a follow-up pull request). Modals are not good UX practice anyway, I think we should do something like:
- collapse the machine panel down to a very thin line
- add a small text like `Warning: This container will be deleted in 10s...` and count down
- add a `cancel` link to this text, to replace the confirmation modal (yes I really really want to delete this) with a cancellable count down (oh no wait I'm not sure!)

Additionally, we really shouldn't make the user wait for our Docker server's response (the UI can just stop showing the container instantly upon deletion, and we can leave the Docker removal happen in the background).

static/js/janitor.js
nt1m: Forgot to remove it.

templates/contributions.html
nt1m: Yes, if you have 2 buttons inside a form, both buttons are submit buttons

api/hosts-api.js
nt1m: I don't think this is the first usage of this.

api/hosts-api.js
nt1m: There are some places where we use `handler (...) {` as well.

api/hosts-api.js
jankeromnes: > I don't think this is the first usage of this destructuring.

Maybe it's not, but I was just pointing out why I'm worried about it, and think we should avoid it where possible.

But as usual, code that's ready to ship should win over styling issues, otherwise the project doesn't move forward.

templates/contributions.html
jankeromnes: Doesn't look like it:

data:text/html,<form action="https://www.mozilla.org"><button type="button">submit not</button><button type="submit">submit</button></form>

api/hosts-api.js
jankeromnes: I think we should at least be consistent about it. Should we decide that shorter code wins, i.e. all:

```js
handler: (request, response) => {
```

should be replaced by:

```js
handler (request, response) {
```

?

templates/contributions.html
nt1m: Ah, ok, so I guess you have to include type="button" for it to work, will change that.

api/hosts-api.js
nt1m: To me, arrow functions are just as transparent for newcomers :) If we really wanted to be transparent, we should use `handler: function(request, response) {`. I don't think there's much of a win using explicit notation here. I'd rather use the short one.

nt1m:
@jankeromnes Updated

templates/projects.html
```diff
@@ -10,10 +10,14 @@
             <img class="project-icon" src="{{= project.icon in xmlattr}}" alt="{{= project.name in xmlattr}} Logo">
             <h4 class="project-title">{{= project.name in html}}</h4>
             <div class="project-actions">
-              <a class="btn btn-primary" {{if user then
-                {{data-action="spawn" data-project="{{= id in id}}" href="https://c9.io/open/ssh?name={{= id in id}}-{{= machines[id].id in integer}}&description={{= project.name in uri}}%20on%20janitor.technology&host={{= project.docker.host in uri}}&port={{= machines[id].docker.ports['22'].port in integer}}&user=user&workspaceDir={{= project.docker.path in uri}}&nodePath=%2Fhome%2Fuser%2F.c9%2Fnode%2Fbin%2Fnode" target="_blank"}}
-                else {{href="/login"}}
-              }}><span class="hidden-xs">Edit in </span>Cloud9</a>
+              {{if user then {{
```
jankeromnes: Nit: Please place the first `{{` at the end of the line above, in order not to add surprising whitespaces in the raw HTML output.

templates/projects.html
```diff
@@ -10,10 +10,14 @@
...
+                <form action="/api/hosts/{{= project.docker.host in uri}}/containers?project={{= id in id}}" class="ajax-form" data-redirect-after-success="/contributions/" method="post">
+                  <button class="btn btn-primary">New Container</button>
```
jankeromnes: Also, maybe add a `title` attribute, e.g. with the text `Create a new container for this project`?

templates/projects.html
```diff
@@ -10,10 +10,14 @@
             <img class="project-icon" src="{{= project.icon in xmlattr}}" alt="{{= project.name in xmlattr}} Logo">
             <h4 class="project-title">{{= project.name in html}}</h4>
             <div class="project-actions">
-              <a class="btn btn-primary" {{if user then
-                {{data-action="spawn" data-project="{{= id in id}}" href="https://c9.io/open/ssh?name={{= id in id}}-{{= machines[id].id in integer}}&description={{= project.name in uri}}%20on%20janitor.technology&host={{= project.docker.host in uri}}&port={{= machines[id].docker.ports['22'].port in integer}}&user=user&workspaceDir={{= project.docker.path in uri}}&nodePath=%2Fhome%2Fuser%2F.c9%2Fnode%2Fbin%2Fnode" target="_blank"}}
-                else {{href="/login"}}
-              }}><span class="hidden-xs">Edit in </span>Cloud9</a>
+              {{if user then {{
+                <form action="/api/hosts/{{= project.docker.host in uri}}/containers?project={{= id in id}}" class="ajax-form" data-redirect-after-success="/contributions/" method="post">
```
jankeromnes: Also, eventually it would be nice to highlight the container that was just created after the redirection, e.g. with something like `data-redirect-after-success="/contributions/#abcd1234"` (not sure if that's possible, and you don't have to do it now).

templates/projects.html
```diff
@@ -10,10 +10,14 @@
...
+                  <button class="btn btn-primary">New Container</button>
+                </form>
+              }}
```
jankeromnes: Nit: Please place the `}}` at the end of the line above, in order not to add surprising whitespaces to the raw HTML output.

templates/projects.html
```diff
@@ -10,10 +10,14 @@
...
+              else {{
+                <a class="btn btn-primary" href="/login">New Container</a>
+              }}}}
```
jankeromnes: Nit: Please place these `}}}}` at the end of the line above (same nit as earlier).

templates/projects.html
nt1m: Highlighting the container isn't hard, but doing this means that the "New container" form will start having their own submit handler.

api/hosts-api.js
```diff
@@ -336,9 +336,84 @@ hostAPI.get('/version', {
   }]
 });
 
+hostAPI.post('/containers', {
+  title: 'Create a container',
```
jankeromnes: Nit: This container-specific API handler is added in a weird place, before the API sub-resource that has the description "Containers". This causes the documentation to look like this:

containers-weird

Please fix this by creating a dedicated API-subresource for the path `/containers`. Then please move all container-specific handlers under it.

Note: This will change the `DELETE` and `PATCH` handlers from `/api/hosts/:hostname/:container` to `/api/hosts/:hostname/containers/:container`, and the `GET /api/hosts/:hostname/:container/:port` handler to `GET /api/hosts/:hostname/containers/:container/:port`, so you'll need to update all the places that use it (hint: you can find them by running `git grep '/api/hosts/'`).

api/hosts-api.js
```diff
@@ -336,9 +336,84 @@ hostAPI.get('/version', {
...
+    machines.spawn(user, projectId, (error, machine) => {
+      if (error) {
+        log('[fail] could not spawn machine', error);
+        response.statusCode = 500; // Internal Server Error
+        response.json({ error: 'Could not create new container' }, null, 2);
+        return;
+      }
+
+      response.json({
+        machine: machine.id,
```
jankeromnes: Nit: This internal machine ID probably isn't useful to anyone (and we should remove it at some point). It's currently being used in conjunction with `machine.project` to identify a machine, but instead we should probably move to using `machine.docker.host` and `machine.docker.container` instead. I.e., I think you can remove the `machine: machine.id` entry (otherwise some API users could start relying on it!).

static/js/janitor.js
```diff
@@ -236,6 +242,14 @@ Array.forEach(document.querySelectorAll('.editable-label'), function (label) {
   });
 });
 
+// Setup modals
+$('.modal-form').on('show.bs.modal', function (event) {
+  var menuItem = $(event.relatedTarget);
+  $(this).find('.modal-title').text(menuItem.data('confirm'));
+  $(this).find('.modal-details').text(menuItem.data('details'));
+  $(this).attr('action', menuItem.data('form-action'));
```
nt1m: @jankeromnes Tbh, `$('.modal-form').on('show.bs.modal', function (event) {` is already unavoidable jQuery. So removing the modal code (which I'll try to do with the new contributions page), will remove this jQuery anyway. I don't think it's worth spending too much time on this code.

What do you think ?

jankeromnes: Fair enough! ☺️

jankeromnes:
Please also rebase on top of the latest changes

nt1m:
@jankeromnes Updated and rebased.

api/hosts-api.js
```diff
@@ -336,8 +336,45 @@ hostAPI.get('/version', {
   }]
 });
 
+const containersAPI = hostAPI.api('/containers', 'Containers');
```
jankeromnes: Nit: Please add the usual comment above this line, e.g. `API sub-resource to manage Docker containers on a cluster host.`.

api/hosts-api.js
```diff
@@ -436,6 +473,45 @@ containerAPI.patch({
...
+    }
+
+    machines.destroy(user, machine.project, machine, error => {
+      if (error) {
+        log('[fail] destroying machine', error);
+        response.statusCode = 500; // Internal Server Error
+        response.json({ error: 'Could not delete container' }, null, 2);
+        return;
+      }
+      response.end();
```
jankeromnes: Nit: Please use `response.statusCode = 204; // No Content` when calling `response.end()` directly, especially when the handler is about deleting something (204 means "deleted successfully" in several APIs I know, e.g. GitHub's).

jankeromnes: For an example in Janitor's API, see https://janitor.technology/reference/api#reset-a-user-configuration

api/hosts-api.js
```diff
@@ -336,8 +336,45 @@ hostAPI.get('/version', {
...
+
+  handler (request, response) {
+    const { user } = request;
+    if (!user) {
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' }, null, 2);
+      return;
+    }
+
+    const projectId = request.query.project;
```
jankeromnes: Note: You just implemented a `POST` handler that doesn't care about any data being POSTed to it (because it only supports the `project` query parameter). Nothing wrong with it, but it might confuse API implementers because it's not common. (I'm not sure it's worth it to support JSON bodies like `{ project: 'firefox' }`, but maybe some users will expect that at some point and get confused.)

PR #124, opened 2017-07-31
Add temporary proxy heuristics to mitigate ambiguous requests.

jankeromnes:
We should remove them once Let's Encrypt supports wildcard certificates:
https://letsencrypt.org/2017/07/06/wildcard-certificates-coming-jan-2018.html

They should become available around January 2018, and will allow us to enforce
explicit proxy requests by specifying container ID and port in domains like:
  'https://8088.abc123.<hostname>/index.html'
instead of the prefixed URLs we use today (which don't work with WebSockets):
  'https://<hostname>/abc123/8080/index.html'

### DIFF
--- a/lib/proxy-heuristics.js
+++ b/lib/proxy-heuristics.js
@@ -1,0 +1,176 @@
+// Copyright © 2017 Jan Keromnes. All rights reserved.
+// The following code is covered by the AGPL-3.0 license.
+
+const nodeurl = require('url');
+
+const log = require('./log');
+const routes = require('./routes');
+
+// In this regex, we expect:
+// - a 16+ hex-digit container ID.
+// - a numeric port.
+// - an optional path that starts with a '/'.
+// These anonymous patterns will be captured in a `match` array as `match[1]`,
+// `match[2]` and maybe `match[3]`.
+exports.proxyUrlPrefix = /^\/([0-9a-f]{16,})\/(\d+)(\/.*)?$/;
+
+// Parse request URLs (or referer URLs) that look like '/:container/:port/*' to
+// identify which container ID and port a request should be proxied to.
+exports.handleProxyUrls = function (request, response, next) {
+  let url = null;
+  let containerId = null;
+  let port = null;
+  let path = null;
+
+  // Look for a container ID and port in `request.url`.
+  let match = exports.proxyUrlPrefix.exec(request.url);
+  if (match) {
+    [url, containerId, port, path] = match;
+
+    // We want the proxied `path` to always begin with a '/'.
+    // However `path` is empty in URLs like '/abc123/8080?p=1', so we redirect
+    // them to '/abc123/8080/?p=1' (where `path` is '/').
+    if (!path) {
+      url = url.includes('?') ? url.replace('?', '/?') : url + '/';
+      routes.redirect(response, url, true);
+      return;
+    }
+
+    // Locally remove the prefix from `request.url`.
+    request.url = request.url.replace('/' + containerId + '/' + port, '');
+  } else if (request.headers.referer) {
+    // Look for a container ID and port in `request.headers.referer`.
+    const referer = nodeurl.parse(request.headers.referer);
+    match = exports.proxyUrlPrefix.exec(referer.pathname);
+    if (match) {
+      [url, containerId, port, path] = match;
+    }
+  }
+
+  if (containerId && port) {
+    // Add the requested container ID and port to `request.query`.
+    request.query.container = containerId;
+    request.query.port = port;
+  }
+
+  next();
+};
+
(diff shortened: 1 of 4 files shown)


lib/proxy-heuristics.js
```diff
@@ -0,0 +1,170 @@
...
+    if (this._cloud9VfsUrlPrefix) {
+      return url.startsWith(this._cloud9VfsUrlPrefix);
+    }
+    const match = exports.cloud9VfsUrlPrefix.exec(url);
+    if (match) {
+      this._cloud9VfsUrlPrefix = match[0];
+      return true;
+    }
+    return false;
+  }
```
nt1m: It should probably handle port :8080 as well for Janitor

jankeromnes: This is a bit harder, because Janitor uses absolute URLs everywhere, so when you click on any link, you remove the proxy URL prefix, and then *any* follow-up request is ambiguous.

Implementing a heuristic would make us list basically every URL Janitor can use. Instead, I opted for not implementing a heuristic, and letting the simple behavior "use whatever container + port was used last (that is not Cloud9 or noVNC)" do its thing.

In the future, we should find a way to make every internal Janitor URL relative, in order to keep any URL prefix in place. (Kresus already does this, so no requests from Kresus are ever ambiguous, which is pretty awesome!)

jankeromnes:
Nit: This makes the deprecated in-app proxy not handle URLs like /vfs/1/9ceokVZPKGlhYWec/workspace/.c9/state.settings

Maybe there is a way to support some of these too? (I'm thinking of a middleware that tests heuristic functions on URLs).

jankeromnes:
> Nit: This makes the deprecated in-app proxy not handle URLs like /vfs/1/9ceokVZPKGlhYWec/workspace/.c9/state.settings

Fixed by routing requests that match `proxyHeuristics.cloud9VfsUrlPrefix`.

> Maybe there is a way to support some of these too? (I'm thinking of a middleware that tests heuristic functions on URLs).

No need to go quite this far for a deprecated proxy that should be removed (once we don't host containers on the same server as the web app anymore).

lib/proxy-heuristics.js
```diff
@@ -0,0 +1,176 @@
...
+        // We already knew this request. Let's forget the old one.
+        pastRequests.splice(i, 1); // Remove 1 item at position i.
+        break;
+      }
+    }
+  }
+
+  // If we have a likeliness heuristic function for this port, remember it too.
+  const likely = exports.requestLikelyForPort[port];
+  pastRequests.unshift({ container, port, likely });
```
nt1m: rememberProxyRequest adds an object of type { container, port, likely }  and guessProxyRequest modifies it to become { index, container, port, score }.

It seems a bit strange to have the data structure arbitrarily changing in guessProxyRequest.

lib/proxy-heuristics.js
```diff
@@ -0,0 +1,176 @@
...
+  // Compute the likeliness of each previously requested container port for this
+  // new ambiguous request.
+  // Scores:
+  //    1: likely
+  //    0: neutral
+  //   -1: unlikely
+  pastRequests = pastRequests.map((pastRequest, index) => {
+    let score = 0;
+    if (typeof pastRequest.likely === 'function') {
+      score = pastRequest.likely(request) ? 1 : -1;
```
nt1m: If I understand correctly, pastRequest.likely will disappear after this mapping, which feels strange. Not sure if that's expected.

lib/proxy-heuristics.js
jankeromnes: `guessProxyRequest` doesn't actually modify any entry in `pastRequests`. It creates local copies to process them.

lib/proxy-heuristics.js
jankeromnes: It doesn't disappear, it's just not included in the local copy resulting from the `map` call. Maybe this code would be less confusing if I didn't re-assign different things to the same variable `pastRequest`, but created separate intermediary variables for each copied collection.

lib/proxy-heuristics.js
nt1m: Nevermind, I misread the code. I shouldn't review things during evenings then comment the next day :)

lib/proxy-heuristics.js
nt1m: I agree, maybe the second `pastRequests` should be called  `sortedRequests` or something like that.

lib/proxy-heuristics.js
jankeromnes: I made my code a little clearer by creating an intermediary variable for the result from `map()` instead of re-assigning. Thanks for complaining about this! 😄

PR #125, opened 2017-08-01
Remove deprecated project rebuild feature.

jankeromnes:
We can now use `Pull` to download up-to-date project Docker images from Docker Hub, instead of building them ourself. (We were doing that manually anyway, since our admin interface doesn't support local `ADD` Dockerfile instructions).

It's still possible to build incremental update layers using the `Update` button, and this can be a good option to provide more frequent project updates and save disk space by limiting the number of different base image versions used in production.

### DIFF
--- a/app.js
+++ b/app.js
@@ -423,28 +423,6 @@ boot.executeInParallel([
     end({ status: 'success' });
   });
 
-  // Rebuild the base image of a project.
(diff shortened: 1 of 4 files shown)


jankeromnes:
Yaaay 😄 thanks!

PR #129, opened 2017-08-03
Make the Docker host administration page editable again.

jankeromnes:
The Docker hosts administration section became read-only when the old Ajax API was deprecated.

This change updates it to the new JSON API using `window.fetch`, and is helpful to manage Docker hosts (e.g. to add new remote hosts, or to update the TLS certificates, etc).

@nt1m, please have a look to double-check I haven't made any mistakes, or forgotten anything while updating this page. (In theory, it works, because I was successful at using it locally to set up and remote control janitor.technology's Docker host).

### DIFF
--- a/static/js/admin.js
+++ b/static/js/admin.js
@@ -22,6 +22,25 @@ ajaxForm('#invite-form', 'invite', function (form, data) {
   updateFormStatus(form, status, message);
 });
 
+// New host form.
+var newHostForm = document.querySelector('#newhost-form');
+if (newHostForm) {
+  window.setupAsyncForm(newHostForm);
+  newHostForm.addEventListener('submit', function (event) {
+    var hostname = newHostForm.elements.hostname.value;
+    window.fetchAPI('POST', '/api/hosts/' + hostname, {}, function (error, data) {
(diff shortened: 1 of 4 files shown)


static/js/admin.js
```diff
@@ -22,6 +22,25 @@ ajaxForm('#invite-form', 'invite', function (form, data) {
   updateFormStatus(form, status, message);
 });
 
+// New host form.
+var newHostForm = document.querySelector('#newhost-form');
+if (newHostForm) {
```
nt1m: Doesn't the ajax-form class handle this already?

jankeromnes: Actually no, the `.ajax-form` class doesn't add any `'submit'` handlers. It's just there for styling (e.g. for feedback things).

The [code that automatically sets up Janitor forms](https://github.com/JanitorTechnology/janitor/blob/1bdc7de2861c2df82178e4a4ffa04cde8fc00edb/static/js/janitor.js#L12) works by selecting on the `form[method]` query selector.

You'll notice that this `#newhost-form` element doesn't have a `method` attribute. I agree this is quite subtle.

PR #130, opened 2017-08-03
Add an API to list Docker images and containers on a host.

jankeromnes:
These administrator-only API handlers will be useful in order to find out what's taking so much disk space on a Docker host.

I'm hoping to use it in a follow-up pull request to build a Docker admin interface that shows which images and containers take up the most disk space (e.g. as a tree), and that allows inspecting containers (to see if they contain anything valuable) and deleting them if they're useless to free up disk space.

### DIFF
--- a/api/hosts-api.js
+++ b/api/hosts-api.js
@@ -87,7 +87,7 @@ hostAPI.get({
       }
     }
 
-    response.statusCode = 404;
+    response.statusCode = 404; // Not Found
     response.json({ error: 'Host not found' }, null, 2);
   },
 
@@ -228,14 +228,14 @@ credentialsAPI.get({
   handler: (request, response) => {
     const { user } = request;
     if (!users.isAdmin(user)) {
-      response.statusCode = 404;
+      response.statusCode = 404; // Not Found
       response.json({ error: 'Host not found' }, null, 2);
       return;
     }
 
     const host = hosts.get(request.query.hostname);
     if (!host) {
-      response.statusCode = 404;
+      response.statusCode = 404; // Not Found
       response.json({ error: 'Host not found' }, null, 2);
       return;
     }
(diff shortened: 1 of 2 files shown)


api/hosts-api.js
```diff
@@ -289,6 +289,39 @@ credentialsAPI.delete({
...
+    const { user } = request;
+    if (!users.isAdmin(user)) {
+      response.statusCode = 404;
+      response.json({ error: 'Host not found' }, null, 2);
+      return;
+    }
+
+    const { hostname } = request.query;
+    if (!hosts.get(hostname)) {
+      response.statusCode = 404;
```
nt1m: nit: // Not Found

api/hosts-api.js
```diff
@@ -289,6 +289,39 @@ credentialsAPI.delete({
...
 });
 
+hostAPI.get('/images', {
+  title: 'List host images',
+  description: 'List the Docker images available on this host.',
+
+  handler: (request, response) => {
+    const { user } = request;
+    if (!users.isAdmin(user)) {
+      response.statusCode = 404;
```
nt1m: nit: // Not Found

nt1m: I think it would make more sense to send a 403 forbidden response here, with the error describing you should be an admin.

api/hosts-api.js
```diff
@@ -339,6 +372,39 @@ hostAPI.get('/version', {
...
 
+containersAPI.get({
+  title: 'List containers',
+  description: 'List all Docker containers on this host.',
+
+  handler: (request, response) => {
+    const { user } = request;
+    if (!users.isAdmin(user)) {
+      response.statusCode = 404;
+      response.json({ error: 'Host not found' }, null, 2);
```
nt1m: Same comment as above.

api/hosts-api.js
```diff
@@ -339,6 +372,39 @@ hostAPI.get('/version', {
...
+      return;
+    }
+
+    const { hostname } = request.query;
+    if (!hosts.get(hostname)) {
+      response.statusCode = 404;
```
nt1m: nit: // Not Found

api/hosts-api.js
jankeromnes: The problem with replying with `403` is that we'll leak the fact that this particular host exists (because non-admin API users could try many different hosts, and look for `404` and `403` responses).

Not sure this is very dramatic though.

PR #131, opened 2017-08-03
Add an administration section for Docker images and containers.

jankeromnes:
This new administration interface will help analyzing the disk space captured by older containers that prevent older images from being garbage collected.

I will also add a "Garbage Collect" button to this interface soon.

### DIFF
--- a/static/css/janitor.css
+++ b/static/css/janitor.css
@@ -521,6 +521,10 @@ h4 {
   max-width: 100%;
 }
 
+.docker-tree pre {
+  line-height: 1.19; /* The Docker containers ASCII tree shows gaps > 1.20 */
+}
+
 /* Sections */
 
 section.alert {
--- a/static/js/admin.js
+++ b/static/js/admin.js
@@ -69,3 +69,126 @@ Array.map(document.querySelectorAll('button[data-action]'), function (button) {
     };
   }));
 });
+
+// Docker version button.
+Array.forEach(document.querySelectorAll('.docker-version'), function (div) {
+  var button = div.querySelector('button');
+  var pre = div.querySelector('pre');
+  button.addEventListener('click', function (event) {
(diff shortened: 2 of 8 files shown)


static/css/janitor.css
```diff
@@ -511,6 +511,10 @@ h4 {
   max-width: 100%;
 }
 
+.docker-tree pre {
+  line-height: 1.19;
```
nt1m: 😮

static/js/admin.js
```diff
@@ -69,3 +69,126 @@ Array.map(document.querySelectorAll('button[data-action]'), function (button) {
     };
   }));
 });
+
+// Docker version button.
+Array.map(document.querySelectorAll('.docker-version'), function (div) {
```
nt1m: nit: Array.forEach

static/js/admin.js
```diff
@@ -69,3 +69,126 @@ Array.map(document.querySelectorAll('button[data-action]'), function (button) {
...
+        pre.textContent = error;
+        return;
+      }
+      pre.textContent = JSON.stringify(version, null, 2);
+    });
+  });
+});
+
+// Docker Container tree button.
+Array.map(document.querySelectorAll('.docker-tree'), function (div) {
```
nt1m: nit: Array.forEach

static/css/janitor.css
jankeromnes: This is to format the Docker container tree `<pre>` correctly (otherwise there are spaces between vertical `|`s).

nt1m: No, I mean, why 1.19 ? I don't like magic values.

jankeromnes: It's because at 1.20 the gaps become visible and the container tree doesn't look so good. I'll add a comment to explain this next to the magic value.

jankeromnes:
All three nits addressed!

Issue #128, opened 2017-08-02
Show when async actions are pending when you clicked on a button


jankeromnes:
Today, when you click on a button on Janitor, chances are you won't see any UI changes for anywhere between a few seconds to like a minute or so (after that, when the server replies, you see whether the action has succeeded or failed, and you might see a message box appear with more information).

This is bad UX, because users don't know if the click worked, or if the server is doing anything at all. This caused some users to take on the habit of clicking some buttons multiple times, "just to be sure".

Instead, once an async action has been triggered, we should show a pending state (e.g. a spinner), and disable the button until the server replies (or the action fails some other way).

This should be pretty easy to do, by adding a spinner to the auto-generated `.form-control-feedback` element, and by showing it and disabling all form controls whenever an action is triggered.

If anyone is interested in taking this, I can mentor this bug, and show you where in the code this fix could be implemented.

jankeromnes:
Partly addressed in 50a470de401926c3557a029298305caac371ac4d (by disabling buttons and inputs when a form is pending).

arshdkhn1:
@jankeromnes I'll do it

PR #132, opened 2017-08-04
Disable form elements when awaiting a server response.

jankeromnes:


jankeromnes:
This partly addresses #128 by disabling form buttons and input elements when you submit asynchronously. Everything gets re-enabled when the form status changes (i.e. success or error).

Some users took the habit of clicking on our buttons multiple times. This change addresses this as well.

jankeromnes:
@nt1m Please have a look.

static/js/janitor.js
```diff
@@ -155,8 +155,16 @@ function setupAsyncForm (form) {
     addFormFeedback(form);
   }
 
-  // Ensure that submitting the <form> doesn't reload the page.
   form.addEventListener('submit', function (event) {
+    // Disable all form elements while waiting for the server.
+    Array.map(form.elements, function (element) {
```
nt1m: nit: use forEach, not map.

nt1m: Also, while you're changing this, can you update https://github.com/jankeromnes/janitor/blob/cc7bf393eff4dde18697b84cb85f6839ec1af0d9/static/js/janitor.js#L183 as well ?

jankeromnes: Done! 👍

PR #134, opened 2017-08-04
Add temporary machines support

nt1m:


nt1m:
@bnjbvr @jankeromnes Not sure who can get to this first, but I'd like to have some feedback on this PR.

So this PR allows generating temporary containers (it seems to work in my local testing).
There's no elaborate UI atm, it just uses the Projects page, no way to get a link to your container or anything atm. I plan to add this later.

I have some questions as well:
- What should happen if a task failed ? Atm, the PR just removes the task, which doesn't seem like a great idea, because then we'd have some containers hanging around.
- How do you access VNC/C9 for containers spawned from a Janitor container ? I tried the usual `host/containerId/port`, with the host being the one I set on the admin page, but it didn't work.

jankeromnes:
Will try to have a look today! Thanks a lot for doing this 😄 👍 💯

jankeromnes:
Ok, I'll have a look at your code.

> What should happen if a task failed ? Atm, the PR just removes the task, which doesn't seem like a great idea, because then we'd have some containers hanging around.

Without having looked at the code yet, this sounds unrelated to temporary containers. Maybe it's better to make a separate pull request that implements tasks? (It's ok to land it while it's unused, and start using it in a follow-up pull request).

> How do you access VNC/C9 for containers spawned from a Janitor container ?

I don't understand this question. Why would you spawn containers "from" an existing container?

Also, in general you can access VNC, C9, and any other web service running inside a container by using the usual https://hostname/containerId/port/ URLs. I'll look at your code to see if your question makes more sense.

api/hosts-api.js
```diff
@@ -409,20 +409,31 @@ containersAPI.put({
...
+    const { user, session } = request;
+    if (!user) {
+      if (!session) {
+        return;
+      }
+
+      machines.spawnTemporary(session.id, projectId, (error, machine) => {
+        if (error) {
+          response.statusCode = 500; // Internal Server Error
+          response.json({ error: 'Could not create container' }, null, 2);
```
jankeromnes: Might want to return after than, in order not to end a response twice.

api/hosts-api.js
```diff
@@ -409,20 +409,31 @@ containersAPI.put({
...
     }
 
+    const { user, session } = request;
+    if (!user) {
+      if (!session) {
+        return;
+      }
+
+      machines.spawnTemporary(session.id, projectId, (error, machine) => {
+        if (error) {
```
jankeromnes: Maybe log the error to know what's happening? E.g. `log('[fail] creating temporary container', error)`.

lib/boot.js
```diff
@@ -345,3 +346,13 @@ exports.registerDockerClient = function (next) {
     next();
   });
 };
+
+exports.loadTasks = function () {
```
jankeromnes: Nit: Boot tasks usually take a `next` callback function. Might want to accept and call it here, just to stay consistent and avoid surprises.

lib/tasks.js
```diff
@@ -0,0 +1,53 @@
...
+    throw new Error('[fail] task', task, 'already exists');
+  }
+
+  exports.taskTypes.set(type, task);
+};
+
+exports.add = function (date, type, data) {
+  taskCounter++;
+
+  const dateHash = date.getTime();
```
jankeromnes: Nit: This is not a hash.

lib/tasks.js
```diff
@@ -0,0 +1,53 @@
+const db = require('./db');
+const log = require('./log');
+let taskCounter = 0; // Used to generate unique task IDs
```
jankeromnes: Nit: I would really prefer tasks to be a separate pull request.

api/hosts-api.js
nt1m: Whoops, good catch!

lib/tasks.js
nt1m: Whoops, the var name is a leftover from some older code I wrote.

nt1m:
> I don't understand this question. Why would you spawn containers "from" an existing container?

Not sure how to explain this. I spawned some containers using the "local" janitor instance I had, but I can't access it using `https://hostname/containerId/port/`. The host of the project was set to janitor.technology.

api/hosts-api.js
nt1m: It's already logged inside `_spawn` in machines.js

lib/tasks.js
nt1m: Moved to #140

bnjbvr:
(removing myself from reviewers since Jan seems to be on it; thanks for doing that!)

api/hosts-api.js
```diff
@@ -409,20 +409,32 @@ containersAPI.put({
   title: 'Create a container',
 
   handler (request, response) {
-    const { user } = request;
```
nt1m: trst

nt1m: test

PR #135, opened 2017-08-04
Transition Cloud9 IDE links from c9.io to the SDK installed in each container.

jankeromnes:
## An update on how we use Cloud9 IDE for Janitor

We consider that every Janitor container, based on an image that was updated after August 2nd 2017, will ship with a working Cloud9 SDK running on port 8089.

However, we allow for a smooth transition period, by keeping a link to access Cloud9 SSH workspaces on the third-party service https://c9.io until September 1st 2017.

After September 1st, any pre-existing c9.io SSH workspaces will still work, but we'd like our users to not create new ones, and use the Cloud9 SDK on port 8089 instead (which is both faster to load and more stable).

Eventually, we'd like to stop relying on the https://c9.io third-party service, in order to support only our customized Cloud9 SDK. This will greatly simplify user registration, container creation, IDE access, IDE customization (with automated workflows for each supported project), and more.

We would like to warmly thank the [Cloud9 IDE](https://c9.io) team for supporting Janitor so very generously by letting us use this incredible product for free, and for their relentless help and guidance in helping us install, optimize and personalize the Cloud9 SDK for Janitor.

You guys rock! 👍 💯 🥇 🎆 ❤️ 🚀 ☁️

### DIFF
--- a/templates/contributions.html
+++ b/templates/contributions.html
@@ -27,13 +27,23 @@
               <a class="btn btn-default dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                 <span class="glyphicon glyphicon-cog" aria-hidden="true"></span>
               </a>
-              <ul class="dropdown-menu">
+              <ul class="dropdown-menu">{{
+                const updated = new Date(machine.data.updated);
+                const c9sdkStart = new Date('2017-08-02');
+                const c9ioEnd = new Date('2017-09-02');
+                if (updated > c9sdkStart && updated < c9ioEnd) { }}
+                <li>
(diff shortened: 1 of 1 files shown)


templates/contributions.html
```diff
@@ -2,11 +2,14 @@
...
-              let machine = machines[id];
-              if (machine.status === "new") continue;
+            const machines = user.machines[projectId];
+            const project = projects[projectId];
+            for (const id in machines) {
+              const machine = machines[id];
+              if (machine.status === "new") {
+                continue;
+              }
+              const name = machine.properties.name || (project.name + id);
```
nt1m: project.name + id will return `Firefox1` instead of `Firefox #1`, not sure if that's wanted

jankeromnes: Whoa, nice catch!

jankeromnes:
By the way, thanks a lot for the quick review! 👍

PR #136, opened 2017-08-04
A few small fixes.

jankeromnes:
In this pull request, I:
- fix old container names
- slightly improve the container API handlers and their documentation
- make the pages `/` and `/projects/` load faster

### DIFF
--- a/api/hosts-api.js
+++ b/api/hosts-api.js
@@ -441,7 +441,7 @@ containersAPI.put({
 });
 
 // API sub-resource to manage a single container on a cluster host.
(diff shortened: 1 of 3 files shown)


api/hosts-api.js
```diff
@@ -441,7 +441,47 @@ containersAPI.put({
 });
 
 // API sub-resource to manage a single container on a cluster host.
-const containerAPI = containersAPI.api('/:container', 'Container');
+const containerAPI = containersAPI.api('/:container');
+
+containerAPI.delete({
```
nt1m: I like having `patch` before `delete`

PR #139, opened 2017-08-04
Don't auto-blur form elements upon submit.

### LINKED ISSUE #137: Auto blurring textbox and input makes keyboard warriors life difficult
Hi, textbox and input automatically get blur event after validating the text entered by the user.  If the user tries to copy all the text or paste new text then instead of copy/pasting the desired content whole body element is selected. Check out this video - https://youtu.be/m5nuDth46sA

jankeromnes:
Should fix #137.

### DIFF
--- a/static/js/janitor.js
+++ b/static/js/janitor.js
@@ -79,7 +79,6 @@ function ajaxForm (selector, action, callback) {
       callback(form, data);
     };
     Array.forEach(form.elements, function (element) {
-      element.blur();
       element.classList.add('disabled');
     });
   });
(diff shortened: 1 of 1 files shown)


arshdkhn1:
@jankeromnes do you know why blur event is called as soon as I press control key?

arshdkhn1:
@nt1m https://github.com/jankeromnes/janitor/blob/ec1b1f7f0c955d18047be8460a75e46fbd33bf60/static/js/janitor.js#L159

this part of code executes when submit event is called on form.

The auto blurring problem exists even when no form submission has taken place, that's why i doubt that these two lines are the reason of auto blurrin on settings page.

arshdkhn1:
@jankeromnes if you are on mac, try `cmd` key instead of `cntrl`.

PR #140, opened 2017-08-08
Add support for tasks

nt1m:


nt1m:
@nt1m: review?(@jankeromnes)

lib/boot.js
```diff
@@ -345,3 +346,10 @@ exports.registerDockerClient = function (next) {
     next();
   });
 };
+
+exports.loadTasks = function (callback) {
```
bnjbvr: nit: common pattern is to name the argument `next` here

lib/tasks.js
```diff
@@ -0,0 +1,53 @@
+const db = require('./db');
+const log = require('./log');
+let taskCounter = 0; // Used to generate unique task IDs
+
+exports.tasks = db.get('tasks');
```
bnjbvr: General comment for this file: why are so many things exported and not used outside this module? I'd rather have them unexported until they're actually used elsewhere. This is also a big confusing in my opinion to export fields of the database file.

lib/tasks.js
```diff
@@ -0,0 +1,53 @@
...
+exports.taskTypes = new Map();
+
+exports.check = function () {
+  const now = Date.now();
+  for (const task of Object.values(exports.tasks)) {
+    if (now > Number(task.date)) {
```
bnjbvr: nit: yoda condition: please invert LHS and RHS and replace `>` by `<`

lib/tasks.js
```diff
@@ -0,0 +1,53 @@
...
+  return task;
+};
+
+exports.remove = function (id) {
+  delete exports.tasks[id];
+  db.save();
+};
+
+exports.execute = function ({ type, data, id }) {
+  const task = exports.taskTypes.get(type);
```
bnjbvr: It seems that `task` is actually a function, so `taskTypes` sounds poorly named. Can you think of a better name here?

lib/tasks.js
```diff
@@ -0,0 +1,53 @@
...
+  log('[ok] Task', id, 'executed', data);
+  task(data);
+  exports.remove(id);
```
bnjbvr: Can a task be fallible? If so, shouldn't it use a callback form instead, and the task be removed only if it succeeded?

bnjbvr: Also, can a task remove itself or add new tasks to the list of pending tasks? There should probably be a comment explaining this somewhere in this file.

lib/tasks.js
```diff
@@ -0,0 +1,53 @@
...
+    throw new Error('[fail] task', task, 'already exists');
+  }
+
+  exports.taskTypes.set(type, task);
+};
+
+exports.add = function (date, type, data) {
+  taskCounter++;
+
+  const msSince1970 = date.getTime();
```
bnjbvr: nit: rename `msSince1970` to `timestamp` maybe?

lib/tasks.js
```diff
@@ -0,0 +1,53 @@
...
+  for (const task of Object.values(exports.tasks)) {
+    if (now > Number(task.date)) {
+      exports.execute(task);
+    }
+  }
+};
+
+exports.addType = function (type, task) {
+  if (exports.taskTypes.has(type)) {
+    throw new Error('[fail] task', task, 'already exists');
```
bnjbvr: nit: pretty sure the args to the `Error` ctor aren't doing what you expect here, so you'll need to use a template string here instead.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

nt1m:
@bnjbvr Addressed your comments, thanks for the review!

lib/tasks.js
bnjbvr: (not addressed)

lib/tasks.js
bnjbvr: (still present in the template string next line)

lib/tasks.js
bnjbvr: (not addressed)

lib/tasks.js
bnjbvr: (not addressed)

bnjbvr:
@nt1m Some previous comments have not been addressed, can you comment in the code or at least in a Github comment the choices you've made?

jankeromnes:
Hello! If you don't mind, I'd like to take over this feature in #236, so I'll close this unmerged pull request. Sorry for duplicating this great work!

Issue #141, opened 2017-08-09
Add traits to machine properties


nt1m:
It would be nice to add support for `traits` in each machine object for easy backwards compatibility (instead of using the machine updated date).

For example: traits: ["c9sdk", "web-terminal"]

It would simplify the templates for contributions.

nt1m:
I'll work on this issue.

jankeromnes:
I don't think it's worth implementing such a complex feature, because the contributions template is only complex temporarily (until we've fully switched to c9sdk). The same goes for web-terminal, at some point we'll support it in all containers, and the traits feature might become unused.

However, if you feel this would be really useful to your work, we should at least use a less ambiguous name than "traits" (which can mean a lot of things). Maybe "supports"?

Additionally, how do you know which traits to associate to a container? Is it tired to a project?

In general, it's not clear to me that this feature is useful, not that it can be implemented in a sane way. Just my opinion though, please let me know what you think.

nt1m:
Traits can be associated to projects and containers.

For example, if you add a new tool to newer containers that old containers don't have and you later integrate it to the Janitor interface, it's very convenient to have traits, so you can actually hide the integration easily without using a Date.now() hack.

Another case where traits is useful is for the KDE docker image, which doesn't have c9sdk or web-terminal.

PR #142, opened 2017-08-09
Add support for machine traits

nt1m:


nt1m:
See #141 for more context.

PR #148, opened 2017-08-17
Fix feedback when deleting containers

### LINKED ISSUE #144: No feedback while deleting containers and delete requests gets 404 response
I was trying to delete containers and it was really a bad experience deleting them. There wasn't any page reload and feedback, all I had to do was to reload page manually in order to see that containers were deleted or not.

nt1m:
Fixes #144

### DIFF
--- a/lib/machines.js
+++ b/lib/machines.js
@@ -205,8 +205,6 @@ exports.spawn = function (user, projectId, callback) {
   docker.runContainer({ host, image, ports }, (error, container, logs) => {
     if (error) {
       log('spawn', image, error);
-      machine.status = 'start-failed';
-      db.save();
       callback(new Error('Unable to start machine for project: ' + projectId));
       return;
     }
--- a/static/js/janitor.js
+++ b/static/js/janitor.js
@@ -177,12 +177,11 @@ function updateFormStatus (form, status, message) {
     case 'error':
       form.classList.add('has-error');
       break;
-    default:
-      Array.forEach(form.elements, function (element) {
-        element.classList.remove('disabled');
-      });
-      break;
   }
+
+  Array.forEach(form.elements, function (element) {
+    element.classList.remove('disabled');
+  });
 
   var feedback = form.querySelector('.form-control-feedback');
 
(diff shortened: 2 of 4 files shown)


nt1m:
Fixes #144.

Containers that failed to create now are recycled instead of having a "Start failed" status, because the API now handles errors, and will instead return 500 if the container isn't created. Also, the API that deletes containers needs a container ID, and a container that failed to be created doesn't have one.

bnjbvr:
Sorry I am not confident reviewing the JSON-patch and frontend changes, deferring to Jan.

lib/machines.js
```diff
@@ -205,7 +205,9 @@ exports.spawn = function (user, projectId, callback) {
   docker.runContainer({ host, image, ports }, (error, container, logs) => {
     if (error) {
       log('spawn', image, error);
-      machine.status = 'start-failed';
+      // Recycle the machine if it failed to spawn, error feedback is handled by the API 
+      machine.status = 'new';
+      machine.docker.container = '';
```
jankeromnes: You don't need to recycle the machine here, because its `status` is already `'new'` (guaranteed by the above call to `getOrCreateNewMachine()`) and its `machine.docker.container` should already be empty.

Instead, I think you can simply remove the `machine.status = 'start-failed'` (as you did) and also the `db.save()`.

static/js/janitor.js
```diff
@@ -178,12 +178,13 @@ function updateFormStatus (form, status, message) {
...
       break;
     default:
-      Array.forEach(form.elements, function (element) {
-        element.classList.remove('disabled');
-      });
       break;
   }
 
+  Array.forEach(form.elements, function (element) {
+    element.classList.remove('disabled');
```
jankeromnes: Why are you changing the logic here? The previous code purposefully didn't remove the `disabled` class on buttons if the status was a `success` or an `error` (forcing you to modify any field's value before being able to submit again, because submitting the exact same input doesn't make sense if you just had an error or a success).

jankeromnes:
Addressed my own nits.

PR #149, opened 2017-08-17
Removed margin from alert divs

arshdkhn1:
image
Margin creates ^ gap between the jumbotron and alert div, which doesn't look right.

After removing margin:
image

### DIFF
--- a/static/css/janitor.css
+++ b/static/css/janitor.css
@@ -576,6 +576,7 @@ section.video {
   margin: 0;
 }
 
+/* Override bootstrap alert margin */
 .alert {
-  margin: 0;
+  margin-bottom: 0;
 }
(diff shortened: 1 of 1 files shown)


static/css/janitor.css
```diff
@@ -575,3 +575,7 @@ section.video {
   text-align: center;
   margin: 0;
 }
+
+.alert {
+  margin: 0;
```
nt1m: Can you be more specific about the override ? margin-bottom: 0;

static/css/janitor.css
```diff
@@ -575,3 +575,7 @@ section.video {
   text-align: center;
   margin: 0;
 }
+
+.alert {
```
nt1m: Can you please add a comment: /* Override bootstrap alert margin */

arshdkhn1: sure.

static/css/janitor.css
arshdkhn1: oops, my bad.

arshdkhn1:
Changes made as per review.

PR #150, opened 2017-08-25
Stop using jQuery plugin for timestamps

nt1m:


bnjbvr:
Sorry I'll be on PTO for two weeks and won't be able to get this review quickly.

lib/routes.js
```diff
@@ -80,7 +80,7 @@ exports.landingPage = function (response, user = null) {
     user,
     scripts: [
       '/js/landing.js',
-      '/js/jquery.timeago-1.5.4.min.js',
+      '/js/timeago.min.js',
```
jankeromnes: Nit: Please suffix third-party minified scripts with their versions where applicable.

static/js/projects.js
```diff
@@ -42,13 +42,17 @@ Array.map(document.querySelectorAll('*[data-status]'), function (element) {
...
-  element.title = date.toUTCString().replace('GMT', 'UTC');
-  element.setAttribute('datetime', date.toISOString());
-
-  // Use jQuery's live-updating timeago plugin.
-  $(element).timeago();
+  try {
+    // GMT is deprecated (see https://en.wikipedia.org/wiki/UTC).
+    element.title = date.toUTCString().replace('GMT', 'UTC');
+    element.setAttribute('datetime', date.toISOString());
+  } catch (error) {}
```
jankeromnes: Nit: Why do you need a try/catch here? Also, please never ever use `catch (error) {}` (instead, we should handle errors, e.g. log or alert them (or if we really want to ignore a specific failure, mention this in a comment, and only ignore specific error types or messages).

static/js/projects.js
```diff
@@ -42,13 +42,17 @@ Array.map(document.querySelectorAll('*[data-status]'), function (element) {
 
 // Add fuzzy timestamps to elements with a 'data-timestamp' attribute.
 
-Array.map(document.querySelectorAll('*[data-timestamp]'), function (element) {
+var timestampElements = document.querySelectorAll('[data-timestamp]');
+Array.map(timestampElements, function (element) {
```
jankeromnes: Nit: `Array.forEach` is probably more relevant here.

static/js/projects.js
```diff
@@ -42,13 +42,17 @@ Array.map(document.querySelectorAll('*[data-status]'), function (element) {
 
 // Add fuzzy timestamps to elements with a 'data-timestamp' attribute.
```
jankeromnes: Nit: Please drive-by remove the empty line between function comment and definition (which is an older coding style).

static/js/projects.js
```diff
@@ -41,14 +41,14 @@ Array.map(document.querySelectorAll('*[data-status]'), function (element) {
...
   // GMT is deprecated (see https://en.wikipedia.org/wiki/UTC).
   element.title = date.toUTCString().replace('GMT', 'UTC');
   element.setAttribute('datetime', date.toISOString());
-
-  // Use jQuery's live-updating timeago plugin.
-  $(element).timeago();
 });
+
+// Use live-updating timeago plugin.
+timeago().render(timestampElements);
```
jankeromnes: Nit: Why move `timeago().render(element)` out of the `forEach` loop? I don't think you need to do that, nor to create a variable for the collection.

PR #151, opened 2017-09-01
Adding port to sign-in link if hostname is localhost.

arshdkhn1:
feel free to close it if changes doesn't make sense.

### DIFF
--- a/lib/users.js
+++ b/lib/users.js
@@ -12,7 +12,8 @@ const sessions = require('./sessions');
 
 const hostname = db.get('hostname', 'localhost');
 const security = db.get('security');
-const baseUrl = (security.forceHttp ? 'http' : 'https') + '://' + hostname;
+const baseUrl = (security.forceHttp ? 'http' : 'https') + '://' + hostname +
+  (hostname === 'localhost' ? ':' + db.get('ports').https : '');
 
 // Get a user for the current session.
 exports.get = function (request, callback) {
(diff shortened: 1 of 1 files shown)


jankeromnes:
Hm, thanks for implementing this suggestion. I see why it would seem to make more sense, but this would cause production Janitor sign-in links to look like https://janitor.technology:1443/?key=1234 instead of the current https://janitor.technology/?key=1234 By default, the HTTPS server runs on port 1443, and the HTTP redirector on port 1080, and these unprivileged ports are expected to be redirected to the default HTTPS/HTTP ports 443/80 (e.g. by calling `sudo make ports`, or by using a reverse proxy).

Hence, for sign in link I would prefer not specifying various ports. I agree that this makes the logged sign-in key look a bit weird, but maybe we can just fix the URL logging hack instead of the official sign in links? (I got used to seeing a sign-in link like http://localhost/?key=1234 and simply copy/pasting the key from it into whatever URL the Janitor instance is running on).

jankeromnes:
Ah, you made this change for localhost only... Hm. Don't close this yet, I'll review in the coming days.

lib/users.js
```diff
@@ -11,7 +11,9 @@ const sessions = require('./sessions');
 
 const hostname = db.get('hostname', 'localhost');
 const security = db.get('security');
-const baseUrl = (security.forceHttp ? 'http' : 'https') + '://' + hostname;
+const ports = db.get('ports');
+let baseUrl = (security.forceHttp ? 'http' : 'https') + '://' + hostname;
+baseUrl = hostname === 'localhost' ? (baseUrl + ':' + ports.https) : baseUrl;
```
jankeromnes: Nit: No need to make `baseUrl` mutable. Please construct it in a single expression (which will span over two lines).

lib/users.js
```diff
@@ -124,20 +126,20 @@ exports.sendLoginEmail = function (email, request, callback) {
         }
 
         return '<p>Hello,</p>\n' +
-        '<p>To sign in to the Janitor, please click ' +
+          '<p>To sign in to the Janitor, please click ' +
```
jankeromnes: Nit: I'm ok with indenting this line by two spaces, because it's a continuation of the `return` expression above. However, some lines of the email text were indented yet another time, in order to visually separate paragraphs (and make reading the markup string easier). Please indent as you did, but keep the visual separation by indenting the other lines twice.

lib/users.js
```diff
@@ -161,35 +163,35 @@ exports.sendInviteEmail = function (email, callback) {
       htmlMessage (key) {
         const url = baseUrl + '/settings/?key=' + encodeURIComponent(key);
         return '<p>You are invited to join the Alpha of the Janitor.</p>\n' +
-        '<p>To activate your free and unlimited account, please follow the ' +
+          '<p>To activate your free and unlimited account, please follow the ' +
```
jankeromnes: Nit: This text has changed, please rebase your commit.

jankeromnes:
@arshdkhn1 I can't merge this pull request nor push to your repository in order to properly rebase this change. Could you please add me as a collaborator on your repository?

arshdkhn1:
@jankeromnes sure, sent invitation!

jankeromnes:
Perfect, thanks! I was able to push successfully. Let's merge this, once the tests are green again. 😄

lib/users.js
arshdkhn1: I ran `npm run lint-fix` before pushing the commit so the indentation changes are done by that. Isn't there any eslint rule for the kind of formatting you are asking about?

jankeromnes: Strange, `npm run lint-fix` had never done this before, and I'm not able to reproduce this. Are you sure this change wasn't made by some other tool?

Since this formatting is very strange and custom, it likely won't have a corresponding eslint rule. I think it's ok if we simply ignore this case. Hopefully eslint won't attempt to correct this again, nor complain about it, but if it does I guess we'll want to tell eslint to ingore this particular formatting.

arshdkhn1: let me check that again and get back to you.

PR #154, opened 2017-09-07
Small fixes.

jankeromnes:


jankeromnes:
@nt1m Please have a look when you have the time.

- I switched NPM dependencies to exact versions, so that Greenkeeper opens a new pull request for each package update, and so that we can more easily catch regressions.
- I reverted to dockerode 2.4.3 (it was already fixed locally on the server, because 2.5.0 breaks our `execInContainer` method)
- I simplified the Alpha invite email, because we no longer need every user to create a Cloud9 account, nor do they have to copy/paste SSH keys and usernames around anymore. Yay!

PR #155, opened 2017-09-07
Add port to localhost sign-in links.

jankeromnes:
I rebased #151 and addressed my own nits.


jankeromnes:
Updated #151 instead.

PR #160, opened 2017-09-13
Burn bootstrap!

arshdkhn1:


static/css/janitor.new.css
```diff
@@ -0,0 +1,811 @@
+/***** common css starts *****/
+:root {
+  --white: #fff;
```
jankeromnes: Nit: Is it really necessary to create a CSS variable for the color `white`?

static/css/janitor.new.css
```diff
@@ -0,0 +1,811 @@
...
+  --orange-5: #ff922b;
+  --orange-5-rgb: 255, 146, 43;
+  --orange-6: #fd7e14;
+  --orange-6-rgb: 253, 126, 20;
+  --orange-7: #f76707;
+  --orange-7-rgb: 247, 103, 7;
+  --orange-8: #e8590c;
+  --orange-8-rgb: 232, 89, 12;
+  --orange-9: #d9480f;
+  --orange-9-rgb: 217, 72, 15;
```
jankeromnes: Nit: Is it really necessary to implement 9*2*13=234 color variations here? This feels a bit extreme, and I think we could save transfer bytes on this, at least in production.

static/css/janitor.new.css
```diff
@@ -0,0 +1,811 @@
+/***** common css starts *****/
+:root {
+  --white: #fff;
+  --white-rgb: 255, 255, 255;
```
jankeromnes: What is the point of `*-rgb` variable names?

static/css/janitor.new.css
```diff
@@ -0,0 +1,811 @@
...
+  --black: #000;
```
jankeromnes: Nit: Same as for `white`, is this really necessary?

templates/footer.html
```diff
@@ -20,11 +20,12 @@
     <script defer src="/js/bootstrap-3.3.7.min.js"></script>
     <script defer src="/js/scout.min.js"></script>
     <script defer src="/js/janitor.js"></script>{{ for (let src of scripts) { }}
-    <script defer src={{= src in json}}></script>{{ } }}
+    <script defer src={{=s rc in json}}></script>{{ } }}
```
jankeromnes: Nit: This is probably a typo.

templates/footer.html
```diff
@@ -20,11 +20,12 @@
...
     <script async defer>
-      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)})(window,document,'script','//www.google-analytics.com/analytics.js','ga');
+      (function (i, s, o, g, r, a, m) { i['GoogleAnalyticsObject'] = r; i[r] = i[r] || function () { (i[r].q = i[r].q || []).push(arguments) }, i[r].l = 1 * new Date(); a = s.createElement(o), m = s.getElementsByTagName(o)[0]; a.async = 1; a.src = g; m.parentNode.insertBefore(a, m) })(window, document, 'script', '//www.google-analytics.com/analytics.js', 'ga');
```
jankeromnes: Oh, did `npm run lint-fix` misbehave again? I don't think this change is useful, please restore the more compact version.

templates/footer.html
```diff
@@ -20,11 +20,12 @@
...
       ga('create', 'UA-62373597-1', 'auto');
       ga('send', 'pageview');
     </script>
   </body>
-</html>
+
+</html>
```
jankeromnes: Looks like you swapped the last two lines. Please swap them back (no empty line before `<html>`, but one empty line at the end of the file, to make Git and GitHub happy).

templates/header.html
```diff
@@ -1,36 +1,33 @@
 <!doctype html>
 <html lang="en">
+
   <head>
     <meta charset="utf-8">
-    <meta name="application-name" content="The Janitor">
-    <meta name="theme-color" content="#333">
+    <meta name="application-name" content="Janitor">
+    <meta name="theme-color" content="#000">
```
jankeromnes: You're the color-master here, but I'd be interested in knowing why you're reverting the "black" text to pure black. I read in various places online that a "softer black" is perceived as nicer and less brutal by users. Then again I don't know much about these things, and maybe this was the case a few design trends ago.

jankeromnes:
Ha, Tim was already flagged. I'll unflag Ben as he's very busy and I think we've got this.

.gitignore
```diff
@@ -14,3 +14,4 @@ backups
 tokens
 tests/db.json
 tests/templates
+.vscode
```
jankeromnes: Interesting contribution, thanks! Please move it just below `.DS_Store` (this file is ordered alphabetically, and grouped by non-janitor / janitor specific).

templates/footer.html
arshdkhn1: I have corrected it but I haven't pushed the commit yet.

templates/header.html
arshdkhn1: #333 was light in comparison to the black color that I have used as primary color. I can change `#000` to `#111` or `#222` but `#333` is much lighter.

templates/footer.html
arshdkhn1: empty line at the end of each file? @jankeromnes

templates/footer.html
arshdkhn1: Will do!

static/css/janitor.new.css
arshdkhn1: I copied a css boilerplate for colors which had rgb values as well. I will remove the rgb ones.

static/css/janitor.new.css
arshdkhn1: @jankeromnes once we are done with all pages, I will surely remove all the colors that we aren't using.

.gitignore
arshdkhn1: oops, I probably didnt take a look at this file carefully otherwise i would have understood the order.

static/css/janitor.new.css
arshdkhn1: Not necessary. I prefer using same terminology through out the css file for using colors.

static/css/janitor.new.css
arshdkhn1: Not necessary. I prefer using same terminology through out the css file for using colors. In this case, white variable is exactly same as `white` or `#fff` color so it might not make sense here but generally `white variable` can be any shade of white e.g., `#fafafa`(my fav), `#fefefe` etc

templates/footer.html
jankeromnes: In general, yes. 😄 But we haven't been very consistent with this rule.

templates/header.html
jankeromnes: Agreed, I believe `#333` was the lightest I could get away with. Please see if `#222` or `#111` can work well with your design, otherwise `#000` is fine too I guess. https://blog.nightly.mozilla.org/ uses `#000` for its titles and I love Mozilla's new design.

.gitignore
jankeromnes: The order of this file is not very important, it's just that I'm very nit-picky about everything 😄 and whenever things can be ordered alphabetically, I become very annoying about it.

arshdkhn1: Me too.

static/js/janitor.js
```diff
@@ -293,3 +293,10 @@ if ('serviceWorker' in navigator) {
       console.error(error);
     });
 }
+
+function toggleMenu() {
+  document.querySelector('.menu-toggle').classList.toggle('on')
+  document.querySelector('.navbar-menu').classList.toggle('open');
+}
```
nt1m: @jankeromnes wanted a bottom nav for mobile navigation.

Also, you can do this without any javascript using a checkbox hack if you're interested.

```css
input:checked + .navbar-menu {
  display: block;
}
```

Then you can hide the checkbox and make a pretty label for it.

static/css/janitor.new.css
```diff
@@ -0,0 +1,811 @@
...
+  --tracking-medium: 0.5rem;
+  --tracking-large: 0.8rem;
+  --shadow-small: 0 -4px 15px 0 var(--gray-5);
+  --shadow-large: 0 16px 32px 0 rgba(0, 0, 0, 0.1);
+  --dark: #000;
+  --slate: #0f1113;
+  --light-gray: #fafafa;
+  --bg-color: #fff;
+  --fg-color: var(--dark);
+  --fg-color-light: var(--gray-6);
```
nt1m: Can we avoid CSS variables altogether ? They're overly used here IMO, and not really awesome for compatibility (would be nice to be able to run Janitor on at least the latest IE). Plus, they make the CSS hard to read in this case: you have to make back and forths between the CSS and the variable definitions.

static/css/janitor.new.css
```diff
@@ -0,0 +1,811 @@
...
+  color: var(--fg-color);
+  font-weight: 500;
+}
+
+p a {
+  --color: var(--fg-color);
+  --border: var(--border-light);
+  color: var(--color);
+  padding-bottom: 3px;
+  border-bottom: 1px solid var(--border);
```
nt1m: Use text-decoration instead of border-bottom for an underline.

static/css/janitor.new.css
```diff
@@ -0,0 +1,811 @@
...
+}
+
+.navbar-menu a:before {
+  content: "";
+  position: absolute;
+  width: 100%;
+  height: 1px;
+  bottom: -5px;
+  left: 0;
+  background-color: var(--color);
```
nt1m: Use `background-color: currentColor;`

static/css/janitor.new.css
arshdkhn1: When text-decoration is used for underline then it appears just below the text. Using border, I can create underline below 3px.  I also prefer border because the underline width can be changed and borders can be animated. 
image

^ `blogs` uses `text-decoration` while `my-proposal` uses `border` for underline.

static/js/janitor.js
arshdkhn1: I dont think that it will be possible to make a sliding effect without using js. This code is changed in next commit. Please check that code because I dont think I am using best way to do the animation.

static/css/janitor.new.css
arshdkhn1: I normally try to stay consistent using same terminology throughout the stylesheet for colors. I have used css var for getting color all over the stylesheet.

static/css/janitor.new.css
arshdkhn1: In this case I wanted underline to appear below 3px because underline just below text looks ugly to me.

static/css/janitor.new.css
arshdkhn1: Yes, we can avoid it. I asked it on IRC but we didn't land to any definite result. I also think support to IE is necessary. I also agree, I have used css variables roughly here. Will make changes in future commits.

static/js/janitor.js
arshdkhn1: @nt1m https://media.giphy.com/media/3ov9jJ0IEssC11xuxy/giphy.gif 
this is the animation ^

arshdkhn1:
I forgot to add an empty line in janitor.js. Please look at code for `navbar-menu` animation and let me know any way to improve it.

static/js/janitor.js
arshdkhn1: @nt1m There are links in footer for navigation. I am not sure what kind of bottom navigation @jankeromnes wanted.

nt1m: I mentioned display: block/none, but you can of course make an animation with the checkbox hack.

templates/header.html
nt1m: The theme-color is supposed to match whatever is the header color (in this case white).

nt1m:
@arshdkhn1 Pure CSS toggle menu: https://jsfiddle.net/ntim/g0y6wmr2/ (No JS involved!)

arshdkhn1:
@ntim cool! I guess I didn't wrote about the static height part in last few comments. In your example you know that the height for each span will be `20px` after opening opening the menu. Since we are using system fonts who `character height` varies iirc. For ex Ubuntu font is wider and shorter in comparison to Roboto. Similar is the case with other fonts like Seoge UI, Helvatica etc. 

If you check out the latest code you will see I am using js to get `scrollHeight` and then assigning height to `navbar-menu` using js. Earlier version of menubar had bug which I also pointed out in the summary comment. Sorry for not clearing it earlier.

nt1m:
@arshdkhn1 I've updated the fiddle to not use the height property: https://jsfiddle.net/ntim/g0y6wmr2/

FYI, you can also use height: 1.1em if you don't know the character height.

arshdkhn1:
@nt1m I will try it out.

PR #161, opened 2017-09-13
Add a feature to list changed files in a container (docker diff).

jankeromnes:
The new documentation entry looks like this:

docker-diff

Here is what the feature looks like when you inspect filesystem changes in a given container from the Docker administration interface:

docker-diff-files

EDIT: All paths are now sorted alphabetically.

### DIFF
--- a/api/hosts-api.js
+++ b/api/hosts-api.js
@@ -602,8 +602,9 @@ containerAPI.get('/changes', {
     }
 
     const { hostname } = request.query;
+    // FIXME: For admins, find the machine regardless of which user owns it.
     const machine = machines.getMachineByContainer(user, hostname, container);
-    if (!machine) {
+    if (!machine && !users.isAdmin(user)) {
       response.statusCode = 404; // Not Found
       response.json({ error: 'Container not found' }, null, 2);
       return;
@@ -627,7 +628,7 @@ containerAPI.get('/changes', {
       urlParameters: {
         hostname: 'example.com',
         container: 'abcdef0123456789',
-      },
+      }
     },
     response: {
       body: JSON.stringify([
--- a/api/hosts-api.js
+++ b/api/hosts-api.js
@@ -431,9 +431,8 @@ containersAPI.put({
         return;
       }
 
-      response.json({
-        container: machine.docker.container
-      }, null, 2);
+      response.statusCode = 201; // Created
+      response.json({ container: machine.docker.container }, null, 2);
     });
   },
 
@@ -543,21 +542,22 @@ containerAPI.patch({
 containerAPI.delete({
   title: 'Delete a container',
 
-  handler ({ user, query }, response) {
-    const { container } = query;
+  handler: (request, response) => {
+    const { user } = request;
     if (!user) {
       response.statusCode = 403; // Forbidden
       response.json({ error: 'Unauthorized' }, null, 2);
       return;
     }
 
+    const { container } = request.query;
     if (container.length < 16 || !/^[0-9a-f]+$/.test(container)) {
       response.statusCode = 400; // Bad Request
       response.json({ error: 'Invalid container ID' }, null, 2);
       return;
     }
 
-    const { hostname } = query;
+    const { hostname } = request.query;
     const machine = machines.getMachineByContainer(user, hostname, container);
     if (!machine) {
       response.statusCode = 404; // Not Found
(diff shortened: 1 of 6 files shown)


jankeromnes:
@arshdkhn1 If you have time, could you please have a look at this change, and see if it makes sense to you? Please ask about anything that looks weird, or not right, or if you don't understand something.

If you don't have time, please feel free to cancel the review request, and I'll ask someone else to review it. 😄 Thanks!

static/js/admin.js
```diff
@@ -192,3 +192,126 @@ function formatContainerTree (image, linePrefix, lastChild, parentSize) {
...
+  if (node.Hidden && Object.keys(node.Children).length > 0) {
+    // If this node is hidden, append its children to a collapsed <div> instead.
+    parentElement = document.createElement('div');
+    parentElement.classList.add('collapse');
+    // Add a link to reveal hidden child nodes.
+    var a = document.createElement('a');
+    a.href = 'javascript:void(0)';
+    a.textContent = Object.keys(node.Children).length + ' hidden';
+    a.addEventListener('click', function (event) {
+      window.$(parentElement).collapse('toggle');
```
arshdkhn1: why jquery?

lib/docker.js
```diff
@@ -278,6 +278,23 @@ exports.execInContainer = function (parameters, callback) {
...
+  getDocker(host, (error, docker) => {
+    if (error) {
+      callback(error);
+      return;
+    }
+
+    const container = docker.getContainer(containerId);
+    container.changes((error, changedFiles) => {
+      callback(error, changedFiles);
+    });
```
nt1m: This is equivalent to: `container.changes(callback);`

arshdkhn1: good catch

api/hosts-api.js
```diff
@@ -580,6 +580,65 @@ containerAPI.delete({
...
+      response.json(changedFiles, null, 2);
+    });
+  },
+
+  examples: [{
+    request: {
+      urlParameters: {
+        hostname: 'example.com',
+        container: 'abcdef0123456789',
+      },
```
arshdkhn1: extra `,`

static/js/admin.js
```diff
@@ -192,3 +192,126 @@ function formatContainerTree (image, linePrefix, lastChild, parentSize) {
...
+  // By default, append child nodes to the same root element.
+  var parentElement = element;
+  if (node.Hidden && Object.keys(node.Children).length > 0) {
+    // If this node is hidden, append its children to a collapsed <div> instead.
+    parentElement = document.createElement('div');
+    parentElement.classList.add('collapse');
+    // Add a link to reveal hidden child nodes.
+    var a = document.createElement('a');
+    a.href = 'javascript:void(0)';
+    a.textContent = Object.keys(node.Children).length + ' hidden';
```
jankeromnes: `Object.keys(node.Children).length` is incorrect in this case, because Children can have Children themselves, and these are not counted here. (So the link may say "hidden 2", and actually reveal 15 elements, e.g.).

static/js/admin.js
jankeromnes: It was the quickest way for me to make a container that collapses on demand. Would you prefer I use something else? If so, please point me to a self-contained implementation of such a collapsing behavior.

arshdkhn1: Well you can do this for doing it with vanilla js - https://codepen.io/arshdkhn1/pen/oGXoRb?editors=0011

jankeromnes: Thanks for your suggestion! However, it doesn't animate.

I tried re-implementing an animated collapse in HTML/CSS/JS, but `transition: height` can't be used when one of the heights is `auto` (it need to be fixed, and the Bootstrap/jQuery collapse plugin does that by remembering the element's `offsetHeight`). Additionally, Bootstrap properly sets the `aria-expanded` attributes, which is great for accessibility.

Overall, I don't think I can come up with an on-par re-implementation of the Bootstrap/jQuery collapse, so I'll leave it as is for now. Sorry to add one more small dependency to Bootstrap/jQuery! But let's revisit a "vanilla collapse" in your bootstrap-removing pull request, when we're ready to port the admin section.

Thanks again for making Janitor better! Your help is very much appreciated 👍 😄

lib/docker.js
jankeromnes: Indeed! But like for other callbacks, we repeat the callback arguments here in order to make it more obvious/transparent what is actually returned. It's `lib/docker.js`'s job to know what the third-party Dockerode API returns, and users of `lib/docker.js` in this project should not have to read up on the Dockerode external library.

Simply repeating the argument here makes it very obvious what this callback returns, making it a lot easier to use and debug this function.

PR #168, opened 2017-10-01
Add Servo SVG logo

Coder206:
My SVG logo for the Servo project has been [approved](https://github.com/servo/servo/pull/18695). Here is the new image for the Servo project we can use for the project logo. 

@jankeromnes I recall the admin of the project sets the image icon, is that true for the Servo project as well?


jankeromnes:
Exciting! Thanks a lot for making this finally happen! 😄 🎆 💯 

> @jankeromnes I recall the admin of the project sets the image icon, is that true for the Servo project as well?

Correct! Well, technically we don't have "project-specific admins" yet, only site-wide admins that have access to https://janitor.technology/admin/projects/ but I'll merge this PR and update the logo used for Servo. 🍰 

Thanks again!

jankeromnes:
New logo successfully deployed to https://janitor.technology ! 🎉 🎉 🎉

Coder206:
@jankeromnes My pleasure! 

"Small question: Did you use svgomg to optimize this logo? I think it can make SVGs a lot smaller, especially with options like "multi-pass" (but it may take a few minutes)."

Yes, I ran out through that software with multi-pass. However, lowering the precision below 5 would alter the image of doge and split it into pieces. (You would have spaces seperating features of the animal)

Issue #59, opened 2017-04-18
Improve the Containers view and tools ("Containers v2.0")


jankeromnes:
We would like the Janitor web app to offer developers more control over their containers, so that they can do more with their containers, and troubleshoot common problems more easily.

Additionally, new container features should also be available in the upcoming Janitor API.

To improve the view and tools of Janitor containers, we should implement the following changes to Janitor's web app and server:

- [ ] Rename "Contributions" to "Containers" on the website (also, https://janitor.technology/contributions/ should redirect to https://janitor.technology/containers/)
- [ ] Make the existing "Firefox #0" container name/label editable (so that you can give it a name that reminds you what you use this particular container for, e.g. a feature you're implementing, a bug you're fixing, some tests you're running, etc)
- [ ] The (now empty) "Logs" text area should be replaced with a button (or "toggle-able text area") that allows you to start streaming live Docker container logs
- [ ] Add templated instructions describing how you can access your container via SSH (i.e. an `ssh` one-liner with the right host, port, and a link to the Settings section where you can authorize personal SSH public keys)
- [ ] Add buttons/links to access exposed container ports different from SSH (22), noVNC (8088) and Cloud9 (8089), for example Janitor containers should have a link to access their port 8080 (to preview the running Janitor web app through the Janitor proxy with a URL like https://moz1.janitor.technology/abcdef0123456789/8080/)
- [ ] Add a simple "diff view" to quickly see what was modified in this container (for example, the list of added/changed/removed files, or maybe a full diff if that's reasonable)
- [ ] Display the container ID somewhere
- [ ] Display the container's disk size footprint (actual size of the container, and also the total size of older Docker images you're preventing from being garbage collected)
- [ ] Display the actual Docker container status (running, stopped, not found) instead of what Janitor thinks the container status should be
- [ ] Add buttons to "Stop" and "Start" containers

If you have ideas, suggestions or feedback related to the items above, or if you want to help, please comment below. Thanks! 😄

jankeromnes:
We should also show which host the container is on (e.g. a very small, light grey "moz1.janitor.technology" somewhere would be amazing). Added to the list.

jankeromnes:
I'm currently implementing the SSH one-liner for containers, but I'd like your input on the UI.

## Option 1: An in-line text field

![ssh-oneliner2](https://user-images.githubusercontent.com/599268/31345285-8d29b182-ad15-11e7-8ef4-5fa69580cfb3.png)

[...]

PR #169, opened 2017-10-02
Rename Contributions to Containers

Coder206:
This PR addresses part of #59, renaming Contributions to Containers and also redirecting from contributions to containers. 

Upon scanning the repository with Cloud 9, I found that `/dockerfiles/ubuntu-dev/client-workspace-janitor.js` has a [reference](https://github.com/JanitorTechnology/dockerfiles/blob/master/ubuntu-dev/workspace-janitor.js#L27) to contributions but I was unsuccessful in adding it to the PR. Would it be better to commit this to the [dockerfiles repository](https://github.com/JanitorTechnology/dockerfiles/)?

### DIFF
--- a/app.js
+++ b/app.js
@@ -228,8 +228,13 @@ boot.executeInParallel([
     });
   });
 
-  // User contributions list.
+  // User contributions list. (legacy - redirect to containers page)
   app.route(/^\/contributions\/?$/, (data, match, end, query) => {
+    routes.redirect(query.res, '/containers/', true);
+  });
+
+  // User containers list.
+  app.route(/^\/containers\/?$/, (data, match, end, query) => {
     const { req: request, res: response } = query;
     const { user } = request;
     if (!user) {
@@ -237,7 +242,7 @@ boot.executeInParallel([
       return;
     }
 
-    routes.contributionsPage(response, user);
+    routes.containersPage(response, user);
   });
 
   // User notifications.
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -189,10 +189,10 @@ exports.loginPage = function (response) {
   ]);
 };
 
-// User contributions list.
-const contributionsTemplate = camp.template('./templates/contributions.html');
-exports.contributionsPage = function (response, user = null) {
-  const title = 'My Contributions';
+// User containers list.
+const containersTemplate = camp.template('./templates/containers.html');
+exports.containersPage = function (response, user = null) {
+  const title = 'My Containers';
   const projects = db.get('projects');
 
   response.template({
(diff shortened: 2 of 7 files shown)


app.js
```diff
@@ -237,8 +237,21 @@ boot.executeInParallel([
       return;
     }
 
-    routes.contributionsPage(response, user);
+    routes.containersPage(response, user);
   });
+  
+  // User container list.
```
nt1m: nit: containers

app.js
```diff
@@ -228,7 +228,7 @@ boot.executeInParallel([
     });
   });
 
-  // User contributions list.
+  // User contributions list. (legacy - redirect to containers page)
   app.route(/^\/contributions\/?$/, (data, match, end, query) => {
```
nt1m: I think it would be better the make the regex below (containing containers) to include contributions rather than duplicate the block.

Coder206: That makes sense would the resulting syntax be `/^\/containers|contributions\/?$/`?

Coder206:
@nt1m Thanks for the review and suggestions, in little time delay!

jankeromnes:
Thanks a lot @Coder206 for this pull request! And also @nt1m for the super quick review :D

> Upon scanning the repository with Cloud 9, I found that /dockerfiles/ubuntu-dev/client-workspace-janitor.js has a reference to contributions but I was unsuccessful in adding it to the PR. Would it be better to commit this to the dockerfiles repository?

Whoa, good catch! Indeed you found https://github.com/JanitorTechnology/dockerfiles/blob/448bef2ee08991e1937f3aad8875075fcfb14638/ubuntu-dev/workspace-janitor.js#L27 which needs to be modified in the dockerfiles repository directly.

However, maybe let's for the rename to be successfully merged and deployed on Janitor before updating it in dockerfiles (plus the Dashboard URL is a minor feature in Cloud9, and "contributions" will redirect to "containers" anyway, so we can take our time there).

app.js
jankeromnes: Interesting suggestion! It does simplify the code somewhat, however the behavior isn't the same:

- The original approach made a permanent HTTP redirection from /contributions/ to /containers/
- The new approach makes the server reply with the same content for both /contributions/ and /containers/

What I find strange in the second approach is that we don't indicate that the page URL was changed. I agree that it's a very minor difference from the user's perspective, however if you still use the old URL (e.g. via bookmarks or browsing history), you'll never see the new URL in your URL bar.

I think I'd prefer a permanent HTTP redirection (just like we do for API requests without a final `/`; the account page; favicon.ico and apple-touch-icons; and various others).

Coder206: Unless we can notify the user about the URL change I think the permanent HTTP redirection might be best for now.

jankeromnes:
Successfully deployed to https://janitor.technology ! Looks much cleaner now. 😄

PR #170, opened 2017-10-03
Fix #167: Focus input when clicking rename container

### LINKED ISSUE #167: Clicking on "rename a container" should focus on the text field
STR
===

- go to my contributions
- click "rename" icon next to a contribution's name

Observed
===

The contribution's name is replaced by a text field, but focus stays outside. This means that pressing Enter or clicking anywhere outside the field doesn't blur the input, and thus doesn't show the name back. Which looks a bit odd to me :)

Expected
===

Right after clicking, the browser focuses in the text field. It might even select all the text inside the text field, so it's easy and quick to replace it entirely.

nt1m:


static/js/janitor.js
```diff
@@ -249,6 +249,7 @@ Array.forEach(document.querySelectorAll('.editable-label'), function (label) {
   }
   toggle.addEventListener('click', function () {
     label.classList.add('editing');
+    label.querySelector('.editable-editor input').focus();
```
jankeromnes: Nit: This places the cursor at the beginning of the input field. I think it would be better to automatically place it at the end of the input, or even better to select the entire content of the input (to make it easier to replace, as suggested by @bnjbvr).

nt1m:
@jankeromnes Done!

static/js/janitor.js
```diff
@@ -249,6 +249,10 @@ Array.forEach(document.querySelectorAll('.editable-label'), function (label) {
   }
   toggle.addEventListener('click', function () {
     label.classList.add('editing');
+    
+    let editorInput = label.querySelector('.editable-editor input');
```
jankeromnes: Nit: Please use `var` instead of `let` (but actually, this variable probably isn't necessary, see below).

static/js/janitor.js
```diff
@@ -249,6 +249,10 @@ Array.forEach(document.querySelectorAll('.editable-label'), function (label) {
...
+    editorInput.focus();
+    editorInput.select();
```
jankeromnes: Nit: I believe that calling `.select()` will already focus the element, so no need to call `.focus()` above, and no need to create an intermediate variable.

nt1m:
review- → review?(@jankeromnes)

PR #174, opened 2017-10-17
Refactor RSA certificates helper to async/await.

jankeromnes:


jankeromnes:
@nt1m Could you please double-check that I didn't mess this up, when you have a minute? 😄

lib/certificates.js
```diff
@@ -229,143 +214,99 @@ exports.createTLSCertificate = function (parameters, callback) {
...
-  // Task: Use the given Let's Encrypt URL to discover its ACME protocol URLs.
-  client.getAcmeUrls(letsEncryptUrl, (error, urls) => {
-    if (error) {
-      callback(error);
-      return;
-    }
-    acmeUrls = urls;
-    done();
-  });
+  const tasks = [];
```
nt1m: A Promise.all() here seems like an odd choice here, why not something like this:

```js
const acmeUrls = await new Promise (...);
if (!accountKey) {
  accountKey = await exports.generateRSAKeyPair({privateKey} => forge.pki.privateKeyToPem(privateKey));
}
if (!httpsKey) {
  httpsKey = await exports.generateRSAKeyPair({privateKey} => forge.pki.privateKeyToPem(privateKey));
}
```

lib/boot.js
```diff
@@ -99,23 +99,20 @@ exports.ensureHttpsCertificates = function (next) {
...
-    letsencrypt.key = accountKey;
-    db.save();
-    log('[ok] new https credentials installed');
-    next();
-  };
-
   log('requesting new https credentials…');
-  certificates.createHTTPSCertificate(parameters, onCertificate);
+  certificates.createHTTPSCertificate(parameters)
+    .then(({ certificate, accountKey }) => {
```
nt1m: nit: I prefer this style:
```js
promise.then(() => {
...
}).catch(() => {
...
});
```
Rather than
```js
promise
  .then(() => {
    ...
  })
  .catch(() => {
    ...
  });
```

lib/boot.js
```diff
@@ -216,17 +210,14 @@ exports.ensureDockerTlsCertificates = function (next) {
     };
 
     log('generating new docker-tls client certificate…');
-    certificates.createTLSCertificate(parameters, (error, crt, key) => {
-      if (error) {
-        log('[fail] tls', error);
-        return;
-      }
-
+    certificates.createTLSCertificate(parameters).then(({ crt, key }) => {
```
nt1m: You could use async/await for this and below, but I don't have a preference in these particular cases.

jankeromnes:
> Sorry it took me so long to look at this.

No need to apologize! 😄 Thanks for eventually doing the review! I was eager to have your opinion on this async/await stuff as I know you like it and might have improvement suggestions.

lib/boot.js
jankeromnes: I also prefer the first style when possible, but the problem here is that it makes the first line grow longer than 80 characters, so I had to break somewhere.

I will refactor the entire `lib/boot.js` file to async/await anyway, but I just didn't want to refactor everything in the same pull request. Are you ok with this being addressed in a follow-up?

lib/boot.js
jankeromnes: Indeed, and I'm planning to do so in a follow-up 😄 (see earlier comment).

lib/boot.js
nt1m: Sounds good to address this in a followup.

lib/certificates.js
jankeromnes: The problem with this, if I understand async/await correctly, is that all the promises can be executed at the same time, but in your suggestion the generator would block at each promise and wait for it to finish.

I'm looking to make this function as fast as possible, by waiting for N asynchronous jobs to finish in parallel, rather than sequentially waiting for one to finish before starting the next one. Does that make sense?

nt1m: I'm not sure if running sequentially vs in parallel makes a big difference here, but if running in parallel is very important, I'm ok with it.

jankeromnes: I'd really prefer running things in parallel when possible, even though I don't know how many milliseconds this saves if any. You can call it irrational or premature optimization if you will. 😝

jankeromnes:
Nice! Now I'll get to use your Cloud9 plugin to work through your review comments.

Thanks! 😄 👍

PR #176, opened 2017-10-17
New containers page

nt1m:


jankeromnes:
Very nice! Thanks a lot for implementing this new Containers page, it looks great!

And it looks similar to @arshdkhn1's new design too. I'm really happy that we're converting Janitor page after page! 😄

Maybe it would be worth working on a common "new CSS" file that we could polish and land in a separate pull request, so that all future re-design PRs can use it? (And maybe even add new app header + footer that use the new CSS file, but maybe that's already too much for one pull request).

And in general, I'm OK with file names like `new-janitor.css`, and variable names like `newAppHeader`, but we should remain consistent throughout the code base.

new-containers-empty-chrome

new-containers-one-firefox

I have a few nits as well:
- [ ] The footer doesn't "stick" to the bottom of the page (maybe the section before the footer needs a `flex-grow: 1`?)
- [ ] Maybe we could add some helpful content when there are no containers, e.g. "You haven't created any containers yet. You can do that from the [Projects](https://janitor.technology) page."
- [ ] The bottom nav bar looks broken (`#janitor (IRC)` is on a separate line, and the links aren't vertically centered)

nt1m:
@jankeromnes @arshdkhn1 Can you please review this? Thanks :)

lib/routes.js
```diff
@@ -210,6 +218,27 @@ exports.containersPage = function (response, user = null) {
   ]);
 };
 
+// User containers list.
+const containersTemplateNew = camp.template('./templates/containers-new.html');
+exports.containersPageNew = function (response, user = null) {
+  const title = 'My Containers';
```
jankeromnes: Nit: How about just "Containers"?

static/js/projects.js
```diff
@@ -18,26 +18,22 @@ Array.map(document.querySelectorAll('a[data-action="spawn"]'), function (link) {
 // Add status badges to elements with a 'data-status' attribute.
 
 Array.map(document.querySelectorAll('*[data-status]'), function (element) {
-  var span = document.createElement('span');
   var status = element.dataset.status;
 
   // Format the status.
-  span.textContent = status.split('-').join(' ');
-  span.classList.add('text-capitalize');
+  element.title = status.split('-').join(' ');
```
jankeromnes: Nit: Maybe add the string `'Status: '` in front of the title, just to be extra explicit?

static/js/projects.js
```diff
@@ -18,26 +18,22 @@ Array.map(document.querySelectorAll('a[data-action="spawn"]'), function (link) {
...
   };
-  label['build-failed'] = 'danger';
-  label['start-failed'] = 'danger';
-  label['update-failed'] = 'warning';
-  span.classList.add('label', 'label-' + (label[status] || 'default'));
-
-  element.appendChild(span);
+  classMap['build-failed'] = 'error';
+  classMap['start-failed'] = 'error';
+  classMap['update-failed'] = 'warning';
```
jankeromnes: Actually `classMap` may be unnecessary, because we no longer use anything other than `started` (and `new`, but this gets filtered out in the template). So we may wish to simplify this code. Or we could prepare for when we add statuses for `stopped` (grey or black ?) and `unknown` (red? for when Docker host is down, or container isn't found, or other major problem)? As you wish.

templates/containers-new.html
```diff
@@ -0,0 +1,74 @@
...
+        <div class="panel-heading">
+          <span class="status-icon" data-status="{{= machine.status in xmlattr}}"></span>
+          <img class="panel-icon" src="{{= project.icon in xmlattr}}" alt="{{= project.name in xmlattr}} Logo">
+          <div class="grow">
+            <div class="editable-label project-title">
+              <form action="/api/hosts/{{= machine.docker.host in xmlattr}}/containers/{{= machine.docker.container in xmlattr}}" class="ajax-form editable-editor has-feedback" data-refresh-after-success="true" method="patch">
+                <input class="form-control" data-submit-on="blur" name="/name" placeholder="{{= project.name in xmlattr }} #{{= id in id}}" type="text" value="{{= machine.properties.name in xmlattr }}">
+              </form>
+              <div class="editable-value">
+                <h4>{{= name in html}}</h4>
```
jankeromnes: Nit: Please add a `id={{= machine.docker.container in json}}` attribute to this `h4` element (so that we can use URLs like `https://janitor.technology/containers-new/#abcdef0123456789`).

templates/containers-new.html
```diff
@@ -0,0 +1,74 @@
...
+            const name = machine.properties.name || `${project.name} #${id}`;
+      }}
+      <div class="panel panel-default panel-project">
+        <div class="panel-heading">
+          <span class="status-icon" data-status="{{= machine.status in xmlattr}}"></span>
+          <img class="panel-icon" src="{{= project.icon in xmlattr}}" alt="{{= project.name in xmlattr}} Logo">
+          <div class="grow">
+            <div class="editable-label project-title">
+              <form action="/api/hosts/{{= machine.docker.host in xmlattr}}/containers/{{= machine.docker.container in xmlattr}}" class="ajax-form editable-editor has-feedback" data-refresh-after-success="true" method="patch">
+                <input class="form-control" data-submit-on="blur" name="/name" placeholder="{{= project.name in xmlattr }} #{{= id in id}}" type="text" value="{{= machine.properties.name in xmlattr }}">
```
jankeromnes: Nit: Unrelated to your current pull request, but I think I previously noticed an off-by-one bug with the `placeholder` attribute. Could you please double-check it works properly?

templates/containers-new.html
```diff
@@ -0,0 +1,74 @@
...
+        </div>
+        <div class="panel-body tabs">
+          <ul class="tab-nav">
+            <li class="tab" data-tab="info"><a href="#">Info</a></li>
+            <li class="tab" data-tab="terminal"><a href="#">Terminal</a></li>
+            <li class="tab" data-tab="advanced"><a href="#">Advanced</a></li>
+          </ul>
+          <div class="tab-panels">
+            <div class="tab-panel" data-tab="info">
+              <p class="project-updated">Built <time class="project-timestamp" data-timestamp="{{= machine.data.updated in integer}}"></time>.</p>
```
jankeromnes: Nit: Maybe this is a good time to install `timeago` on the server, and pre-fill this tag before the client-side takes over? (And we could also default to `never` when there is no timestamp).

templates/containers-new.html
```diff
@@ -0,0 +1,74 @@
...
+          const project = projects[projectId];
+          for (const id in machines) {
+            const machine = machines[id];
+            if (machine.status === "new") {
+              continue;
+            }
+            const name = machine.properties.name || `${project.name} #${id}`;
+      }}
+      <div class="panel panel-default panel-project">
+        <div class="panel-heading">
```
jankeromnes: Nit: The panel looks really good, thanks a lot! However, I feel like the panel heading could use a little more padding. The status badge, project logo and buttons feel too "crammed" against the edge of the heading.

Also the buttons look very small, given their usefulness I'd prefer them to be bigger and a bit further apart, making them very easy to hit even on mobile.

new-containers-feedback

Also, please make it clearer what the "Danger zone" is (e.g. with a red outline, light red background, or something similar).

And finally, do you think that people will be worried that the `settings` icon is gone? If that's the case, it might be more reassuring to add a `settings` icon back that simply opens "Advanced" (but I don't have a strong opinion on this).

arshdkhn1:
@ntim I will add my review very soon. Sorry for being late!

PR #178, opened 2017-10-20
Allow users to have multiple email addresses.

jankeromnes:


lib/users.js
```diff
@@ -291,6 +291,12 @@ function getOrCreateUser (email) {
     db.save();
   }
 
+  // Temporary migration code: Previous users didn't have multilple emails.
+  if (!user.emails) {
```
bnjbvr: I'd do the check the other way around: `if (typeof user.email !== 'undefined')`

jankeromnes:
Thanks a lot for this review!

> I really don't like how the current code relies on emails[0] in many places. How about having a getter for this

Sure! I'll implement a hidden `primaryEmail` or `_primaryEmail` getter for this.

> Then in the getter, you could handle the non-migrated users (but maybe it doesn't make sense, if migration happens before any call to this getter).

No need, users are already being migrated in the `getOrCreateUser()` function, which is what's used to get any `user` in the server.

> I'd do the check the other way around: `if (typeof user.email !== 'undefined')`

I feel that it's safer/cleaner to check for what's really important, i.e. the existence of `user.emails` (because later code relies on it). This allows handling any corner cases like users that have neither `user.email` nor `user.emails` (this has never happened before, but I see these migration checks as "ensure user requirement X is satisfied", rather than "see if we find something in an older format").

lib/users.js
bnjbvr: OK, then you should probably use a more precise check here, like `typeof user.emails === 'undefined'`. I'm always nervous with implicit conversions.

jankeromnes: Ok thanks, then I will use `Array.isArray()` (because I want to overwrite `user.emails` if it's `undefined` or `null` or `0` or `{}` or whatever else isn't the expected format).

bnjbvr: Sounds good, cheers!

jankeromnes:
Thanks again! 💯

PR #179, opened 2017-10-23
New base janitor templates

nt1m:


jankeromnes:
@arshdkhn1 Could you please have a look at this new CSS file too, when you have a chance? Will it be easy for you to base your new design / new login page pull request on it? If not, what major differences or blockers do you see?

jankeromnes:
@nt1m Thanks a lot for splitting your new design resources into a separate pull request! It makes it much easier to review. 😄 

I can't wait to see it merged!

However, I don't really like that `janitor.js` gets mostly duplicated to `janitor-new.js`. What about keeping all the common parts in `janitor.js` only, and add any new incompatible code in `janitor-new.js`, and also move any deprecated code into a `janitor-old.js`. Does this sound fair to you?

Here is an example of what this could look like:
- `janitor-new.js` contains the new `$` and `$$` helpers, the tab setup, and the new `updateFormStatus` + `addFormFeedback`
- `janitor-old.js` contains the `[data-action]` handler and `ajaxForm` function, the old `updateFormStatus` + `addFormFeedback`, the deprecated `getFormData` helper and the modals setup
- `janitor.js` contains all the rest, which is common between these two files

(I basically did `diff -up static/js/janitor{,-new}.js` to find out what needed to be in each file, i.e. the `+` lines in `janitor-new.js`, the `-` in `janitor-old.js` and the unmodified context lines in `janitor.js`).

Sorry if this makes the split a little bit harder / more complex, but I think it's the best way to preserve Git history. Please let me know if you'd like some help on this split.

Additionally, no need to do similar splits for the headers or CSS files. However, I guess that the `-old/-new` split will remove the need for a `footer-new.html` (given that it will remain unchanged with 
importing `janitor.js`, and you might have to tweak the `scripts` array of the routes in `lib/routes.js` to add `janitor-old.js` there).

jankeromnes:
Note: If you really don't feel like splitting the JS file, you can leave it as is. It's not really that important to preserve the Git history of each line.

nt1m:
@jankeromnes The one major difference I see is that the new design isn't responsive yet. I'll try to work on it as soon as I can, but I can't promise it'll be part of this PR.

jankeromnes:
@nt1m Would it help if I provide you with a mockup of what the mobile version could look like? I believe we could add a few media breakpoints for your design to work great on mobile.

templates/header-new.html
```diff
@@ -0,0 +1,46 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="application-name" content="Janitor">
+    <meta name="theme-color" content="#333">
```
arshdkhn1: @nt1m you were right about the theme color. It should be same as header color.

static/css/janitor-new.css
```diff
@@ -0,0 +1,428 @@
...
+
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+  /* Use github's list of system fonts */
+  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
```
arshdkhn1: roboto ?

static/css/janitor-new.css
```diff
@@ -0,0 +1,428 @@
...
+  cursor: pointer;
+}
+
+/* Alerts */
+
+.alert {
+  padding: 10px;
+}
+
+.alert.alert-warning {
```
arshdkhn1: The background color doesn't look good IMO and box-shadow of header gets mix with the warning background. I was thinking of moving the alert warning to top of header, what do you say?

static/css/janitor-new.css
```diff
@@ -0,0 +1,428 @@
...
+
+.form-control.disabled,
+.form-control:disabled {
+  pointer-events: none;
+  background: #f5f5f5;
+}
+
+/* Navigation bar */
+
+.navbar {
```
arshdkhn1: should we have a stickly navbar?

templates/header-new.html
nt1m: Whoops, this was leftover from the old design.

static/css/janitor-new.css
nt1m: I like Roboto but I don't think macOS or Windows include Roboto by default. We would have to import a web font to make it work everywhere.

For now, I'd like to keep it simple by avoiding adding web fonts.

static/css/janitor-new.css
nt1m: How about adding some margin around the alert like: http://design.firefox.com/photon/components/message-bars.html

static/css/janitor-new.css
nt1m: I don't mind either way, I guess sticky could be nice.

static/css/janitor-new.css
jankeromnes: I think GitHub used to have `Roboto` in their font stack (without using web fonts) because it seems to be available on some systems (e.g. Android ICS 4+), but I see now that it has been removed from their stack.

The font stack @nt1m uses now is the current one in use on github.com . I'm not sure why they removed Roboto, but I'm also not against adding it back (without web fonts) if it provides a small advantage.

jankeromnes:
> Sorry for reviewing it so late.

No worries at all! Thanks a lot for volunteering your time to review it! 😄 I'm really happy that you were able to do it, because I wasn't the best reviewer to ask for a second opinion on @nt1m's excellent work.

> @jankeromnes these templates are similar to my templates.
> I am right now on Linux Mint and I feel like we can make font a bit more pretty. There is not much to review about the design of templates right now.

Indeed, thanks for suggesting we add Roboto back. If it can improve our app on some systems, at no cost, then I'm all for it. 👍 

> I think we should make a storybook(theme) kind of page(like this one) where all the different components(footer, headers, buttons etc) can be seen. It would be easier to collaborate if we make individual components first and then add them to Janitor. @nt1m what do you say?

That's a great idea! I was thinking we could land new components one by one, with each pull request introducing their own components, but I also like the idea of having a page that includes all the components so that we can work on the design independently from the features.

Maybe let's go ahead with these base templates, and open a follow-up pull request to collect all components we need and expose that page at a hidden URL like `/design/` or `/components/`?

static/css/janitor-new.css
arshdkhn1: I was using 
``` 
font-family: -apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Oxygen, Ubuntu, Cantarell,
    Fira Sans, Droid Sans, Helvetica Neue, sans-serif;
```
 In new design CSS and robotics seems to be available in many OSs.  I also now think that using web font is not necessary at all.

static/css/janitor-new.css
```diff
@@ -0,0 +1,428 @@
+/* General */
+
+* {
+  box-sizing: border-box;
+}
+
+body {
```
jankeromnes: Nit: When there isn't enough content (e.g. try the `/data/` page with the new design), the footer doesn't stick to the bottom, as you can see in this screenshot:

footer-problem

What about adding some rules like `html, body { height: 100%; }`, `body { display: flex; flex-direction: column; }` and `section:last-of-type { flex-grow: 1 }` to force the footer to the bottom of the page when the content isn't high enough?

static/css/janitor-new.css
jankeromnes: Great alerts! I'm really starting to like the Photon design system, so +1 on taking design queues from it.

jankeromnes:
@nt1m Looks great! Thanks again for splitting out the new design resources into a separate pull request. Please address our nits when you have some time, and then we'll be good to go. 😃

nt1m:
@arshdkhn1 @jankeromnes Solved all of the issues! Also, I decided to go to Fira Sans instead of Roboto, as it is the Photon design system font!

arshdkhn1:
I will review it in next few days. Sessional exams going on.

static/css/janitor-new.css
```diff
@@ -0,0 +1,438 @@
+/* General */
+
+* {
+  box-sizing: border-box;
+}
+
+html, body {
+  height: 100%;
```
arshdkhn1: change it to `min-height` because when the content height of body is greater than 100% then header size shrinks to height of 2em(padding) or you can remove `overflow: auto` from header.

static/css/janitor-new.css
```diff
@@ -0,0 +1,438 @@
...
+}
+
+.grow {
+  flex-grow: 1;
+}
+
+.container {
+  padding: 0 20vw;
+  overflow: auto;
+  flex-grow: 1;
```
arshdkhn1: Container does not have a flex parent. Maybe here you are trying to make sure that content height is 100%?

static/css/janitor-new.css
```diff
@@ -0,0 +1,438 @@
...
+.float-right {
+  float: right;
+}
+
+.gray-text {
+  color: #737373;
+}
+
+.grow {
+  flex-grow: 1;
```
arshdkhn1: image
^ you can see the height is not 100%. You can just add `align-items: flex-end; flex-grow: 1` to footer and that will do the work. There are other ways feel free to choose what you think is the best.

static/css/janitor-new.css
```diff
@@ -0,0 +1,438 @@
...
+  display: inline-block;
+  margin: 0;
+  padding: 0;
+}
+
+.navbar ul li a {
+  text-decoration: none;
+}
+.navbar ul li:not(:first-child) a {
+  padding-left: 2em;
```
arshdkhn1: for anchor tags I always give same padding to left and right because that makes sure the outline is sym wrt to inner text.

image
^ you can see the outline of not:first items have empty space at left and for first item it will be just outside the text. I like same size symetrical outlines.

static/css/janitor-new.css
```diff
@@ -0,0 +1,438 @@
...
+.editable-toggle {
+  margin-left: 10px;
+  cursor: pointer;
+}
+
+/* Alerts */
+
+.alert {
+  padding: 10px;
+  margin: 1em;
```
arshdkhn1: it still doesn't look good to me. sorry! We need to think a bit more about it.

static/css/janitor-new.css
arshdkhn1: image
What about adding a max-width?

static/css/janitor-new.css
arshdkhn1: The yellow color is too warm. Solid black and this yellow color combo doesnt work IMO.

static/css/janitor-new.css
nt1m: If you look at my contributions page PR, `.container` is a direct children of body, which has display: flex; on it.

static/css/janitor-new.css
nt1m: It doesn't work on this specific page, because it's not using the new `.container` class I've just introduced.

static/css/janitor-new.css
nt1m: Good point, I'll fix that.

static/css/janitor-new.css
nt1m: This is part of the photon style guide: https://bwinton.github.io/photon/components/message-bars.html

Anyway, I don't think the warning is super important, as it only appears in development mode.

nt1m:
>  What do you think about adding a wrapper(main tag) for main content ?

I currently use the .container class for that, but a main tag could also work.

static/css/janitor-new.css
arshdkhn1: @nt1m it also shows up when a particular project is running down ? I saw this alert a couple of times in production.

nt1m: Right. Maybe we can use the attention style (the lightblue one) instead of the warning style from the style guide, for "Janitor is down" alerts.

static/css/janitor-new.css
```diff
@@ -0,0 +1,438 @@
...
+
+html, body {
+  height: 100%;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+  /* Use github's list of system fonts */
+  font-family: "Fira Sans", Roboto, "Droid Sans", -apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
```
jankeromnes: Nit: It seems that the first three fonts are not from GitHub. Could you please edit your comment to something more precise, e.g.: `The list of system fonts after "Droid Sans" are from github.com`?

static/css/janitor-new.css
jankeromnes: I like sticky navbars. The only thing that concerns me about them is that they take away valuable screen "real estate" (on small screens, this is a problem if you need a lot of space for your content).

Maybe in our case it's more useful to a non-sticky navbar, in order to give our content more space, making the navigation links just slightly harder to reach? (But please make the navbar sticky if you don't agree with my conclusion, I don't have a strong opinion about this.)

static/js/janitor-new.js
```diff
@@ -0,0 +1,262 @@
...
+
+  if (form.dataset.redirectAfterSuccess && status === 'success') {
+    setTimeout(function () {
+      location.href = form.dataset.redirectAfterSuccess;
+    }, 400);
+  }
+}
+
+// Add visual feedback elements to a given <form>.
+function addFormFeedback (form) {
```
jankeromnes: General comment: I wish we had a better way to show visual feedback on asynchronous actions.

The problems with our current approach (which is simply moved over here):
- Feedback is per `<form>` and not per control (e.g. you could imagine a form with several buttons, as well as several input fields, that all trigger different async actions)
- There is no real "pending" state, see #128 (we don't know when an async action is in progress in the background, and controls are not disabled while an action is in progress. This confuses our users, and causes them to sometimes trigger actions like "spawn" and "delete" several times in parallel, leading to errors and more confusion)

templates/footer-new.html
```diff
@@ -0,0 +1,25 @@
...
+        <li><a href="https://github.com/janitortechnology/janitor/blob/master/LICENSE">&copy; Janitor</a></li>
+      </ul>
+      <ul>
+        <li><a href="/reference/api">API</a></li>
+        <li><a href="/blog/">Blog</a></li>
+        <li><a href="https://discourse.janitor.technology/">Community</a></li>
+        <li><a href="/data/">Data</a></li>
+        <li><a href="https://github.com/janitortechnology/janitor">GitHub</a></li>
+        <li><a href="https://github.com/janitortechnology/janitor/issues/new">Report Bug</a></li>
+        <li><a href="https://kiwiirc.com/client/irc.freenode.net/?#janitor">#janitor (IRC)</a></li>
```
jankeromnes: Random idea: We could link to `/nextclient/` instead of `/client/` here. :)

templates/header-new.html
```diff
@@ -0,0 +1,46 @@
...
+          </a>
+        </li>
+        <li>
+          <a href="/settings/">
+            Settings
+          </a>
+        </li>
+        <li>
+          <a href="/logout">
+            Sign out
```
jankeromnes: Nit: I feel like all this expanded markdown makes it harder to read than simply:

```html
<li><a href="/projects/">Projects</a></li>
<li><a href="/settings/">Settings</a></li>
<li><a href="/containers/">Containers</a></li>
<li><a href="/logout">Sign out</a></li>
```

static/css/janitor-new.css
jankeromnes: I confirm, when the content gets tall enough, both header and footer sections are vertically shrinked, and they even get tiny scrollbars in Chrome:

header-tiny-scrollbar

static/css/janitor-new.css
```diff
@@ -0,0 +1,438 @@
...
+  align-items: center;
+  text-decoration: none;
+  color: inherit;
+  font-weight: 600;
+}
+
+.navbar-brand img {
+  width: 2em;
+  height: 2em;
+  margin-right: 1em;
```
jankeromnes: Nit: I feel like this spacing is a bit too big (the logo and "Janitor" text are very far apart).

templates/header-new.html
```diff
@@ -0,0 +1,46 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="application-name" content="Janitor">
+    <meta name="theme-color" content="#ffffff">
```
jankeromnes: General note: I agree that sticking to default styles where possible is better (in order to avoid unnecessary complexity), but design guidelines like http://bettermotherfuckingwebsite.com seem to indicate that it's better not to use "pure black" text and "pure white" backgrounds, but colors that are just a little off to be less aggressive. (And it also says to use plenty of spacing.)

Photon does seem to use "pure white" backgrounds and texts, and I'm ok with that, but its dark text color is not "pure black" (instead it's "[Grey 90](http://design.firefox.com/photon/visuals/color.html#grey)" a.k.a. "#0c0c0d", or "[SomeOtherColor 90](http://design.firefox.com/photon/visuals/color.html#color-combinations)" on colored backgrounds where white text doesn't work).

static/js/janitor-new.js
```diff
@@ -0,0 +1,262 @@
...
+    return;
+  }
+  toggle.addEventListener('click', function () {
+    label.classList.add('editing');
+    label.querySelector('.editable-editor input').select();
+  });
+});
+
+// Remove the query string (e.g. '?key=123') from the URL.
+function removeQueryString () {
```
jankeromnes: Please also import my changes from https://github.com/JanitorTechnology/janitor/commit/5d6a4c933c3a9901f7639377777d03c9e4a47e4d 😃 

I hope it won't be too hard to keep the `*-new` files up-to-date during this transition period, and that we won't lose precious work when we switch to the new design. I guess we'll have to be careful when we overwrite the original files.

static/css/janitor-new.css
nt1m: I think it looks OK, we can adjust this later if needed

nt1m:
Latest screenshots:
Desktop:

screen shot 2017-11-14 at 15 52 35

Mobile:
screen shot 2017-11-14 at 15 53 00

static/js/janitor-new.js
nt1m: Added pending state :) More elaborate styling can be added later :)

jankeromnes: Amazing! Thanks 😄 I'm looking forward to try it.

static/css/janitor-new.css
jankeromnes: Fair enough :)

static/js/janitor-new.js
arshdkhn1: https://codepen.io/arshdkhn1/pen/ooGEYM

If you are looking for few simple loading animations

arshdkhn1: just a reminder: cn't login as ajaxForm method is missing

jankeromnes: > just a reminder: cn't login as ajaxForm method is missing

That's expected, because we're dropping this deprecated function in the new design (instead, async forms should use `fetchAPI`, either automatically by having a `method` attribute, or by implementing a custom submit handler).

jankeromnes: > If you are looking for few simple loading animations

Thanks a lot for providing example animations! However, they feel very basic. Ideally I'd like something quite fast and elegant (for example, MailChimp has very nice pending animations directly in its buttons).

static/css/janitor-new.css
jankeromnes: > I think header should be position absolute and its height change should not push down the body content. Feel free to ignore this nit if you have different opinion.

Thanks for this suggestion! I don't have a strong opinion on this either, but I generally dislike UI elements that covers the content (especially when it's badly done, e.g. there is a bug and I can't get to the content anymore). Here it wouldn't be badly done, but still, I have this bad feeling about covering content. So maybe it's simpler/safer to leave the mobile menu behavior as is. :)

static/css/janitor-new.css
```diff
@@ -0,0 +1,440 @@
...
+    display: block;
+    padding-left: 1em;
+  }
+
+  .navbar ul li {
+    display: block;
+    text-align: center;
+  }
+  .navbar ul li a {
+    display: block;
```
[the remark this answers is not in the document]
jankeromnes: In my checkout, `<li>` items don't have padding (only the `<header>` and the `<a>` inside the `<li>` have padding).

But +1 on not underlining when there is already a grey background (and a lighter grey would be nice too).

static/css/janitor-new.css
```diff
@@ -0,0 +1,440 @@
+/* General */
+
+* {
+  box-sizing: border-box;
+}
+
+html, body {
+  min-height: 100%;
```
jankeromnes: Nit: I'm not sure `min-height` actually works as intended here, because I just had an `html` that was actually 100% of the height, but `body` that was only as high as the content. Changing it to simply `height: 100%` fixed it. (However, I seem to remember that we switched to `min-height` because of another side-effect, which I don't remember). Oh well. Please use the one you prefer, and let's tweak it further in later pull requests. 🙂

templates/header-new.html
notriddle: According to the document you linked to:


> Firefox Grey is commonly used for interfaces (e.g., the Firefox for desktop light theme).


> We use white for the background of doorhangers or in-content menus. 


I notice that the top panel is Grey 10, while the overflow menu is pure white. Maybe Janitor's main background should be Grey 10?

static/css/janitor-new.css
nt1m: Yeah, min-height seemed to cause https://github.com/JanitorTechnology/janitor/pull/179#discussion_r148537377

Going to leave that is it is for now.

static/css/janitor-new.css
```diff
@@ -0,0 +1,440 @@
...
+
+.alert.alert-warning {
+  background: #ffe900;
+}
+
+/* Icons */
+
+.icon::before,
+.form-control-feedback::before {
+  content: "";
```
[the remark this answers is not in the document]
nt1m: It makes sense for the menu (I'll fix that), but not sure it makes sense for other usages, the form feedback doesn't do anything on click, and a lot of the time, the icon is only presentational.

static/js/janitor-new.js
nt1m: I only put `...` when pending for now, but if feel free to adjust this in a pull request once this is merged.

static/css/janitor-new.css
```diff
@@ -0,0 +1,440 @@
...
+    background-color: #ededf0;
+  }
+  main {
+    padding: 1em;
+  }
+}
+
+/* Footer */
+
+.footer {
```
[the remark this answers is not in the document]
nt1m: I'll add some padding.

As for the font-size, I'll see how I can improve it.

static/css/janitor-new.css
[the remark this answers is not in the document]
nt1m: I'm not sure how you'd like to change the padding, but I definitively agree on the underline+background part (I unintentionally left it)

static/css/janitor-new.css
nt1m: I don't have a strong opinion on this. Pushing down the content was the easier option technically speaking, but I'll see if I can avoid that (using a transform animation for example).

static/css/janitor-new.css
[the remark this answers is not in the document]
nt1m: I've added some bottom padding to the footer and reduced the font-size to 1.1em (it was 1.2em)

static/css/janitor-new.css
```diff
@@ -0,0 +1,440 @@
...
+}
+
+.navbar-brand {
+  display: flex;
+  align-items: center;
+  flex: 1;
+  text-decoration: none;
+  color: inherit;
+  font-weight: 600;
+  padding: 1em 1em;
```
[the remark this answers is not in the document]
nt1m: Fixed, thanks!

static/js/janitor-new.js
jankeromnes: That sounds like a great idea! Thanks a lot 😄 (also, nit: please use the unicode `…`).

jankeromnes:
Maybe there is no need to make it more complex. I like the fact that the
content never gets hidden by our UI. :)

PR #181, opened 2017-10-24
Refactor OAuth2 and GitHub helpers to async/await, import GitHub emails, and fix a bug.

jankeromnes:


lib/github.js
```diff
@@ -25,49 +25,40 @@ function load () {
...
+      state,
+    }
+  };
 
-    oauth2.getAuthorizationUrl(parameters, (error, authorizeUrl) => {
-      callback(error, authorizeUrl);
-    });
-  });
+  const authorizeUrl = await oauth2.getAuthorizationUrl(parameters);
+  return authorizeUrl;
```
nt1m: `return oauth2.getAuthorizationUrl(parameters);` is equivalent, no need for the variable in between.

lib/github.js
```diff
@@ -131,41 +113,23 @@ exports.request = function (method, path, data, accessToken, callback) {
...
-    if (error) {
-      callback(error);
-      return;
-    }
-
-    callback(null, profile);
-  });
+exports.getUserProfile = async function (accessToken) {
+  const profile = await fetchGitHubAPI('GET', '/user', null, accessToken);
+  return profile;
```
nt1m: You can simply do:

```js
exports.getUserProfile = (accessToken) => fetchGithubAPI('GET', '/user', null, accessToken); 
```

This works the same.

lib/github.js
```diff
@@ -131,41 +113,23 @@ exports.request = function (method, path, data, accessToken, callback) {
...
-  exports.request('GET', url, null, null, (error, sshPublicKeys) => {
-    if (error) {
-      callback(error);
-      return;
-    }
-
-    callback(null, sshPublicKeys);
-  });
+  const sshPublicKeys = await fetchGitHubAPI('GET', url, null, null);
+  return sshPublicKeys;
```
nt1m: Same here, you can use a one liner:

```js
exports.getSSHPublicKeys = (username) => fetchGitHubAPI('GET', `/users/${username}/keys`, null, null);

File: lib/users.js [hunk 4]
```diff
@@ -96,12 +96,7 @@ exports.resetSSHKeyPair = function (user) {
 
 // Refresh a user's GitHub account details using an OAuth2 access token.
 exports.refreshGitHubAccount = function (user, accessToken, refreshToken) {
```
nt1m: This function would look a lot cleaner with async/await

jankeromnes: Actually, I don't think it would look that much cleaner, because we don't want to actually `await` the two Promises below (as you can see, we don't currently block on them, but instead we resolve `refreshGitHubAccount` early while letting the two promises finish in their own time).

So even with async/await, we'd still have `.then()` and `.catch()` on `github.getSSHPublicKeys()` and `github.getVerifiedEmails()` instead of `await`.

nt1m: That would still be cleaner as we'd have less indentation :)

jankeromnes:
@nt1m Thanks again for your review! I believe I've addressed all your concerns. Please have another look if you have the time. 😃

app.js
```diff
@@ -170,15 +170,16 @@ boot.executeInParallel([
...
-      if (error) {
-        log('[fail] github authenticate', error);
-        routes.notFoundPage(response, user);
-        return;
-      }
-
-      users.refreshGitHubAccount(user, accessToken, refreshToken);
-      routes.redirect(response, '/settings/integrations/');
+    github.authenticate(request).then(({ accessToken, refreshToken }) => {
+      users.refreshGitHubAccount(user, accessToken, refreshToken).then(() => {
```
nt1m: This could use async/await

lib/github.js
```diff
@@ -131,41 +112,20 @@ exports.request = function (method, path, data, accessToken, callback) {
...
+exports.getVerifiedEmails = async function (accessToken) {
+  const emails = await fetchGitHubAPI('GET', '/user/emails', null, accessToken);
 
-    // Don't trust un-verified email addresses.
-    const verifiedEmails = emails.filter(email => email.verified);
-    callback(null, verifiedEmails);
-  });
+  // Don't trust un-verified email addresses.
+  const verifiedEmails = emails.filter(email => email.verified);
+  return verifiedEmails;
```
nt1m: You can avoid the intermediary variable.

lib/oauth2.js
```diff
@@ -25,73 +23,66 @@ function getClient (providerId, callback) {
...
-      callback(error);
-      return;
-    }
-
-    const authorizeUrl = client.getAuthorizeUrl(options);
-    callback(null, authorizeUrl);
-  });
+  const { client } = await getClient(provider);
+  const authorizeUrl = client.getAuthorizeUrl(options);
+  return authorizeUrl;
```
nt1m: Same here, you could also avoid the intermediary variable.

app.js
jankeromnes: Indeed, although the error handling makes the async/await version actually more verbose: https://github.com/JanitorTechnology/janitor/pull/181/commits/61bf7dbadb863f7f5564a1e6104ceac2cbb9dfa2

nt1m: Looks fine by me :) Less indentation!

jankeromnes:
Thanks for all the helpful reviews! 👍 I think this looks good now.

PR #185, opened 2017-11-06
Automatically remove used OAuth2 code and state parameters from proxy URLs.

jankeromnes:
When you use a Janitor proxy for the first time (i.e. you have no authentication cookie for that particular proxy host), a "transparent" OAuth2 round-trip happens with the main janitor.technology app in order to authenticate you.

But when your very first proxy request is finally authenticated, you're left with two additional parameters in your URL, something like `?code=1234&state=5678`. The problem is that these parameters are single-use, and by the time you see the requested page appear, they're already expired.

This means that if you try to refresh the page (with both parameters still in the URL), the proxy will consider this a login attempt, but reject the expired code and state, resulting in a "403 Forbidden" page.

The solution against this is, for URLs like `https://proxy.janitor.technology/abcdef/8080/?code=1234&state=5678`, to redirect to the "safer" URL `https://proxy.janitor.technology/abcdef/8080/` instead of directly serving the requested content.

### DIFF
--- a/join.js
+++ b/join.js
@@ -4,6 +4,7 @@
 const camp = require('camp');
 const http = require('http');
 const nodepath = require('path');
+const nodeurl = require('url');
 
 const boot = require('./lib/boot');
 const db = require('./lib/db');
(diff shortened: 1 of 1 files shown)


jankeromnes:
@bnjbvr This is a relatively small fix. Could you please double-check my logic when you have some time? 😃

join.js
```diff
@@ -130,7 +131,25 @@ function handleOAuth2Code (request, response, next) {
...
+
+    const requestUrl = nodeurl.parse(request.url);
+    if (!requestUrl.search) {
+      // There are no URL parameters to remove, proceed without redirection.
+      next();
+      return;
+    }
+
+    // Remove the used OAuth2 code and state parameters from the requested URL.
+    const oldUrlParameters = requestUrl.search.slice(1).split('&');
```
bnjbvr: Stupid question, but is any encode/decodeURIComponent needed here?

jankeromnes: I think it is safe to leave everything URL-encoded, because we don't need to decode URL parameters to filter out `code` and `state`.

jankeromnes:
Thank you!

PR #188, opened 2017-11-08
Sultiple multiple hostnames.

jankeromnes:


notriddle:
[Don't use .io](https://getstream.io/blog/stop-using-io-domain-names-for-production-traffic/)

Have you considered using a .tech domain? It's a reasonable length, and it's not being run by a government branch that has no idea what they're doing.

jankeromnes:
> /home/travis/build/JanitorTechnology/janitor/join.js
>  255:5  error  Unexpected lexical declaration in case block  no-case-declarations

For real? 😐

jankeromnes:
@bnjbvr If you have some time, could you please have a look at these changes?

I basically:
- Removed unused code
- Replaced `"hostname": "single.hostname.com"` with `"hostnames": [ "multiple.com", "hostnames.com" ]` in `db.json`

This is in order to use [janitor.technology](https://janitor.technology) and [jntr.io](http://jntr.io) interchangeably.

notriddle:
In the interest of SEO, `<link rel="canonical" href="{{hostnames[0]}}/{{path}}">` seems like something you'd need?

jankeromnes:
Great catch @notriddle, thanks! I'll add that to our header templates.

jankeromnes:
@bnjbvr or @notriddle, I've rebased this change, and added canonical URLs to every page.

Please have a look whenever it's convenient for you. 😄

app.js
```diff
@@ -38,13 +38,13 @@ boot.executeInParallel([
...
-    hostname + ':' + ports.https);
+    hostnames[0] + ':' + ports.https);
 
   // Protect the server and its users with a security policies middleware.
   const enforceSecurityPolicies = (request, response, next) => {
-    // Only accept requests addressed to our actual hostname.
+    // Only accept requests addressed to our actual hostnames.
     const requestedHostname = request.headers.host;
-    if (requestedHostname !== hostname) {
+    if (!requestedHostname || hostnames.indexOf(requestedHostname) < 0) {
```
bnjbvr: nit: `!hostnames.includes(requestedHostname)`

join.js
```diff
@@ -14,15 +14,15 @@ const proxyHeuristics = require('./lib/proxy-heuristics');
...
 
-// Change this to your actual hostname in './db.json':
-const hostname = db.get('hostname', 'localhost');
+// Add your actual server hostnames in './db.json':
+const hostnames = db.get('hostnames', [ 'localhost' ]);
 
-if (!hostname || hostname === 'localhost') {
-  throw new Error('Cannot join cluster as [hostname = ' + hostname + ']: ' +
+if (!hostnames || hostnames[0] === 'localhost') {
+  throw new Error('Cannot join cluster as [hostname = ' + hostnames[0] + ']: ' +
```
bnjbvr: nit: here and below you could use template strings

bnjbvr: A suggestion: instead of having this file doing this check, what about having a helper somewhere else (any `utils`-like file) that has a `function mainHostname() {}`, and moving the assert in this same file? It would be less custom than all the `hostnames[0]` scattered in all the files, in case the way to store the main hostname ever changes.

lib/certificates.js
```diff
@@ -90,10 +90,12 @@ exports.isValid = function (parameters, enableDebug) {
         .filter(altName => altName.type === 2) // DNS
         .map(altName => altName.value);
     }
-    if (altNames.indexOf(hostname) < 0) {
-      debug('Certificate is not valid for hostname "' + hostname + '" ' +
-        '(should be one of: "' + altNames.join('", "') + '").');
-      return false;
+    for (const hostname of hostnames) {
+      if (altNames.indexOf(hostname) < 0) {
```
bnjbvr: nit: `!altNames.includes(hostname)`

PR #191, opened 2017-11-15
Update Firefox SVG logo to Quantum.

jankeromnes:
While we're at it, let's also update the Firefox logo to the more expected Quantum logo.


notriddle:
Weirdly, if I try to view this SVG from within GitHub itself, it shows it too small. I think it's related to it not having width or height attributes...

jankeromnes:
Hm, you're right, that's pretty strange. It showed up correctly for me when I created this pull request, but now I only see top very top of the logo (as if it was cropped).

jankeromnes:
Now live at https://janitor.technology/img/firefox.svg and I switched both Firefox project logos to the new one: https://janitor.technology/projects/

PR #193, opened 2017-11-16
Configure Docker daemon with IPv6 enabled by default.

jankeromnes:
Enabling IPv6 would add support for rails 5.0.0.1, which is needed for the Discourse image: https://github.com/JanitorTechnology/dockerfiles/pull/107#issuecomment-344772486

Before merging this, we should test enabling IPv6 on our existing Docker hosts, to make sure everything still works with it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ docker: docker.ca docker.crt docker.key
 	printf "[Service]\nExecStart=\nExecStart=/usr/bin/dockerd\n" | sudo tee /etc/systemd/system/docker.service.d/simple_dockerd.conf # work around "-H fd://" conflict
 	sudo systemctl daemon-reload
 	sudo cp /etc/docker/daemon.json /etc/docker/daemon.json.old 2>/dev/null; true # backup any prior daemon.json, but ignore errors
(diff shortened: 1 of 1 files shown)


jankeromnes:
This causes the following error, preventing the Docker daemon from starting:

```
Nov 16 16:10:14 pianosa dockerd[11862]: Error starting daemon: Error initializing network controller: Error creating default "bridge" network: could not find an available, non-overlapping IPv6 address pool among the defaults to assign to the network
```

Seems like we need to specify a `--subnet`, or a `--fixed-cidr-v6`, even if we don't want globally routed IPv6 addresses. See https://github.com/moby/moby/issues/29443#issuecomment-267401488

notriddle:
We can work around this by specifying `-b 0.0.0.0` when starting the Discourse Rails app, so this shouldn't be needed right away.

jankeromnes:
Thank you @notriddle for implementing the workaround!

Since enabling IPv6 in our Docker servers isn't necessary right now, and we don't know what a good value for `--subnet` or `--fixed-cidr-v6` looks like, let's give up on this for now.

To anyone in the future, please re-open this issue if things change or if you know good values for either `--subnet` or `--fixed-cidr-v6` mandatory IPv6 options.

PR #194, opened 2017-11-16
Do not strip the path prefix from Discourse

notriddle:
Discourse supports path prefixes, as long as you leave them in when forwarding the request on.


notriddle:
We should probably do it with an actual configuration option, instead of hard-coding port numbers.

notriddle:
Obsoleted by #199

PR #195, opened 2017-11-22
Fix allowing admin users to sign in without invitation.

jankeromnes:
This is a small regression from a previous refactor of user emails (it only slightly affects Janitor users during development, but it's still nice to fix it).

@nt1m could you please have a look? 😄


jankeromnes:
Thanks!

PR #198, opened 2017-11-26
Further tweak the colors and styles

notriddle:
* Fix the footer pinning behavior, so that it works like the old site

* Use Yellow90 for the text in the alert box

<details><summary>Previous revision of this pull request made more radical changes</summary>

* Fix the footer pinning behavior, so that it works like the old site

* Use Yellow90 for the text in the alert box (as described in the Photon docs)

* Use Gray10 for the site background, reserving White only for pop-ups, floating boxes, and active form fields (as described in the Photon docs)

* Recolor the footer in Photon Ink (this is the part I'm most willing to revert, making the footer the same as the rest of the document)
</details>


static/css/janitor-new.css
```diff
@@ -4,8 +4,9 @@
   box-sizing: border-box;
 }
 
-html, body {
-  min-height: 100%;
+html {
+  height: 100%;
+  overflow-y: scroll;
```
jankeromnes: Interesting, why did you add this rule?

static/css/janitor-new.css
```diff
@@ -16,6 +17,7 @@ body {
   font-size: 14px;
   display: flex;
   flex-direction: column;
+  min-height: 100%;
```
jankeromnes: In testing, I was able to fix the "pinned footer" by just using `height: 100%` instead of `min-height: 100%` on just the body, which seems a little simpler, but since I don't know anything about CSS I'll defer the review to experts like @nt1m and @arshdkhn1.

jankeromnes:
Ping @nt1m (please have a look at this when convenient) 😄

static/css/janitor-new.css
nt1m: I started by using height: 100%, but it seems to squish the header when there is a lot of content in the page.

static/css/janitor-new.css
```diff
@@ -125,12 +129,14 @@ main {
...
   background: #fff;
   color: #0b0b0d;
-  box-shadow: 0 4px 16px rgba(12,12,13,.1);
-  overflow: auto;
   z-index: 999;
   font-size: 1.1em;
+}
+
+.navbar-sticky {
   position: sticky;
```
arshdkhn1: `position: sticky` not supported in IE and old browsers?

static/css/janitor-new.css
```diff
@@ -4,8 +4,11 @@
   box-sizing: border-box;
 }
 
-html, body {
-  min-height: 100%;
+html {
+  height: 100%;
+  background-color: #f9f9fa;
+  color: #000;
+  overflow-y: scroll;
```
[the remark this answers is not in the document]
notriddle: I want the scroll bar to appear all the time. Otherwise, the centered content jumps around when you switch pages.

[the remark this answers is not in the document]
arshdkhn1: @notriddle I also didnt like that jumping but I guess that's fine. Only chrome has that jumping behaviour. there is a margin technique to avoid scrollbar and jumping

notriddle:
Okay. Gray background and ink footer are going away as soon as I'm back at my computer.

static/css/janitor-new.css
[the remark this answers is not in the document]
arshdkhn1: but that margin technique is not worth the pain.

static/css/janitor-new.css
notriddle: Actually, it's mostly so that I don't have to override all this stuff on the footer.

notriddle:
Fixed.

static/css/janitor-new.css
```diff
@@ -4,8 +4,11 @@
   box-sizing: border-box;
 }
 
-html, body {
-  min-height: 100%;
+html {
+  height: 100%;
+  background-color: #fff;
+  color: #0b0b0d;
+  overflow-y: scroll;
```
nt1m: overflow-y: auto;

notriddle: I *intentionally* put in a permanent scrollbar. It's to keep the centered content from jumping around as it goes from being big enough to scroll to being small enough to not scroll.

nt1m: I personally prefer having the content jumping for a tad bit, than having a permanent scrollbar.

notriddle:
Based on ntim's approval, I guess it's suitable.

PR #199, opened 2017-11-27
Add support for proxying HTTPS without URL rewrite

notriddle:
The current behavior is put under the `https-auto` proxy type, and the new behavior is `https` itself.

### DIFF
--- a/lib/proxy-heuristics.js
+++ b/lib/proxy-heuristics.js
@@ -37,13 +37,16 @@ exports.handleProxyUrls = function (request, response, next) {
     }
 
     // Locally remove the prefix from `request.url`.
-    request.url = request.url.replace('/' + containerId + '/' + port, '');
+    request.query.url = request.url.replace('/' + containerId + '/' + port, '');
   } else if (request.headers.referer) {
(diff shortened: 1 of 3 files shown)


lib/proxy-heuristics.js
```diff
@@ -37,13 +37,16 @@ exports.handleProxyUrls = function (request, response, next) {
...
+    request.query.url = request.url.replace('/' + containerId + '/' + port, '');
   } else if (request.headers.referer) {
     // Look for a container ID and port in `request.headers.referer`.
     const referer = nodeurl.parse(request.headers.referer);
     match = exports.proxyUrlPrefix.exec(referer.pathname);
     if (match) {
       [url, containerId, port, path] = match;
+      // Since there is no prefix,
+      // treat the full requested URL as the "stripped" version.
+      request.query.url = request.url;
```
notriddle: `request.query.url` is always run through `nodepath.normalize` whenever it is used, so I wonder if it would be better to normalize it here, where it is created, instead of at the edges where it is used?

notriddle:
Found a way to make Discourse work with URL rewriting.

PR #203, opened 2017-12-02
Alpha: Mark 9.

jankeromnes:


jankeromnes:
Now that all remaining issues for `0.0.9` have been resolved, it's time to officially mark our 9th Alpha. I also seized this opportunity to update our `dockerfiles` Git submodule to its latest commit.

After this is merged, I'll update our various deployments (we should really find a way to automate this) and also our npm package (which is not super useful right now, but it'll update our README badge).

jankeromnes:
Thanks @nt1m! 🍰

PR #204, opened 2017-12-02
Mark the "Preview" port as the "preview" port

notriddle:
https://github.com/JanitorTechnology/c9.ide.janitorconfig/pull/9


jankeromnes:
Thank you @notriddle for the fix and @nt1m for the review! In cases like these where a second review isn't needed, feel free to merge the change as well. :smile:

PR #205, opened 2017-12-04
Add Kresus SVG logo.

jankeromnes:
I stole this from https://github.com/bnjbvr/kresus/blob/master/static/images/favicon/safari-pinned-tab.svg

@bnjbvr Please have a look. 😄 This would eventually be hosted at https://janitor.technology/img/kresus.svg


nt1m:
I've optimized the SVG: https://gist.github.com/nt1m/d1fc65bc87de3c0cbc4981cc68196c99

bnjbvr:
r=me for the optimized SVG; note there's going to be another logo later. Probably we'd want to keep hosting it ourselves, which would prevent some back and forth. Any chance we could provide you with a link hosted on kresus.org instead? Thanks!

jankeromnes:
> I've optimized the SVG: https://gist.github.com/nt1m/d1fc65bc87de3c0cbc4981cc68196c99

So cool! I've updated the PR with it. Thanks!

> r=me for the optimized SVG

Thank you! 😄 

> note there's going to be another logo later

Looking forward to it. ✨

> Probably we'd want to keep hosting it ourselves, which would prevent some back and forth. Any chance we could provide you with a link hosted on kresus.org instead? Thanks!

I agree this would be better going forward, but in the meantime I don't see the harm in merging this pull request so thanks again for the r+. 😄 

Once you're hosting the SVG logo (e.g. on kresus.org), please give me the link (e.g. here or on IRC) and I'll modify the Kresus meta-data on Janitor to point directly to it. Note that there is no rush, and this can wait for the release of your new logo. 👍

PR #208, opened 2017-12-08
Implement a Discourse-backed blog page

notriddle:
How it works
=========

Create the topic in the Newsletter category. If you don't tag it as `#published`, it won't show up. Once you tag it as published (and only staff have permission to tag a topic as published), it will show up on Janitor's blog page.

Also, the webhook payload is not read, and the signature is not validated. So you can manually trigger a poll by pointing your browser at that URL.

Importing old entries
=========

The "slug" in Discourse, which can be seen in the URL, becomes the `#id` of the post on Janitor (for example, the slug of this topic is `newsletters-composed-in-discourse-and-published-in-janitor`). If you don't want to break old links, make sure the post title will result in a correct slug.

Also, you can click the admin wrench and change the topic timestamp. The Janitor's blog page sorts posts by their timestamp, so old newsletter entries will line get sorted correctly.

### DIFF
--- a/api/blog-api.js
+++ b/api/blog-api.js
@@ -4,6 +4,7 @@
 const selfapi = require('selfapi');
 
 const blog = require('../lib/blog');
+const log = require('../lib/log');
 
 // API resource to manage Janitor's Discourse-backed news section.
 const blogAPI = module.exports = selfapi({
@@ -18,7 +19,7 @@ webhookAPI.get({
   description: 'Pull the blog section from Discourse.',
 
   handler: (_request, response) => {
-    blog.pull();
+    blog.pull().then(null, log);
     response.json({}, null, 2);
   },
 
--- a/api/blog-api.js
+++ b/api/blog-api.js
@@ -11,21 +11,27 @@ const blogAPI = module.exports = selfapi({
   title: 'Blog'
 });
 
-// API sub-resource for the sync webhook.
-const webhookAPI = blogAPI.api('/sync-webhook');
-
-webhookAPI.get({
-  title: 'Synchronize',
+// API sub-resource for the Discourse-Blog synchronization webhook.
+// should show up as https://janitor.technology/api/blog/synchronize
+blogAPI.get('/synchronize', {
+  title: 'Synchronize Blog',
   description: 'Pull the blog section from Discourse.',
 
-  handler: (_request, response) => {
-    blog.pull().then(null, log);
-    response.json({}, null, 2);
+  handler: async (_request, response) => {
+    try {
+      const status = await blog.synchronize();
+      log('synchronized blog', status);
+      response.json(status, null, 2);
+    } catch (error) {
+      log('[fail] synchronized blog', error);
+      response.statusCode = 500; // Internal Server Error
+      response.json({ error: 'Could not synchronize' }, null, 2);
+    }
   },
 
   examples: [{
     response: {
-      body: JSON.stringify({}, null, 2)
+      body: JSON.stringify({count: 1}, null, 2)
     }
   }]
 });
--- a/api/blog-api.js
+++ b/api/blog-api.js
@@ -1,0 +1,30 @@
+// Copyright © 2017 Michael Howell. All rights reserved.
+// The following code is covered by the AGPL-3.0 license.
+
+const selfapi = require('selfapi');
+
+const blog = require('../lib/blog');
+
+// API resource to manage Janitor's Discourse-backed news section.
+const blogAPI = module.exports = selfapi({
+  title: 'Blog'
+});
+
+// API sub-resource for the sync webhook.
+const webhookAPI = blogAPI.api('/sync-webhook');
+
+webhookAPI.get({
+  title: 'Synchronize',
+  description: 'Pull the blog section from Discourse.',
+
+  handler: (_request, response) => {
+    blog.pull();
+    response.json({}, null, 2);
+  },
+
+  examples: [{
+    response: {
+      body: JSON.stringify({}, null, 2)
+    }
+  }]
+});
--- a/api/blog-api.js
+++ b/api/blog-api.js
@@ -11,17 +11,15 @@ const blogAPI = module.exports = selfapi({
   title: 'Blog'
 });
 
-// API sub-resource for the Discourse-Blog synchronization webhook.
-// should show up as https://janitor.technology/api/blog/synchronize
 blogAPI.get('/synchronize', {
   title: 'Synchronize Blog',
   description: 'Pull the blog section from Discourse.',
 
   handler: async (_request, response) => {
     try {
-      const status = await blog.synchronize();
-      log('synchronized blog', status);
-      response.json(status, null, 2);
+      const { count } = await blog.synchronize();
+      log('synchronized blog', { count });
+      response.json({ count }, null, 2);
     } catch (error) {
       log('[fail] synchronized blog', error);
       response.statusCode = 500; // Internal Server Error
@@ -31,7 +29,7 @@ blogAPI.get('/synchronize', {
 
   examples: [{
     response: {
-      body: JSON.stringify({count: 1}, null, 2)
+      body: JSON.stringify({ count: 1 }, null, 2)
     }
   }]
 });
--- a/app.js
+++ b/app.js
@@ -119,12 +119,6 @@ boot.executeInParallel([
     routes.blogPageNew(query.res, {user, blog: blog.getDb()});
   });
 
-  // New public blog page webhook.
-  app.route(/^\/blog-new-webhook\/?$/, (data, match, end, query) => {
-    log('blog-new-webhook');
-    routes.blogWebhookNew(query.res);
-  });
-
   // Public live data page.
   app.route(/^\/data\/?$/, (data, match, end, query) => {
     const { user } = query.req;
--- a/app.js
+++ b/app.js
@@ -117,7 +117,7 @@ boot.executeInParallel([
     const { req: request, res: response } = query;
     const { user } = request;
     log('blog-new');
-    routes.blogPageNew(response, {blog, user, topics: blog.getDb().topics});
+    routes.blogPageNew(response, user, blog);
   });
 
   // Public live data page.
--- a/app.js
+++ b/app.js
@@ -114,9 +114,10 @@ boot.executeInParallel([
 
   // New public blog page.
   app.route(/^\/blog-new\/?$/, (data, match, end, query) => {
-    const { user } = query.req;
+    const { req: request, res: response } = query;
+    const { user } = request;
     log('blog-new');
-    routes.blogPageNew(query.res, {user, blog: blog.getDb()});
+    routes.blogPageNew(response, {blog, user, topics: blog.getDb().topics});
   });
 
   // Public live data page.
--- a/app.js
+++ b/app.js
@@ -6,6 +6,7 @@ const nodepath = require('path');
 const selfapi = require('selfapi');
 
 const api = require('./api/');
+const blog = require('./lib/blog');
 const boot = require('./lib/boot');
 const db = require('./lib/db');
 const github = require('./lib/github');
@@ -109,6 +110,19 @@ boot.executeInParallel([
     const { user } = query.req;
     log('blog');
     routes.blogPage(query.res, user);
+  });
+
+  // New public blog page.
+  app.route(/^\/blog-new\/?$/, (data, match, end, query) => {
+    const { user } = query.req;
+    log('blog-new');
+    routes.blogPageNew(query.res, {user, blog: blog.getDb()});
+  });
+
+  // New public blog page webhook.
+  app.route(/^\/blog-new-webhook\/?$/, (data, match, end, query) => {
+    log('blog-new-webhook');
+    routes.blogWebhookNew(query.res);
   });
 
   // Public live data page.
--- a/lib/blog.js
+++ b/lib/blog.js
@@ -1,0 +1,85 @@
+// Copyright © 2015 Jan Keromnes. All rights reserved.
+// The following code is covered by the AGPL-3.0 license.
+
+const db = require('./db');
+const https = require('https');
+const metrics = require('./metrics');
+
+const fetchDiscourseAPI = function (url) {
+  return new Promise((resolve, reject) => {
+    const doResolve = (succ) => {
+      let results = '';
+      succ.on('data', (chunk) => {
+        try {
+          results += chunk;
+        } catch (e) {
+          succ.destroy(e);
+        }
+      });
+      succ.on('error', reject);
+      succ.on('end', () => {
+        try {
+          resolve(JSON.parse(results));
+        } catch (e) {
+          reject(e);
+        }
+      });
+    };
+    https.get(url + '.json', doResolve);
+  });
+};
+
+const getDb = function () {
+  return db.get('blog', {
+    'posts': [],
+    'topics_url': 'https://discourse.janitor.technology/tags/published',
+    'post_url_template': 'https://discourse.janitor.technology/t/{{slug}}/{{id}}/1',
+    'comments_url_template': 'https://discourse.janitor.technology/t/{{slug}}/{{id}}/2',
+    'filter_category_id': 9,
+    'data': {
+      'updated': null,
+      'pull-time': [
+      ],
+    },
+  });
+};
+
+const pull = async function () {
+  const blog = getDb();
+  const topic = await fetchDiscourseAPI(blog.topics_url);
+  blog.posts = topic.topic_list.topics;
+  const time = Date.now();
+  for (const post_i in blog.posts) {
+    if (blog.posts[post_i].category_id !== blog.filter_category_id) {
+      delete blog.posts[post_i];
+    }
+  }
+  blog.posts.sort(function (a, b) {
+    let ret_val;
+    if (a.created_at > b.created_at) {
+      ret_val = -1;
+    } else if (a.created_at === b.created_at) {
+      ret_val = 0;
+    } else {
+      ret_val = 1;
+    }
+    return ret_val;
+  });
+  for (const post_i in blog.posts) {
+    const post_meta = blog.posts[post_i];
+    const post_body_url = blog.post_url_template
+      .replace('{{slug}}', post_meta.slug)
+      .replace('{{id}}', post_meta.id);
+    const post_body = (await fetchDiscourseAPI(post_body_url)).post_stream.posts[0].cooked;
+    blog.posts[post_i].body_html = post_body;
+  }
+  const now = Date.now();
+  metrics.set(blog, 'updated', now);
+  metrics.push(blog, 'pull-time', [now, Date.now() - time]);
+  db.save();
+};
+
+module.exports = {
+  'pull': pull,
+  'getDb': getDb,
+};
--- a/lib/blog.js
+++ b/lib/blog.js
@@ -68,7 +68,7 @@ exports.getCommentsUrl = function (topic) {
 exports.synchronize = async function () {
   const blog = module.exports.getDb();
   const time = Date.now();
-  let topics = await fetchDiscourseAPI(discourseBlogTopicsUrl).topic_list.topics;
+  let topics = (await fetchDiscourseAPI(discourseBlogTopicsUrl)).topic_list.topics;
   // Remove "published" topics that are not actual blog posts.
   for (const i in topics) {
     if (topics[i].category_id !== discourseBlogCategoryID) {
@@ -88,7 +88,7 @@ exports.synchronize = async function () {
   });
   blog.topics = await Promise.all(topics.map(async topic => {
     const bodyUrl = module.exports.getPostUrl(topic);
-    const bodyHtml = await fetchDiscourseAPI(bodyUrl).post_stream.posts[0].cooked;
+    const bodyHtml = (await fetchDiscourseAPI(bodyUrl)).post_stream.posts[0].cooked;
     topic.post_body_html = bodyHtml;
     return topic;
   }));
--- a/lib/blog.js
+++ b/lib/blog.js
@@ -21,17 +21,17 @@ const discourseBlogCommentsUrlTemplate = 'https://discourse.janitor.technology/t
 // Perform an asynchronous Discourse API request.
 function fetchDiscourseAPI (url) {
   return new Promise((resolve, reject) => {
-    https.get(url + '.json', response_stream => {
+    https.get(url + '.json', responseStream => {
       let json = '';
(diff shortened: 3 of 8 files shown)


jankeromnes:
So cool! Thanks a lot for porting the Blog page over to our new design, and also for preparing it for pulling from our Discourse!

I wanted to have a look this morning but then I got distracted. Let's try again. 😄

lib/routes.js
```diff
@@ -125,6 +133,25 @@ exports.blogPage = function (response, user = null) {
...
+const blogSectionNew = camp.template('./templates/blog-new.html');
+exports.blogPageNew = function (response, params) {
+  const title = 'Blog';
+  const {user, blog} = params;
+
+  response.template({
+    title,
+    user,
+    blog,
+    log,
```
jankeromnes: Interesting, why are you passing `log` to the template? 🙂

templates/blog-new.html
```diff
@@ -0,0 +1,14 @@
+    <section class="blog">
+      <h2 class="header">Blog</h2>
+      {{
```
jankeromnes: Nit: Our templating system is a bit weird, but I like to set up indentation just right so that the produced HTML doesn't contain suspicious whitespaces. Here, everything outside `{{` and `}}` will be included in the document, i.e. several trailing spaces and then a new line. To work around this, you can do something like:

```html
      <h2 class="header">Blog</h2>{{
        for (let post in blog.posts) {
          const {title, body} = blog.posts[post];
      }}
      <div class="panel panel-default">
        ...
      </div>{{
        }
      }}
    </section>
```

templates/blog-new.html
```diff
@@ -0,0 +1,14 @@
...
+        for (let post in blog.posts) {
+          const {title, body} = blog.posts[post];
+      }}
+      <div class="panel panel-default">
+        <div class="panel-heading"><h3>{{= title in html}}</h3></div>
+        <div class="panel-body"><section>{{= body}}</section></div>
```
jankeromnes: So the title will be successfully escaped, but how do we prevent `body` from including nefarious `<script>` tags that steals user credentials or performs random API calls? Do we trust everything we'll import from Discourse?

lib/routes.js
notriddle: Erm, I was debugging stuff. It's not needed any more...

templates/blog-new.html
notriddle: I would just trust everything we import from Discourse.

notriddle:
The blog is now stored in the database, and synced with Discourse when `/blog-webhook-new/` is called. Which it will be whenever anything happens in the 'newsletter' category (I configured a Discourse webhook for that).

notriddle:
To avoid messing up link-buttons, I moved the link styling over to being blog-specific.

lib/blog.js
```diff
@@ -0,0 +1,85 @@
+// Copyright © 2015 Jan Keromnes. All rights reserved.
```
jankeromnes: Nit: Copyright 2017 @notriddle

lib/routes.js
```diff
@@ -149,6 +150,29 @@ exports.blogPage = function (response, user = null) {
...
+  }, [
+    appHeaderNew,
+    blogSectionNew,
+    appFooterNew,
+  ]);
+};
+
+exports.blogWebhookNew = function (response) {
+  blog.pull();
+  response.json({}, null, 2);
```
jankeromnes: Nit: I feel like this web hook would make more sense within the [Janitor API](https://janitor.technology/reference/api/) (since you're returning JSON anyway).

The API handler implementations are in [/api/](https://github.com/JanitorTechnology/janitor/tree/master/api), and you could create a new `blog-api.js` file with just the webhook for now.

notriddle:
Moved the webhook.

app.js
```diff
@@ -111,6 +112,19 @@ boot.executeInParallel([
...
 
+  // New public blog page.
+  app.route(/^\/blog-new\/?$/, (data, match, end, query) => {
+    const { user } = query.req;
+    log('blog-new');
+    routes.blogPageNew(query.res, {user, blog: blog.getDb()});
+  });
+
+  // New public blog page webhook.
+  app.route(/^\/blog-new-webhook\/?$/, (data, match, end, query) => {
```
jankeromnes: Nit: Since you moved the webhook to the API, please remove this handler.

api/blog-api.js
```diff
@@ -0,0 +1,30 @@
...
+
+// API sub-resource for the sync webhook.
+const webhookAPI = blogAPI.api('/sync-webhook');
+
+webhookAPI.get({
+  title: 'Synchronize',
+  description: 'Pull the blog section from Discourse.',
+
+  handler: (_request, response) => {
+    blog.pull();
```
jankeromnes: Nit: Don't you want to handle async results or errors here?

lib/blog.js
```diff
@@ -0,0 +1,85 @@
...
+    https.get(url + '.json', doResolve);
+  });
+};
+
+const getDb = function () {
+  return db.get('blog', {
+    'posts': [],
+    'topics_url': 'https://discourse.janitor.technology/tags/published',
+    'post_url_template': 'https://discourse.janitor.technology/t/{{slug}}/{{id}}/1',
+    'comments_url_template': 'https://discourse.janitor.technology/t/{{slug}}/{{id}}/2',
```
jankeromnes: Nit: Do you really wish to make these templates editable at run-time, via an Admin page that can change the DB? If not, I think it would be simpler to move these template strings to constant variables at the top of this file.

lib/blog.js
```diff
@@ -0,0 +1,85 @@
...
+};
+
+const pull = async function () {
+  const blog = getDb();
+  const topic = await fetchDiscourseAPI(blog.topics_url);
+  blog.posts = topic.topic_list.topics;
+  const time = Date.now();
+  for (const post_i in blog.posts) {
+    if (blog.posts[post_i].category_id !== blog.filter_category_id) {
+      delete blog.posts[post_i];
```
jankeromnes: Nit: You're importing every blog post, and then deleting the ones that don't match our filter. This is dangerous, because if there is an exception between these two steps, we'll have many unwanted posts in our DB. Instead, it would be great to apply the filter to an intermediary variable, e.g. like so:

```js
const postsToImport = topic.topic_list.topics.filter(post => post.category_id === blog.filter_category_id);
```

lib/blog.js
```diff
@@ -0,0 +1,85 @@
...
+  blog.posts.sort(function (a, b) {
+    let ret_val;
+    if (a.created_at > b.created_at) {
+      ret_val = -1;
+    } else if (a.created_at === b.created_at) {
+      ret_val = 0;
+    } else {
+      ret_val = 1;
+    }
+    return ret_val;
```
jankeromnes: Nit: I find this sorting function a bit too verbose. Please use a shorter variant like:

```js
blog.posts.sort((a, b) => {
  if (a.created_at === b.created_at) {
    return 0;
  }
  return b.created_at < a.created_at;
});
```

lib/blog.js
```diff
@@ -0,0 +1,85 @@
...
+  });
+  for (const post_i in blog.posts) {
```
jankeromnes: Nit: Here you could use a `for (const post of blog.posts)` loop.

lib/blog.js
```diff
@@ -0,0 +1,85 @@
...
+    const post_meta = blog.posts[post_i];
```
jankeromnes: Nit: I find this variable name confusing, and also our style guide generally uses `camelCase` instead of `snake_case`. Additionally, it doesn't seem to be useful (you could use `blog.posts[post_i]` directly as you do below, or simply `post` if you accept my `for of` suggestion above).

lib/blog.js
```diff
@@ -0,0 +1,85 @@
...
+    const post_body_url = blog.post_url_template
+      .replace('{{slug}}', post_meta.slug)
+      .replace('{{id}}', post_meta.id);
+    const post_body = (await fetchDiscourseAPI(post_body_url)).post_stream.posts[0].cooked;
```
jankeromnes: Nit: Instead of using `await` for several blog posts in a row (this fetches them in sequence), I believe it would be more efficient to do them in parallel, with something like this:

```js
// Import (or update) all HTML posts from Discourse in parallel.
const promises = blog.posts.map(post => {
  /* ... */
  return fetchDiscourseAPI(postBodyUrl).then(({ post_stream }) => {
    post.body_html = post_stream.posts[0].cooked;
  }
});

// Wait for all Discourse imports to finish.
await Promise.all(promises);
```

(See also [this Promise.all() example](https://github.com/JanitorTechnology/janitor/blob/9fdf012d489ca59906f46995375e14c684b38e6c/lib/certificates.js#L255-L285).)

lib/blog.js
```diff
@@ -0,0 +1,85 @@
...
+    blog.posts[post_i].body_html = post_body;
+  }
+  const now = Date.now();
+  metrics.set(blog, 'updated', now);
+  metrics.push(blog, 'pull-time', [now, Date.now() - time]);
+  db.save();
```
jankeromnes: Very cool use of `metrics` and `db.save()`! Thanks 👍

static/js/janitor-new.js
```diff
@@ -263,6 +263,10 @@ Array.forEach(document.querySelectorAll('h1[id],h2[id]'), element => {
   element.appendChild(small);
 });
 
+Array.forEach(document.querySelectorAll('.blog article p a'), element => {
+  element.target = '_blank';
```
jankeromnes: Don't all blog links already have a `target="_blank"` attribute? (Seriously asking, I don't know if they do, and if they don't then your code here is probably necessary.)

templates/blog-new.html
```diff
@@ -0,0 +1,22 @@
+    <section class="blog">{{
+        for (let post in blog.posts) {
+          const {title, body_html, slug, id, posts_count} = blog.posts[post];
+          const title_url = blog.post_url_template
+            .replace("{{slug}}", slug)
+            .replace("{{id}}", id);
```
jankeromnes: Nit: Oh, I see why you added the template URLs in the DB instead of using variables global to the `blog` module: You're re-using them here. Instead of adding them to the DB and duplicating the templating code, I suggest adding a "secret getter" to all `post` objects in `/lib/blog.js`, which can be used in this HTML template but won't get serialized to the DB file.

You can do it like so:

```js
if (!post.hasOwnProperty('_titleUrl')) {
  Object.defineProperty(post, '_titleUrl', {
    get () {
      return PostUrlTemplate.replace('{{slug}}', this.slug).replace('{{id}}', this.id);
    }
  });
}
```

Then you'll be able to use `post._titleUrl` in this new blog HTML template.

(See also [this example](https://github.com/JanitorTechnology/janitor/blob/6a184f2ced5adff15c75420246244e6bc6d60592/lib/machines.js#L484-L491).)

templates/blog-new.html
```diff
@@ -0,0 +1,22 @@
...
+          // the newsletter itself counts as a "post" in this "topic"
+          // all other posts are comments
+          const comments_count = posts_count - 1;
+          const comments_url = blog.comments_url_template
+            .replace("{{slug}}", slug)
+            .replace("{{id}}", id);
```
jankeromnes: Nit: Same comment about the "secret getter".

templates/blog-new.html
```diff
@@ -0,0 +1,22 @@
...
+          const url = comments_count > 0 ? comments_url : title_url;
+      }}
+      <article>
+        <h2 id="{{= slug in xmlattr}}">{{= title in html}}</h2>
+        {{= body_html}}
```
jankeromnes: Nit: This looks dangerous to me, even if Discourse promises to strip all `<script>` elements from its cooked HTML. Maybe we should add a Camp template parser that leaves HTML code untouched except for `<script>` elements? (See [these short examples](https://github.com/JanitorTechnology/janitor/blob/743661bdb86c0ad90fa87fb0ba1bfaa7692b500e/lib/routes.js#L15-L28) for how this could be done.)

templates/blog-new.html
```diff
@@ -0,0 +1,22 @@
...
+      </article>
+      <footer><a class="btn" href="{{= url}}">{{= comments_count in html}} comments</a></footer>{{
```
jankeromnes: Nit: Would escaping `{{= url}}` with `in uri` break the URL? (See also the [list of default template parsers](https://github.com/espadrine/fleau#parsers) for brief explanations of what they do.)

api/blog-api.js
```diff
@@ -0,0 +1,31 @@
...
+
+// API sub-resource for the sync webhook.
+const webhookAPI = blogAPI.api('/sync-webhook');
+
+webhookAPI.get({
+  title: 'Synchronize',
+  description: 'Pull the blog section from Discourse.',
+
+  handler: (_request, response) => {
+    blog.pull().then(null, log);
```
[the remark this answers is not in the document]
nt1m: +1 for async/await

static/js/janitor-new.js
nt1m: Might be nice to put this into a custom server-side blog formatter (eg. the one that would remove script tags), rather than janitor.js ?

notriddle: > Don't all blog links already have a target="_blank" attribute?

templates/blog-new.html
notriddle: HTML sanitization is harder than that. [I know](https://github.com/notriddle/ammonia/).

templates/blog-new.html
notriddle: `in uri` mode actually escapes special characters like `://`, since it's intended to be part of the URL, not the whole thing. It would be better to use `in xmlattr`, I guess.

api/blog-api.js
```diff
@@ -0,0 +1,31 @@
...
+});
+
+// API sub-resource for the sync webhook.
+const webhookAPI = blogAPI.api('/sync-webhook');
+
+webhookAPI.get({
+  title: 'Synchronize',
+  description: 'Pull the blog section from Discourse.',
+
+  handler: (_request, response) => {
```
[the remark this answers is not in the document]
notriddle: I usually do that, yeah. It's a compiler-recommended lint in both Rust and Elixir, so I've kinda picked it up.

lib/blog.js
```diff
@@ -0,0 +1,85 @@
...
+    'topics_url': 'https://discourse.janitor.technology/tags/published',
+    'post_url_template': 'https://discourse.janitor.technology/t/{{slug}}/{{id}}/1',
+    'comments_url_template': 'https://discourse.janitor.technology/t/{{slug}}/{{id}}/2',
+    'filter_category_id': 9,
+    'data': {
+      'updated': null,
+      'pull-time': [
+      ],
+    },
+  });
```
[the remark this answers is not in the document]
notriddle: This isn't discourse's database. This is Discourse's *public API*. It isn't going to change very often.

lib/blog.js
notriddle: Is that actually correct? I thought the `Array.prototype.sort` function had to return one of 1, -1, and 0.

notriddle:
Fixed most of the nits.

lib/blog.js
```diff
@@ -0,0 +1,100 @@
...
+        } catch (error) {
+          reject(error);
+        }
+      });
+    });
+  });
+}
+
+// Get the cached Discourse-backed blog.
+module.exports.getDb = function () {
```
nt1m: nit: dropping `module.` is more consistent with the current code, and also works.

lib/blog.js
```diff
@@ -0,0 +1,100 @@
...
+  return discourseBlogCommentsUrlTemplate
+    .replace('{{slug}}', topic.slug)
+    .replace('{{id}}', topic.id);
+};
+
+// Synchronize the Discourse-backed blog into the database.
+module.exports.synchronize = async function () {
+  const blog = module.exports.getDb();
+  const time = Date.now();
+  let topics = (await fetchDiscourseAPI(discourseBlogTopicsUrl)).topic_list.topics;
```
nt1m: nit: you can drop the brackets

Also, can you use destructuring ?

```js
let { topics } = await fetchDiscourseAPI(discourseBlogTopicsUrl).topic_list;
```

static/css/janitor-new.css
```diff
@@ -480,4 +481,84 @@ main {
 .danger-zone {
   color: #3e0200;
   background: rgba(255, 0, 57, 0.1);
-}
\ No newline at end of file
+}
+
+/* Blogs page */
```
nt1m: nit: "Blog page"

static/css/janitor-new.css
```diff
@@ -480,4 +481,84 @@ main {
...
+.blog {
+  font-size: 16px;
+}
```
nt1m: nit: Please add a blank line after each CSS rule to be consistent with the rest of the CSS file

lib/blog.js
```diff
@@ -0,0 +1,100 @@
...
+  return discourseBlogCommentsUrlTemplate
+    .replace('{{slug}}', topic.slug)
+    .replace('{{id}}', topic.id);
+};
+
+// Synchronize the Discourse-backed blog into the database.
+exports.synchronize = async function () {
+  const blog = module.exports.getDb();
+  const time = Date.now();
+  let topics = await fetchDiscourseAPI(discourseBlogTopicsUrl).topic_list.topics;
```
notriddle: @nt1m Are you sure I can remove the brackets here? Because the tests are failing now.

nt1m: Hmm, apparently not. I remember something like this was/is possible, but maybe I'm confusing this with something else.

notriddle:
Okay, I changed it back.

nt1m:
> Okay, I changed it back.

@notriddle My comment about using destructuring still applies though.

notriddle:
Okay, @nt1m. Switched to using destructuring.

static/css/janitor-new.css
```diff
@@ -480,4 +481,98 @@ main {
...
   color: #3e0200;
   background: rgba(255, 0, 57, 0.1);
-}
\ No newline at end of file
+}
+
+/* Blog page */
+
+.blog {
+  font-size: 16px;
```
jankeromnes: Nit: Ok, this is starting to look really good now, but I think the new blog's readability can still be improved.

Please significantly bump up the *font size*, the *line height*, and narrow the *post width*, like in the [old blog](https://janitor.technology/blog/). You can take inspiration from these [CSS rules](https://github.com/JanitorTechnology/janitor/blob/743661bdb86c0ad90fa87fb0ba1bfaa7692b500e/static/css/janitor.css#L455-L466) (especially, `font-size: 18px`, `line-height: 1.6` and `max-width: 700px; padding-left: 25px; padding-right: 25px`).

jankeromnes: Nit: Also, please don't indent titles like `1. Announcing Windows Environments` (I don't like how Discourse indents them, it hurts readability. I also don't like the enormous bullet points, but I can live with them so I won't complain about them officially.)

notriddle: I didn't intent anything. You wrote Markdown Numbered Lists, and that's what Discourse gave me.

I went ahead and changed them to actual headings.

lib/blog.js
```diff
@@ -0,0 +1,102 @@
...
+
+const discourseBlogCategoryID = 9;
+const discourseBlogTopicsUrl = 'https://discourse.janitor.technology/tags/published';
+const discourseBlogPostUrlTemplate = 'https://discourse.janitor.technology/t/{{slug}}/{{id}}/1';
+const discourseBlogCommentsUrlTemplate = 'https://discourse.janitor.technology/t/{{slug}}/{{id}}/2';
+
+// Perform an asynchronous Discourse API request.
+function fetchDiscourseAPI (url) {
+  return new Promise((resolve, reject) => {
+    https.get(url + '.json', response_stream => {
```
jankeromnes: Nit: The project coding style prefers `camelCase` to `snake_case`.

lib/blog.js
```diff
@@ -0,0 +1,102 @@
...
+      let json = '';
+      response_stream.on('data', (chunk) => {
+        try {
+          json += chunk;
```
jankeromnes: Nit: Might be safer with `String(chunk)` instead of just `chunk`.

lib/blog.js
```diff
@@ -0,0 +1,102 @@
...
+        } catch (error) {
+          response_stream.destroy(error);
```
jankeromnes: Are we absolutely certain that `stream.destroy(error)` will trigger the `stream.on('error', reject)` below? If not I'm worried that some errors will go un-noticed.

lib/blog.js
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Get the cached Discourse-backed blog.
+exports.getDb = function () {
+  return db.get('blog', {
+    'topics': [],
+    'data': {
+      'updated': null,
+      'pull-time': [],
+    },
```
jankeromnes: Nit: The `metrics` library will automatically add any missing `data` entries, so you don't really need to pre-fill `data` with empty values here.

lib/blog.js
```diff
@@ -0,0 +1,102 @@
...
+  const topicsPromises = topics.map(async topic => {
+    const bodyUrl = module.exports.getPostUrl(topic);
+    const { post_stream } = await fetchDiscourseAPI(bodyUrl);
+    topic.post_body_html = post_stream.posts[0].cooked;
+    return topic;
+  });
+  blog.topics = await Promise.all(topicsPromises);
+  const now = Date.now();
+  metrics.set(blog, 'updated', now);
+  metrics.push(blog, 'pull-time', [now, Date.now() - time]);
```
jankeromnes: Nit: I think you mean `now - time` instead of `Date.now() - time`.

lib/routes.js
```diff
@@ -149,6 +149,33 @@ exports.blogPage = function (response, user = null) {
   ]);
 };
 
+// Public blog page.
+const blogSectionNew = camp.template('./templates/blog-new.html');
+exports.blogPageNew = function (response, user, blog) {
+  const title = 'Blog';
+  const { topics } = blog.getDb();
+
+  const contents = topics.map(topic => {
```
jankeromnes: Nit: s/contents/posts/?

templates/blog-new.html
```diff
@@ -0,0 +1,12 @@
+    <section class="blog">{{
+        for (const topic of contents) {
```
jankeromnes: Nit: Is it plain wrong to call these variables `for (const post of posts)`?

lib/blog.js
notriddle: [We're certain](https://nodejs.org/api/stream.html#stream_writable_destroy_error).

Issue #183, opened 2017-10-29
Provide a default db.json in README.md so that no errors comes up on setting janitor locally.


arshdkhn1:
Running `node app` and other commands on fresh install give this error -
```
[2017-10-29T16:04:16.981Z] [fail] letsencrypt { Error: getaddrinfo EAI_AGAIN acme-v01.api.letsencrypt.org:443
    at Object.exports._errnoException (util.js:1020:11)
    at errnoException (dns.js:33:15)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:76:26)
  code: 'EAI_AGAIN',
  errno: 'EAI_AGAIN',
  syscall: 'getaddrinfo',
  hostname: 'acme-v01.api.letsencrypt.org',
  host: 'acme-v01.api.letsencrypt.org',
  port: 443 }
```
There are no instructions given on README.md to solve this.

arshdkhn1:
db.json should have below configurations by default - 
```js
  "security": {
    "forceHttp": true,
    "forceInsecure": true
  }
```
and other configurations for sign in.

jankeromnes:
Indeed, this is a problem. I really like the idea of mentioning this in the readme. Also you're suggesting great defaults for db.json, in fact they're already in use when you develop Janitor on Janitor, see the default db.json in dockerfiles https://github.com/JanitorTechnology/dockerfiles/blob/master/janitor/db.json

jankeromnes:
This was fixed in 05bc614ae74412865666dd9755faec448c539704

Thanks @KiLlOrBe!

PR #209, opened 2017-12-09
Add instruction for local installation

KiLlOrBe:
Hi,
I have added an instruction for local installation about `db.json`.
Issue https://github.com/JanitorTechnology/janitor/issues/183

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,6 @@ Sign in to [janitor.technology](https://janitor.technology).
 
 Install [Node.js](https://nodejs.org) (and optionally [Docker](https://www.docker.com)).
 
-You will need to edit `/db.json` with the following [configuration](https://github.com/JanitorTechnology/dockerfiles/blob/master/janitor/db.json).
-
 Clone this repository:
 
     git clone https://github.com/janitortechnology/janitor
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,6 +18,8 @@ Sign in to [janitor.technology](https://janitor.technology).
 
 Install [Node.js](https://nodejs.org) (and optionally [Docker](https://www.docker.com)).
 
+You will need to edit `/db.json` with the following [configuration](https://github.com/JanitorTechnology/dockerfiles/blob/master/janitor/db.json).
+
```
jankeromnes: Nit: Please move this instruction down below (I find it a bit confusing here, because at this point you don't have a `janitor/` folder to put it in).

Maybe just before `node app`? E.g.

```
Configure a local `db.json` (or simply download this example)
```

With `this example` being a link to the *raw* `db.json` example.

KiLlOrBe: Ok thanks, I will change that!

jankeromnes:
Thank you @KiLlOrBe for updating our `README.md` instructions! And @nt1m for approving them! 😄

PR #211, opened 2017-12-15
Add Photon-style links

notriddle:


static/css/janitor-new.css
```diff
@@ -71,6 +71,31 @@ main {
   flex-grow: 1;
 }
 
+a {
+  box-shadow: 0 0 2px 0 transparent;
```
nt1m: nit: Please remove this transparent box-shadow, it doesn't seem to do anything.

static/css/janitor-new.css
```diff
@@ -71,6 +71,31 @@ main {
...
+  color: #0060df;
+}
+
+a:active {
+  color: #003eaa;
+}
+
+a:focus {
+  box-shadow: 0 0 2px 0 #36cae4, 0 0 4px 0 rgba(54, 202, 228, 0.3);
+  border-radius: 4px;
```
nt1m: Can you please use these styles instead ?

```css
box-shadow: 0 0 0 1px #0a84ff, 0 0 0 4px rgba(10, 132, 255, 0.2);
border-radius: 2px;
```

static/css/janitor-new.css
```diff
@@ -89,14 +113,18 @@ main {
 
 .btn:hover {
   background: #d7d7db;
+  color: #0c0c0d;
```
nt1m: nit: Please avoid repeating the text color for the hover/active/focused states here and below

templates/design.html
```diff
@@ -14,6 +14,10 @@ <h2>Icons</h2>
 <i class="icon edit"></i>
 <i class="icon menu"></i>
 
+<h2>Links</h2>
+Compare <a href="">regular</a> and <a href="" class="btn">button</a> links.
```
nt1m: I don't think we need to document button links for now, especially since there are some rough edges around them.

static/css/janitor-new.css
notriddle: The problem is that `a:hover` has a higher specificity than `.btn`, so the text in link buttons turns blue when I mouse over it.

jankeromnes: I believe that @nt1m meant that you could factor the `:hover`, `:active` and `:focused` color rules in order to only specify the text color value once.

notriddle:
To avoid a lot of this nastiness, I made the link styling blog-specific.

PR #222, opened 2018-01-10
Fix GitHub SSH public keys in ~/.ssh/authorized_keys.

jankeromnes:
If you enabled the GitHub integration, you probably have a few `[Object object]` in your containers' `~/.ssh/authorized_keys` files.

This is because SSH public keys imported from GitHub look like `{ id: 1, key: "ssh-rsa AAA..." }` instead of just `"ssh-rsa AAA..."`.


jankeromnes:
Thanks!

PR #226, opened 2018-01-11
Swap body/heading grey background in new design panel.

jankeromnes:
I find it more logical and cleaner to have a grey panel heading and white panel body, as opposed to the current white heading and grey body.

@nt1m please have a look when you have a minute. 😄


nt1m:
@jankeromnes When the gray color is on top, it kinda fades with the top shadow, whereas at the bottom, the shadow is stronger, so the problem isn't there.

jankeromnes:
:sad:

PR #236, opened 2018-01-19
Allow scheduling system events persistently.

jankeromnes:
This is basically a re-work of https://github.com/JanitorTechnology/janitor/pull/140/ (sorry for the duplicate work!), which will allow us to schedule potentially long-term system events like "invalidate this token in 24 hours", "delete this container in 1 year", etc.

Here are the differences between this implementation and the "tasks" approach:

- I felt that using an `EventEmitter` was better-suited for registering + triggering callbacks on specific task/event types.
- I also regularly process upcoming tasks/events, but if they're due before the next loop, I use `setTimeout` to trigger them just in time.
- My approach doesn't need a unique task/event IDs.
- It supports multiple listeners for the same event type.
- It doesn't check for errors from the listeners (just if at least one listener was called for this event).

### DIFF
--- a/api/admin-api.js
+++ b/api/admin-api.js
@@ -91,16 +91,6 @@ azureAPI.get('/virtualmachines', {
 const oauth2providersAPI = adminAPI.api('/oauth2providers', {
   beforeEachTest: next => {
     const providers = db.get('oauth2providers');
-    if (!providers.github) {
-      // FIXME Remove this if block when the GitHub pull request is merged:
-      // https://github.com/JanitorTechnology/janitor/pull/80
-      providers.github = {
-        id: '',
-        secret: '',
-        hostname: 'github.com',
-        api: 'api.github.com'
-      };
-    }
     providers.github.id = '1234';
     providers.github.secret = '123456';
     next();
--- a/api/admin-api.js
+++ b/api/admin-api.js
@@ -6,6 +6,7 @@ const selfapi = require('selfapi');
 
 const azure = require('../lib/azure');
 const db = require('../lib/db');
+const events = require('../lib/events');
 const log = require('../lib/log');
 const users = require('../lib/users');
 
@@ -85,6 +86,55 @@ azureAPI.get('/virtualmachines', {
   },
 
   examples: [],
+});
+
+// API sub-resource to manage scheduled events.
+const eventsAPI = adminAPI.api('/events');
+
+eventsAPI.get({
+  title: 'List past system events',
+
+  handler: (request, response) => {
+    const { user } = request;
+    if (!user || !users.isAdmin(user)) {
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' }, null, 2);
+      return;
+    }
+
+    response.json(events.get(), null, 2);
+  },
+
+  examples: [{
+    response: {
+      body: json => {
+        try { return Array.isArray(JSON.parse(json)); } catch (error) { return false; }
+      }
+    }
+  }]
+});
+
+eventsAPI.get('/queue', {
+  title: 'List upcoming system events',
+
+  handler: (request, response) => {
+    const { user } = request;
+    if (!user || !users.isAdmin(user)) {
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' }, null, 2);
+      return;
+    }
+
+    response.json(events.getQueue(), null, 2);
+  },
+
+  examples: [{
+    response: {
+      body: json => {
+        try { return Array.isArray(JSON.parse(json)); } catch (error) { return false; }
+      }
+    }
+  }]
 });
 
 // API sub-resource to manage OAuth2 providers.
--- a/lib/events.js
+++ b/lib/events.js
@@ -1,0 +1,150 @@
+// Copyright © 2018 Jan Keromnes.
+// The following code is covered by the AGPL-3.0 license.
+
+const EventEmitter = require('events');
+
+// Note: If you miss an important scheduling feature here, maybe import a "real"
+//   cron-based scheduler module instead of reimplementing one.
+// For example: https://github.com/IMA-WorldHealth/TaskList
+
+const db = require('./db');
+const log = require('./log');
+
+const schedulingInterval = 1000 * 60 * 60 * 10; // Every 10 hours.
+
+const emitter = new EventEmitter();
+const events = db.get('events');
+load();
+
+// Load past and upcoming events from the database.
+function load () {
+  emitter.on('error', error => {
+    log('[fail] event emitter error', error);
+  });
+
+  if (!events.history) {
+    events.history = [];
+  }
+
+  if (!events.queue) {
+    events.queue = [];
+  }
+
+  // Events should flow through the following states:
+  // 1. Queued: `event` is added to `events.queue`
+  // 2. Scheduled: `event.scheduledTime` is set
+  // 3. Emitted: `event.emittedTime` is set, and `event` is moved from
+  //   `events.queue` to `events.history`
+
+  // Remove any empty event slots (already emitted).
+  events.queue = events.queue.filter(event => !!event);
+
+  // Re-schedule any un-emitted events.
+  events.queue.forEach(event => { event.scheduledTime = null; });
+}
+
+// Start regularly scheduling events.
+exports.startScheduling = function () {
+  setTimeout(processQueue, 0);
+};
+
+// Get all previously emitted events.
+exports.get = function () {
+  return events.history;
+};
+
+// Get the queue of upcoming events.
+exports.getQueue = function () {
+  return events.queue
+    .filter(event => !!event) // Remove any empty event slots (already emitted).
+    .sort((a, b) => b.dueTime - a.dueTime); // Sort the queue by due date.
+};
+
+// Register a new event listener.
+exports.on = function (eventType, listener) {
+  emitter.on(eventType, listener);
+};
+
+// Register a single-use event listener.
+exports.once = function (eventType, listener) {
+  emitter.once(eventType, listener);
+};
+
+// Emit an event now.
+exports.emit = function (eventType, payload = null) {
+  exports.emitAtTime(eventType, Date.now(), payload);
+};
+
+// Emit an event at a certain due date (a timestamp).
+exports.emitAtTime = function (eventType, dueTime, payload = null) {
+  const event = createEvent(eventType, dueTime, payload);
+  events.queue.push(event);
+  maybeSchedule(event);
+};
+
+// Process all upcoming events, and schedule any events that are due soon.
+// Note: We schedule events at regular intervals instead of just using
+//   `setTimeout` directly, because `setTimeout` has a maximum range of about
+//   24 days, which is not enough for our needs.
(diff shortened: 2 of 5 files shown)


jankeromnes:
@nt1m and @bnjbvr could you please take a look at this approach? :)

lib/events.js
```diff
@@ -0,0 +1,137 @@
+// Copyright © 2018 Jan Keromnes. All rights reserved.
```
bnjbvr: nit: All rights reserved is unnecessary in modern times, since this is [the default](https://opensource.stackexchange.com/questions/2121/mit-license-and-all-rights-reserved)

lib/events.js
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  // Remove any empty event slots (already emitted).
+  events.queue = events.queue.filter(event => !!event);
+
+  // Re-schedule any un-emitted events.
+  events.queue.forEach(event => { event.scheduled = null; });
+
+  // Start regularly scheduling events, once start-up is complete.
+  setTimeout(processQueue, schedulingStartDelay);
```
bnjbvr: How did you determine that 1 minute was enough? Wouldn't it be clearer that there is an exported function `startScheduling` in this module that actually programs the first run, and when start-up has completed in the main module, it calls `startScheduling` instead?

lib/events.js
```diff
@@ -0,0 +1,137 @@
...
+function maybeSchedule (event) {
+  if (!event) {
+    // Already emitted, nothing to do.
+    return;
+  }
+
+  const now = Date.now();
+  if (event.time && event.time > now) {
+    // The event is not due yet, emit it later.
+    if (event.time < now + schedulingInterval && !event.scheduled) {
```
bnjbvr: nit: use a non-overflowing comparison: `event.time - now < schedulingInterval`

bnjbvr: (I also find the intent clearer by using this comparison)

lib/events.js
```diff
@@ -0,0 +1,137 @@
...
+      // Emit it soon, before the next queue processing takes place.
+      setTimeout(() => { emitEvent(event); }, event.time - now);
+      event.scheduled = now;
+    }
+    return;
+  }
+
+  // The event is due or past due, emit it now.
+  emitEvent(event);
```
bnjbvr: `setTimeout(() => { emitEvent(event); }, 0);` ?

api/admin-api.js
```diff
@@ -13,6 +14,55 @@ const adminAPI = module.exports = selfapi({
...
+
+  handler: (request, response) => {
+    const { user } = request;
+    if (!user || !users.isAdmin(user)) {
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' }, null, 2);
+      return;
+    }
+
+    response.json(events.getQueue(), null, 2);
```
bnjbvr: It's fine as it is now, but it could be an easy way to DDOS the server later (when there are many events), thus a paginated API would probably be better here.

lib/events.js
```diff
@@ -0,0 +1,137 @@
...
+// See: https://nodejs.org/api/timers.html#timers_settimeout_callback_delay_args
+function processQueue () {
+  log('processing event queue');
+  events.queue.forEach(maybeSchedule);
+  setTimeout(processQueue, schedulingInterval);
+}
+
+// Emit or schedule an event if it's due soon.
+function maybeSchedule (event) {
+  if (!event) {
```
bnjbvr: I think you can simplify this whole function by having `if (!event || event.scheduled)` here, and then below you can coalesce the two `if`.

lib/events.js
bnjbvr: Also, this will reschedule events that have been scheduled but have no due time :)

lib/events.js
```diff
@@ -0,0 +1,137 @@
...
+
+  if (!events.queue) {
+    events.queue = [];
+  }
+
+  // Remove any empty event slots (already emitted).
+  events.queue = events.queue.filter(event => !!event);
+
+  // Re-schedule any un-emitted events.
+  events.queue.forEach(event => { event.scheduled = null; });
```
bnjbvr: Maybe assert that `event.emitted` is not set here? I think this can't happen (because the db saving happens at the end of the function that sets `emitted`, which  means that if the function gets interrupted, then the event will be scheduled twice)

lib/events.js
```diff
@@ -0,0 +1,137 @@
...
+
+  events.history.push(event);
+  db.save();
+}
+
+// Create a new event to be emitted at a given timestamp.
+function createEvent (type, time, payload = null) {
+  return {
+    type,
+    time,
```
bnjbvr: I like all the attribute names you've used, but this one. How about `due_time` or `due_date` or `due_at`? (or in camelCase)

lib/events.js
```diff
@@ -0,0 +1,137 @@
...
+    payload,
+    scheduled: null,
```
bnjbvr: `scheduledAt`, to make it clear it's a date?

lib/events.js
```diff
@@ -0,0 +1,137 @@
...
+    emitted: null,
```
bnjbvr: `emittedAt`?

tests/tests.js
```diff
@@ -187,6 +187,62 @@ tests.push({
...
+      if (received) {
+        console.error(new Error('Received duplicate TestEvent at ' + delta));
+        return;
+      }
+      received = true;
+      if (tooLate) {
+        console.error(new Error('Received TestEvent too late at ' + delta));
+        return;
+      }
+      if (timeDelta <= tooEarlyDelta - expectedDelta) {
```
bnjbvr: I think if `timeDelta === 1`, then it should be flagged as too early but this condition won't catch it. I think you want `Math.abs()` on both sides.

lib/events.js
```diff
@@ -0,0 +1,137 @@
...
+  emitter.on('error', error => {
+    log('[fail] event emitter error', error);
+  });
+
+  if (!events.history) {
+    events.history = [];
+  }
+
+  if (!events.queue) {
+    events.queue = [];
```
bnjbvr: Can you comment, maybe in a block at the top of this file, what are the expected states of an event:

`in queue => scheduled => emitted`

tests/tests.js
```diff
@@ -187,6 +187,62 @@ tests.push({
   }
 });
 
+tests.push({
```
bnjbvr: Can you add a test for an event without a due date?

jankeromnes:
Thanks a lot for the review! 😄 I'll rebase these commits and address your comments/nits.

lib/events.js
jankeromnes: Yay for modern times! Removed the "All rights reserved." part, because it's both unnecessary and misleading.

lib/events.js
jankeromnes: All events have a due time. 😉

jankeromnes: I don't see the point of `setTimeout(() => { emitEvent(event); }, 0);` as opposed to `emitEvent(event);` but fair enough, I'll do that. 😄

api/admin-api.js
jankeromnes: Agreed, I do believe that we'll need to paginate our APIs at some point.

lib/events.js
jankeromnes: I tweaked the function following your suggestions. Hopefully you'll like how it looks now. 😄

lib/events.js
jankeromnes: Agreed, I don't think this can happen. Also, we don't use asserts in the code today, and I'm not really sure how to do that (e.g. What method should we use? Should we assert in production? Etc.)

lib/events.js
jankeromnes: Fair enough, I went with `dueTime` to be very explicit.

lib/events.js
jankeromnes: I went for `scheduledTime` to be consistent with `dueTime`.

lib/events.js
jankeromnes: Ok for `emittedTime`.

tests/tests.js
jankeromnes: `tooEarlyDelta` and `expectedDelta` are constant values (currently `1000 * 2` and `1000 * 5`, respectively) so the right-hand value is `-3000` (it should always be negative).

`timeDelta` is `Date.now() - expectedTime`, so ideally it's close to `0` if the event is triggered on time, but it's negative when the event is received earlier and positive when it's received later.

So if the event is 1 millisecond late, I don't think that it should be flagged as too early. (Currently, we flag events "too early" when they are more than 3 seconds early, and we flag them "too late" when we receive them more than 3 seconds late.)

tests/tests.js
jankeromnes: Events always have a due date. We could add a test for `exports.emit = function (eventType, payload = null) { exports.emitAtTime(eventType, Date.now(), payload); };` but given that it's just a simple wrapper, I don't see the value in testing it.

notriddle:
I just found out about [this "TaskList" library](https://github.com/IMA-WorldHealth/TaskList), if you're interested in outsourcing some of the complexity of this time ring implementation?

jankeromnes:
Addressed all nits and replied to all questions, except the nit about starting scheduling via an exported method instead of after an arbitrary 1min delay.

Also, I should look into notriddle's suggestion as an alternative to this, as it sounds promising.

lib/events.js
jankeromnes: Sure! Done. 😄

jankeromnes:
> Also, I should look into notriddle's suggestion as an alternative to this, as it sounds promising.

Well, since this code is already written, maybe let's merge it as is without letting it bitrot any further.

TaskList looks great though, and its cron schedules look easy enough to persist and reload. It doesn't seem to allow irregular schedules (e.g. "every 30 hours", although that's not a strong requirement), and it has more than zero NPM dependencies (but not a huge problem either).

I should probably add a comment so that if we need more advanced scheduling functions, we switch to a real cron-based module instead of implementing and maintaining everything ourselves.

lib/events.js
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  // Events should flow through the following states:
+  // 1. Queued: `event` is added to `events.queue`
+  // 2. Scheduled: `event.scheduledTime` is set
+  // 3. Emitted: `event.emittedTime` is set, and `event` is moved from
+  //   `events.queue` to `events.history`
```
bnjbvr: (I might be nitpicky here) Maybe this description is incomplete? It seems the original entry in the `queue` array is first set to `null`, then filtered out/deleted.

tests/tests.js
bnjbvr: It's just conceptually weird that things that are after `expectedDelta` may not be always considered late. Reproducibility is hard, I admit it. Despite this, I think the concept of being late should be binary and not a blurry notion. That being said, it doesn't matter too much :)

lib/events.js
jankeromnes: Hmm, you're right, but I see this as an implementation detail of the word `move`. Technically the event representation does get moved to a different collection, even if it leaves behind a temporary empty slot. And that particular behavior is already explained by another comment. But thanks for nitpicking! :smile:

jankeromnes:
Woohoo! Thanks a lot! :tada:

PR #237, opened 2018-01-20
New login page

arshdkhn1:
#### Desktop view
localhost_1443_login-new

#### Tab view
localhost_1443_login-new ipad

#### Mobile view
localhost_1443_login-new iphone x

### DIFF
--- a/static/css/janitor-new.css
+++ b/static/css/janitor-new.css
@@ -20,6 +20,8 @@ body {
   min-height: 100%;
 }
 
+/* Helper classes */
+
 *:not(:hover) > .show-on-hover {
   visibility: hidden;
 }
@@ -45,16 +47,16 @@ body {
   padding: 0.75em;
 }
 
-.navbar-container,
-.container {
-  max-width: 992px;
-  margin: auto;
-  width: 100%;
+.disabled {
+  cursor: not-allowed;
 }
 
-.container {
-  padding: 0 20px;
+.btn.large,
+.form-control.large {
+  padding: 1em;
 }
+
+/* Typography */
 
 h1 {
   font-weight: 200;
@@ -78,26 +80,39 @@ h4 {
   font-size: 1em;
 }
 
-main {
-  overflow: auto;
-  flex-grow: 1;
-  /* viewport height - header height - 15px extra height (so that footer is visible) */
-  min-height: calc(100vh - 95px);
-  display: flex;
-  flex-direction: column;
-}
-
 p {
   color: #666;
   font-weight: 400;
+}
+
+p.description {
+  color: #0b0b0d;
 }
 
 .text-center {
   text-align: center;
 }
 
-.disabled {
-  cursor: not-allowed;
+/* Layout */
+
+.navbar-container,
+.container {
+  max-width: 992px;
+  margin: auto;
+  width: 100%;
+}
+
+.container {
+  padding: 40px 20px;
+}
+
+main {
+  overflow: auto;
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+  /* viewport height - header height - 20px extra height (so that footer is visible) */
+  min-height: calc(100vh - 100px);
 }
 
 .centered-container {
@@ -135,16 +150,16 @@ p {
 }
 
 .btn-primary {
-  background: #0060df;
+  background: #0c0c0d;
   color: #f9f9fa;
 }
 
 .btn-primary:hover {
-  background: #003eaa;
+  background: #38383d;
 }
 
 .btn-primary:hover:active {
-  background: #002275;
+  background: #2a2a2e;
 }
 
 .btn-error {
@@ -161,6 +176,13 @@ p {
 }
 
 /* Inputs */
+
+.form-control-container > .form-control,
+.form-control-container > .btn {
+  width: 100%;
+  margin: 0;
+  margin-top: 0.75em;
+}
 
 .form-control {
   border: 1px solid rgba(12, 12, 13,0.3);
@@ -277,7 +299,6 @@ p {
 .footer {
   border-top: solid 1px #eee;
   background: #fafafa;
-  padding: 40px 0;
 }
 
 .footer .footer-inner-container {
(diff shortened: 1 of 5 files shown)


static/css/janitor-new.css
```diff
@@ -4,6 +4,27 @@
...
+}
+
+:-ms-input-placeholder { /* IE 10+ */
+  opacity: 1;
+  color: #999;
+}
+
+:-moz-placeholder { /* Firefox 18- */
+  opacity: 1;
+  color: #999;
```
jankeromnes: Nit: If this is a polyfill for the `placeholder` attribute, I don't feel like it's worth having. It's ok to just use `placeholder` attribute and don't worry about very old browsers.

static/css/janitor-new.css
```diff
@@ -48,11 +68,13 @@ body {
 h1 {
   font-weight: 200;
   font-size: 33px;
+  margin: 30px 0;
```
jankeromnes: Nit: Wow, that's a lot of margin. Are we sure we want this on all `h1`'s throughout the website? I'd love to have @nt1m's opinion on this, and on the other changes to the new design CSS.

static/css/janitor-new.css
```diff
@@ -13,8 +34,7 @@ html {
 body {
   margin: 0;
   padding: 0;
-  /* The list of system fonts after "Droid Sans" are from github.com */
-  font-family: "Fira Sans", Roboto, "Droid Sans", -apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
+  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
```
jankeromnes: Nit: Was this font removal intentional?

static/css/janitor-new.css
```diff
@@ -71,8 +93,58 @@ main {
...
+  font-family: inherit;
+}
+
+button:focus, input:focus {
+  outline: none;
+}
+
 /* Buttons */
 
+.button {
```
jankeromnes: Nit: It feels weird to have two different button classes, `.btn` and `.button`. Can't we unify them somehow?

static/css/janitor-new.css
```diff
@@ -408,6 +676,7 @@ main {
   padding: 10px;
   margin: 1.5em 0;
   border-radius: 4px;
+  display: none;
```
jankeromnes: Why?

templates/header-new.html
```diff
@@ -5,27 +5,29 @@
     <meta name="application-name" content="Janitor">
     <meta name="theme-color" content="#ffffff">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <title> {{= title in html}}{{if title then {{ • }}}}Janitor </title>
+    <title>Janitor{{if title then {{ – }}}}{{= title in html}}</title>
```
jankeromnes: Nit: As a personal preference, I prefer having the section name before "Janitor", because that makes smaller browser tabs still explicit about what they hold (e.g. it's confusing to have three tabs called "Janitor -..." while it's better to have "Projects -...", "Contribut..." and "Public Data...").

templates/login-new.html
```diff
@@ -0,0 +1,19 @@
+    <section class="login">
+      <div class="container text-center">
+        <h2>
+          Authenticate
+        </h2>
+        <p class="description">
+          To sign up or log in, fill in your email address below:
```
jankeromnes: Nit: As in previous comment, please use the word "please".

templates/login-new.html
```diff
@@ -0,0 +1,19 @@
...
+        </p>
+        <form class="ajax-form" id="login-form">
+          <div class="input-group">
+            <input class="text-input" name="email" placeholder="you@yourdomain.com" spellcheck="false" autocorrect="off" autocapitalize="off" type="email" autocomplete="off" required>
+            <button class="button dark"><b>Submit</b></button>
+          </div>
+        </form>
+        <div class="login-with">
+          <span class="disabled" title="Coming soon">Use GitHub instead</span>
```
jankeromnes: Nit: As in previous comment, maybe "Sign in with GitHub" is clearer / more explicit?

static/css/janitor-new.css
arshdkhn1: @jankeromnes for desktop view its ok. I ll change it to smaller val in media queries.

static/css/janitor-new.css
arshdkhn1: I was experimenting, I forgot to remove it. Sorry.

nt1m:
@arshdkhn1 I simplified the login page code to only have shared components (no page specific ones):

screen shot 2018-02-01 at 9 53 58 am

I'd be open with changing the design of shared components, but I really don't want us to have special page CSS everywhere.

Can you please review it ?

static/css/janitor-new.css
```diff
@@ -78,26 +88,37 @@ h4 {
   font-size: 1em;
 }
 
-main {
```
arshdkhn1: why did u remove it?

nt1m:
@arshdkhn1 Could you please take a second look at this ? I made the button black, and restored the min-height (also tweaked it a bit, because I felt like the scrollable area felt too large to me).

nt1m:
I didn't tweak the input style, because it's a more sensible area (it affects more pages), and this PR already has enough changes on its own :) I'm comfortable if we do a post-review of the design (eg.  have all pages converted, then tweak later the design). I feel like this would make us iterate faster instead of stepping on each others feet with conflicting page specific CSS :)

@arshdkhn1 What do you think ?

arshdkhn1:
> I didn't tweak the input style, because it's a more sensible area (it affects more pages), and this PR already has enough changes on its own :) I'm comfortable if we do a post-review of the design (eg. have all pages converted, then tweak later the design). I feel like this would make us iterate faster instead of stepping on each others feet with conflicting page specific CSS :)

cool, sounds good :)

I have tweaked the page a little bit. I felt like footer was showing too mch. I also added some padding to .centered-container class so that on screens having small vh, .centered container doesn't get stick to header and footer.

PR #239, opened 2018-01-25
Add timeago server-side and use it to pre-render live-updating fuzzy timestamps.

jankeromnes:
Drive-by changes:
- Update container title anchors to 16-char container IDs
- s/Updated/Built/ and s/build-time/pull-time/ in hidden project page

### DIFF
--- a/templates/containers-new.html
+++ b/templates/containers-new.html
@@ -20,7 +20,7 @@
                 <input class="form-control" data-submit-on="blur" name="/name" placeholder="{{= project.name in xmlattr }} #{{= id in id}}" type="text" value="{{= machine.properties.name in xmlattr }}">
               </form>
               <div class="editable-value">
-                <h4 id={{= machine.docker.container in json}}>{{= name in html}}</h4>
+                <h4 id="{{= machine.docker.container.slice(0,16) in xmlattr}}">{{= name in html}}</h4>
                 <span class="icon edit editable-toggle"></span>
               </div>
             </div>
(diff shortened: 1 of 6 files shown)


templates/containers-new.html
```diff
@@ -48,7 +48,11 @@ <h4 id={{= machine.docker.container in json}}>{{= name in html}}</h4>
           </ul>
           <div class="tab-panels">
             <div class="tab-panel padded" data-tab="info">
-              <p class="project-updated">Built <time class="project-timestamp" data-timestamp="{{= machine.data.updated in integer}}"></time>.</p>
+              <p class="project-updated">Built {{
+                const machineUpdatedFuzzy = machine.data.updated
+                  ? timeago().format(machine.data.updated)
+                  : 'never';
+              }}<time class="project-timestamp" data-timestamp="{{= machine.data.updated in integer}}">{{= machineUpdatedFuzzy in html}}</time>.</p>
```
arshdkhn1: if `machine.data.updated` is falsy then will the text be `Built never.`?

jankeromnes: Correct. This happens when we add a new project, but its Docker image wasn't pulled in yet. Today, we show `Built .`, so I think `Built never.` is an improvement.

(Additionally, you can hover on "never" and the title will say "Invalid Date" to make it very clear what happened.)

jankeromnes:
Thanks for approving! 👍 

> Can we remove the client timeago files now that everything is done server-side ?

I think we'll want to keep the client-side timeago file, because it takes over the fuzzy timestamp and keeps it up-to-date in real time (e.g. if you have something that says "a few seconds ago", it will soon automatically change to "one minute ago").

This will also apply to container expiration, once it's in place (you'll be able to see "Expires in one minute." and then "in a few seconds." without needing to refresh the page).

PR #240, opened 2018-01-25
Remove third-party c9.io integration because we no longer use it.

jankeromnes:
A little Thursday-morning clean-up.

### DIFF
--- a/app.js
+++ b/app.js
@@ -572,32 +572,21 @@ boot.executeInParallel([
     }
 
     let key = '';
-    let match;
-    switch (data.name) {
-      case 'cloud9':
-        // Extract a valid SSH public key from the user's input.
-        // Regex adapted from https://gist.github.com/paranoiq/1932126.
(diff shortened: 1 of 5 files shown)


nt1m:
Wait, would the c9.io links from the containers page still work ?

jankeromnes:
> Wait, would the c9.io links from the containers page still work ?

Yes.

The full story is that `user.keys.c9user` never actually did anything. I simply checked these fields occasionally, to manually verify that every Janitor user was also a member of our c9.io Team Account, in order for them to be able to create c9.io SSH workspaces and access them. And now that we don't use c9.io workspaces anymore, we can drop this field.

Old containers with c9.io links will continue to work as before, but we should eventually encourage everyone to migrate any unsaved work into newer containers, ensure we'll get rid of very old containers.

nt1m:
@jankeromnes Works for me then!

PR #242, opened 2018-01-25
New design tweaks

nt1m:
- Fixed the container page panel shadows
- remove unnecessary uses of flexbox
- remove the `gray-text` class
- Fixed blog page main container
- Slightly reduce the header height (I think it was a bit too large)

### DIFF
--- a/static/css/janitor-new.css
+++ b/static/css/janitor-new.css
@@ -8,13 +8,13 @@ html {
   height: 100%;
   background-color: #fff;
   color: #0b0b0d;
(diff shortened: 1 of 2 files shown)


arshdkhn1:
@nt1m 
screencapture-localhost-1443-design-1516901046099

Nit 1: In tab view, there is no padding on #brand-log. I think #brand-logo should have top-bottom padding of 20px 

Nit2: the footer brand-logo and tagline is not arranged properly

PR #243, opened 2018-01-25
Remove terminal tab from containers page until it actually works

nt1m:


nt1m:
@jankeromnes Thanks for the review! The terminal is very easy to add back if needed, but atm, there are a few things that need fixing: the terminal itself, the interaction with the proxy, lazy loading the terminal instead of loading it right away.

jankeromnes:
> The terminal is very easy to add back if needed, but atm, there are a few things that need fixing: the terminal itself, the interaction with the proxy, lazy loading the terminal instead of loading it right away.

Agreed, and thanks for the details! I suppose that moving to a LetsEncrypt-wildcard-based domain-name-proxy (*phew*) will help fixing the proxy issue.

PR #244, opened 2018-01-25
Move project descriptions to the project-specific page template. 1/2

jankeromnes:
Because otherwise it clutters the projects list on the landing and projects pages.


jankeromnes:
Uh, let's do that after #239 to avoid an unnecessary Git conflict.

PR #245, opened 2018-01-25
Move project descriptions to the project-specific page template.

jankeromnes:
Because otherwise it clutters the projects list on the landing and projects pages. (Was #244.)


jankeromnes:
Woohoo! Thanks!

PR #246, opened 2018-01-25
Fix new design issues reported by @arshdkhn1

nt1m:


static/css/janitor-new.css
```diff
@@ -243,13 +244,17 @@ p {
...
+}
+
 .navbar ul {
   text-align: right;
   flex: 1;
 }
 
 .navbar ul li a {
-  padding: 2.5em 1em;
+  padding: 0 1em;
```
arshdkhn1: I prefer pading-lr of 20px.. In media queries(line 330) u wll see that `ul li a` have padding-lr of 20px so in mobile/tab view atm the li and #brand-log is not aligned horzontaly

arshdkhn1: I would also suggest you to get rid  of .container  on header because due to that reason ur point of fit's law is not workin... since #brand-log and menu btn 's parent wrapper has padding of 1em on right/left ... click on this area will not trigger click on #brand-logo and menubtn respectively

arshdkhn1: border-bottom of li in mobile view also get corpped  due to parent's padding and I dont like that

static/css/janitor-new.css
```diff
@@ -45,11 +45,15 @@ body {
   padding: 0.75em;
 }
 
+.navbar-container,
 .container {
   max-width: 900px;
```
arshdkhn1: just make it 992 as media screen width has been changed to 992

static/css/janitor-new.css
```diff
@@ -243,17 +248,18 @@ p {
...
+  padding-left: 20px;
+}
+
 .navbar ul {
   text-align: right;
   flex: 1;
 }
 
 .navbar ul li a {
-  padding: 2.5em 1em;
```
arshdkhn1: could u please give a tags top bottom padding?

static/css/janitor-new.css
```diff
@@ -243,17 +248,18 @@ p {
...
-}
-
-.navbar ul li:last-child a {
-  padding-right: 0;
```
arshdkhn1: idid this intentionaly to intentionaly align horizontally the last link with main. With 20px padding on right last link(in desktop view) is 20px left to main's content.. This doesn't make much difference. Somepeople dont worry about it.

PR #247, opened 2018-01-25
New settings page

nt1m:


nt1m:
screen shot 2018-01-25 at 20 03 41-fullpage

screen shot 2018-01-25 at 20 03 59-fullpage

nt1m:
> i dont like the focus of button and input

This was never changed in this PR. I'm happy to tweak it in a separate PR though. How do you like Github's focus outline for inputs ?

> clickin on input checkbox of input notification - some time two check marks on right appears

I can't reproduce this, do you have a screenshot ?

> on config tab, due to scrollbar textarea seems to less padded on right. On mac this might now be issue as mac's scroll bar dissappear. It myt sound like a small issue but if you insist on making consitent padding for elements having scrollbar, you ll soon realize it is a bit of hassle. Scrollbar width is differnt on diff brwser.

I've noticed as well, it's because `width: 100%` isn't perfect. I'll see what I can do here.

> tabs individually have no dividin line.. we need a ui that clearly shows that thse are tab btns(accont and config)

What do you mean by dividing line ? This is the same tab design as the containers page.

> animation/transition on info popup Scsful deployed to ... is a bit slow .. I like anim whose end part are not that slow

The info popup is generated by the browser (using setCustomValidity), the CSS has no control over it.

> reset button doesn't have pointer cursor on hover

I'll tweak that in another PR

arshdkhn1:
image
The position of tick mark needs to be adjusted. On Account page as well, the tick mark is a bit down incomparison to the text on left.

arshdkhn1:
> This was never changed in this PR. I'm happy to tweak it in a separate PR though. How do you like Github's focus outline for inputs ?

its good, i ll let u know some other options as well

> What do you mean by dividing line ? This is the same tab design as the containers page.

yes it is same but i realized it now.
image

In above pic u can see there are two text but there is not enough UI to tell user that thse are tabs..

> I can't reproduce this, do you have a screenshot ?

first click on Name input, then with focussed name input click on checkbox

Extra nits:

1. i feel like we can fit the email and name in one row and split them in tab/mob view

2. make paragraph and connect btn vertically center(Not a strong nit i accept)
image

3.  remove margin-left on first tab button

4. Is warning-alert background right for **Note:...**

nt1m:
> i feel like we can fit the email and name in one row and split them in tab/mob view

I actually like keeping them in separate rows, kinda like how the Github settings page is designed.

> In above pic u can see there are two text but there is not enough UI to tell user that thse are tabs..

I'd like those 2 tabs to be a side menu, but that's a bit of a pain to implement :P 
I don't have any ideas on how to improve the tab design though, let me know if you have any :)

> remove margin-left on first tab button

I agree, note that this is still a work in progress.

> Is warning-alert background right for Note:...

This is a leftover from the old design. I'm fine with removing it though.

nt1m:
Design has been updated:
screen shot 2018-01-25 at 9 48 56 pm
screen shot 2018-01-25 at 9 49 52 pm

jankeromnes:
Woo! This looks amazing 😄 thanks a lot @nt1m for the outstanding work and @arshdkhn1 for the excellent reviews! 💯 🥇 👍 

I'll have a look as well. Can't wait to fully enable our new design. We're making strides and that's exciting! 😄

static/js/janitor-new.js
```diff
@@ -210,6 +240,13 @@ function updateFormStatus (form, status, message) {
     form.reportValidity();
   }
 
+  // Re-enable form fields once we receive a server response
+  if (status !== 'pending') {
+    Array.forEach(form.elements, function (element) {
+      element.classList.remove('disabled');
```
jankeromnes: Nit: I haven't tried it yet, but I think that our `.disabled` classes don't actually disable buttons (many users accidentally trigger duplicate container spawns/deletes by clicking on a button several times). Could be worth fixing, e.g. with `pointer-events: none`.

static/js/janitor-new.js
```diff
@@ -210,6 +240,13 @@ function updateFormStatus (form, status, message) {
     form.reportValidity();
   }
 
+  // Re-enable form fields once we receive a server response
```
jankeromnes: Note: I see that you're refactoring the `.disabled` state mechanism. Please bear in mind that we also used to remove all `.disabled` classes on form updates without explicit status (i.e. the server is still pending, but updateFormStatus is called when you modify a field, allowing you to submit another value while the first is still pending).

templates/settings-new.html
```diff
@@ -0,0 +1,92 @@
...
+            <div class="tab-panel" data-tab="account">
+              <section>
+                <h3>Basic information</h3>
+                <div class="setting">
+                  <b>Email</b>
+                  <input class="form-control" disabled="true" type="text" value={{= user._primaryEmail in json}}>
+                </div>
+                <form action="/api/user" class="setting ajax-form has-feedback" method="patch">
+                  <b>Name</b>
+                  <input class="form-control" data-submit-on="blur" name="/name" placeholder="Real name" type="text" value={{= user.profile.name in json}}>
```
jankeromnes: Nit: On wide screens, I feel like these "email" and "name" inputs are stretched way too wide. We used put both of them on the same line when space allowed it, but I'm not sure if that's the best solution. Maybe we need some reasonable `max-width`?

EDIT: I see that @arshdkhn1 agrees: `i feel like we can fit the email and name in one row and split them in tab/mob view`.

EDIT 2: `I actually like keeping them in separate rows, kinda like how the Github settings page is designed` also makes sense, but on GitHub, there is a menu/sidebar that ensures that input fields don't grow to extreme widths. Since we have our menu on the top, inputs width can grow indefinitely, which does look a little weird. But details, details. 😋 I'm ok to not block this pull request on this nit.

templates/settings-new.html
```diff
@@ -0,0 +1,92 @@
+        <div class="tabs">
+          <ul class="tab-nav" id="settings-tab-nav">
+            <li class="tab" data-tab="account"><a href="#">Account</a></li>
+            <li class="tab" data-tab="configurations"><a href="#">Configurations</a></li>
```
jankeromnes: You're moving the content of 4 current settings tab into just 2 tabs.

Regarding @arshdkhn1's comment about not clearly seeing that these are actually tabs, maybe adding at least a third one would make that more obvious visually?

And how are you planning to deal with URLs like https://janitor.technology/settings/integrations/ and https://janitor.technology/settings/configurations/ that might be in our various docs and our user's browsing history? Maybe there is a way to make them work, and point to the right settings section, even if it's now a single page.

templates/settings-new.html
```diff
@@ -0,0 +1,92 @@
...
+                  <b>Email</b>
+                  <input class="form-control" disabled="true" type="text" value={{= user._primaryEmail in json}}>
+                </div>
+                <form action="/api/user" class="setting ajax-form has-feedback" method="patch">
+                  <b>Name</b>
+                  <input class="form-control" data-submit-on="blur" name="/name" placeholder="Real name" type="text" value={{= user.profile.name in json}}>
+                </form>
+              </section>
+              <section>
+                <h3>Website settings</h3>
```
jankeromnes: Nit: I think "Website settings" is too generic. Maybe something like "Communication settings" is more appropriate? We should check how websites like GitHub and Twitter call these things.

templates/settings-new.html
```diff
@@ -0,0 +1,92 @@
...
+                <form action="/api/user/notifications/enabled?enabled={{= !user.notifications.enabled in boolean}}" class="row ajax-form has-feedback" data-refresh-after-success="true" id="notifications-form" method="put">
+                  <div class="checkbox">
+                    <label>
+                      <input autocomplete="off" {{if user.notifications.enabled then {{ checked="checked" }}}} data-submit-on="change" name="enabled" type="checkbox">
+                      Enable notifications
```
jankeromnes: As noted in the top review comment, maybe it would be nice to change this submit-on-change checkbox to a static form with a submit button? For this I'd see how checkboxes work in GitHub settings because they're my reference for excellent settings UX.

Also, random note, I'd like to add a few more checkboxes here soon, e.g. "Receive Janitor News emails" (synced with an upcoming MailChimp integration) and "Receive container expiration reminder emails" (might morph into "Receive important emails regarding your Janitor account" or so, if we add more email/reminder types).

templates/settings-new.html
```diff
@@ -0,0 +1,92 @@
...
+              <div class="panel">
+                <div class="panel-heading padded">
+                  <b class="grow">~/{{= file in html}}</b>{{ if (defaultConfigurations.includes(file)) { }}
+                  <form action="/api/user/configurations/{{= file in uri}}" class="ajax-form has-feedback" data-refresh-after-success="true" method="delete">
+                    <button class="btn btn-error" type="submit">Reset to Default</button>
+                  </form>
+                {{ } }}</div>
+                <div class="panel-body padded">
+                  <form action="/api/user/configurations/{{= file in uri}}" class="account-form ajax-form has-feedback" method="put">
+                    <p><strong>Note:</strong> This configuration will only apply to new containers. Overwrite <strong>~/{{= file in html}}</strong> in all existing containers by clicking <a data-submit-on="click" href="javascript:void(0)">here</a>.</p>
```
jankeromnes: Note: I'm ok with this note being no longer warning-yellow (I originally did this to draw the user's attention to the fact that even if you submitted a new config and it said OK, your existing containers will still have the older config), because the extra emphasis might no longer be needed.

Nit: However, I feel like the "here" link is way too close to the "Reset to Default" button, and this throws off my geographic memory for actions like these. If you don't mind, I'd much prefer this note to move below the config textarea again.

static/js/janitor-new.js
```diff
@@ -46,6 +46,41 @@ Array.forEach($$('.tabs'), function (element) {
...
+    if (name && !(name in data)) {
+      data[name] = element.type === 'checkbox' ? element.checked : element.value;
+    }
+    return data;
+  }, {});
+}
+
+// FIXME: Remove this deprecated function.
+// Set up an ajax <form>.
+function ajaxForm (selector, action, callback) { // eslint-disable-line no-unused-vars
```
jankeromnes: Nit: Instead of importing these two deprecated functions, how about writing a quick API handler to migrate the last remaining old-style form in settings? 😄

I only see `cloud9-form` left, and since it is longer needed for Cloud9 access, it would be nice to transform it into an API handler that allows users to add/remove private SSH keys. But since this is slightly more involved and a nit, you're allowed to simply remove the Cloud9 SSH key field from settings-new, and I'll work on an SSH public key import feature soon.

static/js/janitor-new.js
nt1m: That's already fixed in this PR.

templates/settings-new.html
nt1m: URLs like https://janitor.technology/settings/#account and https://janitor.technology/settings/#configurations will be supported in a separate PR.

The old urls will redirect to one of the two new urls.

static/js/janitor-new.js
nt1m: `pending` is a new state that I've introduced, which corresponds to the form during processing. It wasn't there in the old janitor.js, it's now there, just so we can potentially add some kind of loading throbber for the pending state.

static/css/janitor-new.css
```diff
@@ -665,26 +705,27 @@ article.onebox-body {
   margin-top: 0;
 }
 
-article a {
-  text-decoration: none;
+/* Settings page */
+
+#settings-tab-nav {
+  font-size: 18px;
```
arshdkhn1: its too big. even bigger than h3. we can tweak background and other css to make tab look better but this font size looks inappropriate imho

templates/footer-new.html
```diff
@@ -26,12 +26,13 @@
...
             <span>Janitor</span>
           </a>
-          <p class="footer-logo-tagline">The fastest development system in the world.</p>
+          <span class="footer-logo-tagline grey-text">The fastest development system in the world.</span>
         </div>
       </div>
     </footer>
     <script defer src="/js/promise-6.0.2.min.js"></script>
     <script defer src="/js/fetch-2.0.3.min.js"></script>
+    <script defer src="/js/scout.min.js"></script>
```
nt1m: note to self: undo this

arshdkhn1: note to self: bad review

templates/settings-new.html
```diff
@@ -0,0 +1,82 @@
...
+                      </div>
+                      <div><button class="btn disabled" disabled="true">Coming Soon&trade;</button></div>
+                    </div>
+                  </div>
+                </div>
+              </section>
+            </div>
+            <div class="tab-panel" data-tab="configurations">
+              <p>Configure your containers from a single location, by editing and deploying these configuration files.</p>
+              {{ for (let file of Object.keys(user.configurations).sort()) { }}
```
jankeromnes: Super-nit: This template will produce an HTML with bogus whitespace. Please move the `{{` to the end of the line above.

PR #252, opened 2018-01-31
New 404 page

arshdkhn1:
image

PR #253, opened 2018-02-01
New API page

arshdkhn1:
Updated screenshot:
api


static/css/janitor-new.css
```diff
@@ -306,6 +322,55 @@ p {
   margin: 0;
 }
 
+/* api reference page */
+
+div.reference {
+  padding: 40px 0 60px 0;
+  display: flex;
+  flex: 1;
+  flex-direction: column;
```
arshdkhn1: making reference div flex with column-direction, avoids wrapping of child elements even when individual element width is not 100%.

static/css/janitor-new.css
```diff
@@ -306,6 +322,55 @@ p {
...
+}
+
+div.reference p {
+  margin: 30px 0 0 0;
+}
+
+div.reference h1,
+div.reference h2 {
+  /* this makes h1, h2 tags acts like inline-block and doesn't wrap content in same line */
+  align-self: flex-start;
```
arshdkhn1: flex-start makes sure width of h1, h2 element is equal to their content width. If they have 100% width(that they have by default) then clicking on white-space area will also trigger click on them. 

What i meant by white-space area is the empty red area on the right after `Introduction`  -
image

In api.js I am listening for click on h2 and h1 elements, so to avoid the click trigger on the above empty area, I have used `align-self: flex-start` here.

static/js/api.js
```diff
@@ -0,0 +1,10 @@
+const changeUrlHash = e => {
+    const targetNode = e.target.nodeName;
+    if (targetNode === 'H1' || targetNode === 'H2') {
```
arshdkhn1: if click is on h1 or h2 tag in api page, that means it will be having a child element like - `<small><a href="...">....`. 
This code makes h1,h2 element clickable. Clicking on pound sign only is hard for user.

templates/reference-api-new.html
```diff
@@ -0,0 +1,3 @@
+    <div class="reference">
```
arshdkhn1: changed section to div, based upon @nt1m comments in #252

lib/routes.js
```diff
@@ -135,6 +135,24 @@ exports.apiPage = function (response, api, user = null) {
...
+const apiSectionNew = camp.template('./templates/reference-api-new.html');
+exports.apiPageNew = function (response, api, user = null) {
+  const title = 'API Reference';
+  const htmlReference = api.toHTML();
+
+  response.template({
+    htmlReference,
+    title,
+    user,
+    scripts: ['/js/api.js'],
```
jankeromnes: Nit: Please expand this array on multiple lines (this make `git blame` clearer for individual script changes, e.g. new scripts or renames).

static/css/janitor-new.css
```diff
@@ -306,6 +322,55 @@ p {
   margin: 0;
 }
 
+/* api reference page */
+
+div.reference {
```
jankeromnes: General comment: It seems that by introducing this `.reference` class, you're "forking" the style of Janiror (i.e. below you're making a lot of custom rules for `.reference` that won't apply to all other pages, and this means that the style of API page style and other Janitor pages will evolve separately, in different directions). Maybe it's not practical in this case, but in general I like generic CSS where every style that can be common between every page is common (and not every page has its own custom style).

static/js/api.js
```diff
@@ -0,0 +1,9 @@
+const changeUrlHash = e => {
+  const targetNode = e.target.nodeName;
```
jankeromnes: Nit: We don't use ES6 in client-side scripts yet, in order to support older browsers. Please explain why it's now ok to migrate all client scripts to ES6, or please refrain from using `const` and arrow functions on the client for now.

static/js/api.js
```diff
@@ -0,0 +1,9 @@
...
+  if (targetNode === 'H1' || targetNode === 'H2') {
+    const hash = e.target.querySelector('a').getAttribute('href');
+    window.location.hash = hash;
+  }
+};
+
+document.body.addEventListener('click', changeUrlHash);
```
jankeromnes: Whoa, isn't `document.body` a bit extreme? I would have used a more precise query selector.

static/css/janitor-new.css
arshdkhn1: @jankeromnes I am waiting for Tim to review this. I didn't want to change the h2, h3, h1 etc css definitions. I wanted to confirm with @notriddle whether he is happy with these h1, h2... I know i am using a version control so I must worry about this but I knew I have to update it anyway that's why I left it as it is.

lib/routes.js
arshdkhn1: ohk

static/js/api.js
arshdkhn1: Sorreh

static/js/api.js
arshdkhn1: yes, i was trying this hack to avoid iterating h2/h1 elements. Since you answered the question i asked in selfapi issue, I ll be changing it anyway.

static/js/api.js
nt1m: I really think this JS code is unnecessary, the `#` sign is enough to get a link. If were to do this, then this code should also apply to the blog page.

static/css/janitor-new.css
nt1m: As @jankeromnes said, please avoid adding so many rules that are page-specific. The API page is pretty similar to the Blog page, so I'd like some style sharing being done between the two. Please try to simplify as much as you can the rules you've added :)

static/css/janitor-new.css
```diff
@@ -20,8 +20,15 @@ body {
...
-  visibility: hidden;
+.show-on-hover a{
+  color: #0b0b0d;
+  text-decoration: none;
+  transition: color 200ms ease-in;
+  padding: 10px;
+}
+
+*:not(:hover) > .show-on-hover a {
+  color: #aaa;
```
notriddle: I don't think you really need to use `:not` here any more...

static/js/api.js
notriddle: Isn't there already a loop in janitor-new.js that you would want to add this to?

static/css/janitor-new.css
arshdkhn1: @jankeromnes  @nt1m done!

@notriddle addressed your nits.

static/css/janitor-new.css
```diff
@@ -20,8 +20,15 @@ body {
...
-*:not(:hover) > .show-on-hover {
-  visibility: hidden;
+.glow-on-hover a {
+  color: #aaa;
+  text-decoration: none;
+  transition: color 200ms ease-in;
+  padding: 10px;
+}
+
+*:hover > .glow-on-hover a {
```
arshdkhn1: @nt1m why is it needed to change css styles on hovering of h1/h2 tag. Wdyt about changing color when # sign is hovered?

nt1m: It was needed before for .show-on-hover, it's no longer needed now. You can change it to `.glow-on-hover:hover a`

static/css/janitor-new.css
```diff
@@ -306,6 +332,12 @@ p {
   margin: 0;
 }
 
+/* api reference page */
+
+div.reference {
+  padding-bottom: 60px;
```
nt1m: Does this bottom padding really make a difference ?

static/css/janitor-new.css
```diff
@@ -718,4 +742,4 @@ article a:focus {
   box-shadow: 0 0 0 1px #0a84ff, 0 0 0 4px rgba(10, 132, 255, 0.2);
   border-radius: 2px;
   text-decoration: none;
-}
+}
```
nt1m: Please undo this change

static/css/janitor-new.css
```diff
@@ -20,10 +20,15 @@ body {
   min-height: 100%;
 }
 
-/* Helper classes */
```
nt1m: Please also don't remove this comment.

templates/settings-new.html
```diff
@@ -6,7 +6,7 @@
           <div class="tab-panels">
             <div class="tab-panel" data-tab="account">
               <section>
-                <h3>Basic information</h3>
+                <h2>Basic information</h2>
```
arshdkhn1: @nt1m I think I should create a separate pr ffor this?

templates/login-new.html
```diff
@@ -1,5 +1,5 @@
     <div class="text-center centered-container">
-      <h2>Authenticate</h2>
+      <h1>Authenticate</h1>
```
arshdkhn1: @nt1m  I think I shhould make a separate pr for this or is it fine to do this minor change in this pr itself?

static/js/janitor-new.js
```diff
@@ -257,12 +257,10 @@ if (window.location.search) {
...
   const link = document.createElement('a');
   link.href = '#' + element.id;
-  link.textContent = '#';
-  const small = document.createElement('small');
-  small.classList.add('show-on-hover');
-  small.appendChild(link);
-  element.appendChild(document.createTextNode(' '));
-  element.appendChild(small);
+  const span = document.createElement('span');
+  span.classList.add('icon', 'link');
```
nt1m: I don't think you need this inner span, just make the `a` directly an `.icon`.

templates/login-new.html
nt1m: It's fine in this PR :)

templates/settings-new.html
nt1m: It's fine to me in this PR :)

static/js/janitor-new.js
arshdkhn1: good point!

static/css/janitor-new.css
```diff
@@ -372,6 +390,17 @@ textarea.form-control {
   line-height: 1.5;
 }
 
+/* Link icon shared between blog and api page */
+
+.icon.link {
```
nt1m: Can you introduce a `.icon-button` class that has those opacity and padding styles, and use that for the link icon ?
Also, can you move this with the rest of the `.icon` CSS ?

arshdkhn1: sure

PR #254, opened 2018-02-01
New data page

arshdkhn1:
data

### DIFF
--- a/static/css/janitor-new.css
+++ b/static/css/janitor-new.css
@@ -112,7 +112,7 @@ small {
 p {
   color: #666;
   font-weight: 400;
-  margin: 20px 0;  
+  margin: 20px 0;
 }
 
 pre {
@@ -666,6 +666,8 @@ textarea.form-control {
   background-image: url(../img/icons/link.svg);
 }
 
+/* Icon buttons */
+
 .icon-button {
   padding: 10px;
   opacity: .4;
@@ -804,7 +806,7 @@ article.onebox-body {
   margin-bottom: 2em;
 }
 
-/* api page */
+/* API page */
 
 .reference h1:not(:first-child) {
   margin-top: 90px;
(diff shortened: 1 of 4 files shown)


jankeromnes:
Wow, this looks great! A few quick comments off hand:

- please remove the "0 merged". It's wrong and will probably never work since we've changed what the status does
- please rename "contributions" to "containers"

templates/data-new.html
```diff
@@ -0,0 +1,32 @@
...
+          <span class="unit">confirmed</span>
+          <span class="number">{{= data.users.waitlist.length in integer}}</span>
+          <span class="unit">waitlist</span>
+        </p>
+      </div>
+      <div class="data-topic">
+        <h3>Containers</h3>
+        <p class="numbers">
+          <span class="number">{{= data.contributions.started in integer}}</span>
+          <span class="unit">started</span>
```
arshdkhn1: @jankeromnes I have removed the `merged` status. I hope this is what you meant.

arshdkhn1:
#253 needs to be merged before this pr. I will then have to update this pr.

templates/data-new.html
jankeromnes: Yes, perfect. Thank you! 😄

static/css/janitor-new.css
```diff
@@ -296,6 +307,58 @@ p {
...
+.data .data-topic:nth-child(2n-1) {
+  border-right: none;
+} 
+
+.data div:nth-child(4),
+.data div:nth-child(5) {
+  border-bottom: none;
+}
+
+.data div:last-child::after {
```
nt1m: Can you add a comment with what this represents ? Also, I think `.data::after` might be better, because right now, it kinda makes an assumption that it has a link with `:last-child` when it really doesn't.

static/css/janitor-new.css
```diff
@@ -296,6 +307,58 @@ p {
...
+.data-topic span.unit:nth-child(2) {
+  margin-right: 10px;
+}
+
+.data .data-topic:nth-child(2n-1) {
+  border-right: none;
+} 
+
+.data div:nth-child(4),
+.data div:nth-child(5) {
```
nt1m: Is `:nth-child(5)` used ? Also, can you use .data-topic instead of `div` ?

static/css/janitor-new.css
```diff
@@ -296,6 +307,58 @@ p {
...
+}
+
+.data h1 {
+  width: 100%;
+  margin-bottom: 80px;
+}
+
+.data .data-topic {
+  flex: 0 1 50%;
+  /* background: red; */
```
nt1m: nit: remove this leftover

static/css/janitor-new.css
```diff
@@ -296,6 +307,58 @@ p {
...
+  padding: 30px;
+  border-right: solid 1px #d7d7db;
+  border-bottom: solid 1px #d7d7db;
+  position: relative;  
+}
+
+.data-topic .number {
+  font-size: 18px;
+  font-weight: 700;
+  color: #0c9100;
```
nt1m: Use a photon color :) https://design.firefox.com/photon/visuals/color.html

static/css/janitor-new.css
arshdkhn1: image
^ its use. I know its unnecessry but I like the psuedo separation that it creates between the border.

nt1m: Yeah, but as I said, add a comment, it's not explicit. Also, can you just put this on `.data::after` ?

arshdkhn1:
Since .Data consists of h1 and data topics, therefore positioning this
element will be complicated. Feel free to try it urslf.

nt1m:
@arshdkhn1 Alright, I'm fine with using `.data .data-topic:last-child::after` then, but please do add a comment saying what this does.

static/css/janitor-new.css
arshdkhn1: yes(h1 is first child of data-topic)

arshdkhn1: Edit: I should use nth-f-type instead.

PR #255, opened 2018-02-01
404 page nits given by @jankeromnes

arshdkhn1:


static/css/janitor-new.css
```diff
@@ -82,7 +82,7 @@ main {
   overflow: auto;
   flex-grow: 1;
   /* viewport height - header height - 15px extra height(so that footer is visible) */
-  min-height: calc(100vh - 95px);
+  min-height: calc (100vh - 95px);
```
nt1m: I don't like this style personally.

app.js
```diff
@@ -455,6 +455,13 @@ boot.executeInParallel([
     routes.notFoundPageNew(query.res, user);
   });
 
+  // New 404 Not Found page
+  app.route(/^\/404-new\/?$/, (data, match, end, query) => {
```
nt1m: Shouldn't this be replacing the block above ?

static/css/janitor-new.css
jankeromnes: My bad, I meant the `(` in the comment, not the one after `calc`. I'll fix up this pull request.

app.js
jankeromnes: Will fix this as well.

PR #271, opened 2018-02-08
Add Azure REST API client to manage Windows virtual machines.

### LINKED ISSUE #251: Support 'aux' channel in Docker Remote API streamed JSON responses
Looks like there is now an `aux` channel in some Docker Remote API responses (in addition to the `error` and `stream` channels already supported by [our DockerResponse parser](https://github.com/JanitorTechnology/janitor/blob/cdae5656adce5d3fde29a83662300ca161987ae2/lib/docker.js#L371-L378)).

It seems to add meta-data information, like the ID of the image we've just built:

```
[2018-01-30T00:47:17.435Z] update janx/chromium:janitor-production Error: Unknown format: {"aux":{"ID":"sha256:f35bcc41cf6ff8a1cd5213b677f8d0cb678d2a53af117ab476b5d5d41da8658a"}}
    at DockerResponse._parse (/home/jan/janitor/lib/docker.js:377:26)
    at DockerResponse._transform (/home/jan/janitor/lib/docker.js:359:10)
    at DockerResponse.Transform._read (_stream_transform.js:186:10)
    at DockerResponse.Transform._write (_stream_transform.js:174:12)
    at doWrite (_stream_writable.js:385:12)
    at writeOrBuffer (_stream_writable.js:371:5)
    at DockerResponse.Writable.write (_stream_writable.js:288:11)
    at IncomingMessage.ondata (_stream_readable.js:642:20)
    at emitOne (events.js:115:13)
    at IncomingMessage.emit (events.js:210:7)
```

I think we should support that new `aux` channel, or at least not throw an error when we encounter it.

### LINKED ISSUE #268: Implement an Azure REST API client to manage Windows VMs
Code sample for inspiration: https://github.com/Azure-Samples/key-vault-node-deploy-certificates-to-vm/blob/master/index.js

jankeromnes:
Fixes #268.

I'm also sneaking in the following drive-by changes:

- Refactor stream data processing to use Buffer.concat() (following @bnjbvr's nit suggestion)
- Support 'aux' channel in Docker Remote API response streams, fix #251 (because I came across the code during the refactor)
- Remove annoying package-lock.json file from repository (it always causes conflicts and I don't find it useful)
- Add annoying package-lock.json file to .gitignore (see above)
- Remove deprecated in-app proxy handlers now that app and proxies were separated (stolen from #188 because review is taking too long over there 😋 also it mitigates a WebSocket DDoS found by @Snyk)


jankeromnes:
@nt1m or @bnjbvr could you please have a look at this, whoever has time for it first? 😄 

This is an admin-only feature (so no need to spend a lot of effort on UX tweaks), and here is what it looks like (works locally in my container):

azure-admin

api/admin-api.js
```diff
@@ -13,6 +14,80 @@ const adminAPI = module.exports = selfapi({
...
+      response.statusCode = 403; // Forbidden
+      response.json({ error: 'Unauthorized' }, null, 2);
+      return;
+    }
+
+    const { credentials } = db.get('azure');
+
+    let json = '';
+    request.on('data', chunk => {
+      json += String(chunk);
```
bnjbvr: nit: potential hard to debug encoding issue here; use `Buffer` for the initial value, `Buffer.concat` here, and then JSON.parse the `buffer.toString()` at the end

lib/azure.js
```diff
@@ -0,0 +1,79 @@
...
+
+// Load our Azure Active Directory application configuration.
+function load () {
+  const azure = db.get('azure');
+  if (!azure.credentials) {
+    azure.credentials = {};
+  }
+
+  // You can customize these values in `./db.json` or via `/admin/`.
+  const { credentials } = azure;
```
bnjbvr: Considering that the defaulting above has no effect (not saved), you could just do `const credentials = azure.credentials || {};`

lib/azure.js
```diff
@@ -0,0 +1,79 @@
...
+
+  // "Application ID" or "Client ID".
+  if (!credentials.clientId) {
```
bnjbvr: if it's a numerical value, note it could be falsy for 0.

lib/azure.js
```diff
@@ -0,0 +1,79 @@
...
+    credentials.clientId = '';
+  }
+
+  // "Application Secret" or "Authentication Key".
+  if (!credentials.clientSecret) {
+    credentials.clientSecret = '';
+  }
+
+  // "Domain" or "Directory ID" or "Tenant ID".
+  if (!credentials.tenantId) {
```
bnjbvr: ditto

lib/azure.js
```diff
@@ -0,0 +1,79 @@
...
+    credentials.tenantId = '';
+  }
+
+  // "Azure Subscription ID".
+  if (!credentials.subscriptionId) {
```
bnjbvr: ditto

lib/azure.js
```diff
@@ -0,0 +1,79 @@
...
+    credentials.subscriptionId = '';
+  }
```
bnjbvr: you need to `db.save()` here, right?

lib/azure.js
jankeromnes: I suppose that `0` is not a valid Azure Application ID anyway (they have a certain minimum length), so that's ok.

lib/azure.js
jankeromnes: ditto

lib/azure.js
jankeromnes: ditto

lib/azure.js
```diff
@@ -0,0 +1,79 @@
...
+          reject(error);
+          return;
+        }
+
+        resolve({ credentials, subscriptions });
+      }
+    );
+  });
+
+  const computeClient = new ComputeManagementClient(credentials, subscriptionId);
```
nt1m: You can do:
```
return new ComputeManagementClient(credentials, subscriptionId);
```

lib/azure.js
jankeromnes: Actually no. Here we're in the case were, immediately when started, the server sets a few empty default values. We don't need to `db.save()` them, because even if the server crashes, upon reboot the exact same values will be set again.

(Also, calling `db.save()` at approximately the same time as a server crash exposes us to wiping the `db.json` file. So calling `db.save()` on startup is both pointless *and* dangerous, e.g. if you have obvious accidental `TypeErrors` somewhere.)

lib/azure.js
jankeromnes: Actually no, because we need `azure` to be a real DB reference (`db.save()` only serves to dump the DB to disk, but any changes to something from `db.get(key)` will work anywhere, since it's a direct in-memory DB reference).

With your suggestion, templates retrieving the `azure` object from the DB (e.g. the integrations admin page) won't have empty credential values (the default value for a non-existant `db.get(key)` is `{}`).

jankeromnes:
@bnjbvr Thanks a lot for your review! I've addressed your nits (everywhere applicable, e.g. by refactoring all our stream processing snippets to use `Buffer.concat()`) and replied to the ones I decided not to act upon.

I also fixed a drive-by bug, #251. Could you please have a look at my new changes? Thanks! 😀

api/admin-api.js
```diff
@@ -13,6 +14,79 @@ const adminAPI = module.exports = selfapi({
...
+      const virtualMachines = await azure.getAllVirtualMachines();
+      response.json(virtualMachines, null, 2);
+    } catch (error) {
+      log('[fail] fetching azure virtual machines', error);
+      response.statusCode = 500; // Internal Server Error
+      response.json({ error: 'Could not fetch virtual machines' }, null, 2);
+    }
+  },
+
+  examples: [],
```
nt1m: Does this need examples for tests ?

static/js/admin.js
```diff
@@ -70,6 +70,23 @@ Array.map(document.querySelectorAll('button[data-action]'), function (button) {
...
 });
 
+// Azure Virtual Machines button.
+Array.forEach(document.querySelectorAll('.azure-virtual-machines'), function (div) {
+  var button = div.querySelector('button');
+  var pre = div.querySelector('pre');
+  button.addEventListener('click', function (event) {
+    var url = '/api/admin/azure/virtualmachines';
+    pre.textContent = 'Fetching virtual machines…';
+    window.fetchAPI('GET', url, null, function (error, virtualMachines) {
```
nt1m: nit: janitor/janitor-new.js just use `fetchAPI`, while admin.js uses `window.fetchAPI`. Please be consistent. (I prefer just `fetchAPI` because it's shorter).

api/admin-api.js
jankeromnes: Eventually yes, but I don't want to spend time on that yet. I think it's ok to omit examples for now.

static/js/admin.js
jankeromnes: Actually, this file has a lot of occurrences of `window.fetchAPI`. It was to silence Editor gutter warnings in Cloud9 IDE.

If you don't mind, I'd like to leave this small inconsistency between files for now, but we should definitely restore consistency as a follow-up.

nt1m: Sounds good !

jankeromnes:
Notes:

- I told @Snyk to ignore all `socket.io` vulnerabilities forever, because we don't use `camp`'s optional `socket.io` module. (Since we never read the `camp.io` [magic property](https://github.com/espadrine/sc/blob/1ebcc6ca201427593826d44dcf6cdbf909f702f5/lib/camp.js#L417-L431), we should never even load `socket.io` modules into memory. So in theory, it's just dead weight in our NPM dependencies.)
- I also told @Snyk to ignore the `ws` DDoS vulnerability for 30 days. This one does affect us, although just our Docker host proxies (i.e. `join.js`, and not our main web server `app.js`, since I've just deleted the deprecated WebSocket proxy there). This should resolve itself when @espadrine gets a change to upgrade `camp` to `ws 4.0.0` and/or to enable https://greenkeeper.io (see https://github.com/espadrine/sc/issues/66).

jankeromnes:
Thanks a lot for the reviews! 💯

PR #273, opened 2018-02-10
Remove main from header and reposition warning-alert.

arshdkhn1:
image

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -69,7 +69,7 @@ const appHeader = camp.template([
 const appHeaderNew = camp.template([
   './templates/header-new.html',
 ].concat(!security.forceInsecure ? [] : [
-  './templates/header-insecure.html',
+  './templates/header-insecure-new.html',
 ]));
 
 const appFooter = camp.template('./templates/footer.html');
--- a/static/css/janitor-new.css
+++ b/static/css/janitor-new.css
@@ -170,8 +170,14 @@ main {
   display: flex;
   flex-direction: column;
   flex-grow: 1;
+  z-index: 0;
   /* viewport height - header height - 20px extra height (so that footer is visible) */
   min-height: calc(100vh - 100px);
+}
+
+.global-alert + main {
+  /* viewport height - header height - 20px extra height (so that footer is visible) */
+  min-height: calc(100vh - 155px);
 }
 
 .centered-container {
@@ -344,7 +350,7 @@ textarea.form-control {
 
 .navbar {
   background: #fff;
-  z-index: 999;
+  z-index: 1;
 }
 
 .navbar-container {
@@ -386,6 +392,7 @@ textarea.form-control {
 .footer {
   border-top: solid 1px #eee;
   background: #fafafa;
+  z-index: 0;
 }
 
 .footer .footer-inner-container {
@@ -698,15 +705,28 @@ textarea.form-control {
 
 /* Alerts */
 
+.global-alert {
+  order: -1;
+  z-index: 2;
+}
+
 .alert {
-  margin-bottom: 1.5em;
-  padding: 10px;
-  border-radius: 4px;
+  padding: 15px;
+}
+
+.alert > span {
+  vertical-align: middle;
+}
+
+.alert > .icon::before {
+  width: 1em;
+  height: 1em;
+  margin-right: .5em;
 }
 
 .alert.alert-warning {
-  background: #ffe900;
-  color: #3e2800;
+  background: #fff8b2;
+  color: #715100;
 }
 
 /* Icons */
@@ -737,6 +757,10 @@ textarea.form-control {
 
 .icon.link::before {
   background-image: url(../img/icons/link.svg);
+}
+
+.icon.warning::before {
+  background-image: url(../img/icons/warning.svg);
 }
 
 /* Icon buttons */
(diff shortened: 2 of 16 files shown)


arshdkhn1:
Made this pr in a bit hurry so there might be a lot of areas which need improvements.

templates/announcement-bar.html
```diff
@@ -0,0 +1,6 @@
+    <div id="announcement-bar">
+        <p class="alert alert-warning text-center" role="alert">
```
nt1m: .alert is only used in the "announcement bar". So no need to wrap the alert with the extra div, feel free to tweak the alert style directly (rather than introducing #announcement-bar specific styles)

arshdkhn1: I have added that so that in future if a condition come where we want to add two or more alert(like survey one), it can be useful.

static/css/janitor-new.css
```diff
@@ -699,14 +704,13 @@ textarea.form-control {
...
-  border-radius: 4px;
+  padding: 15px;
 }
 
 .alert.alert-warning {
-  background: #ffe900;
-  color: #3e2800;
+  background: #fcf8e3;
+  color: #3e2800b3;
+  border-color:#faeb99;
```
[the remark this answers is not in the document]
arshdkhn1: gc!

static/css/janitor-new.css
```diff
@@ -739,6 +742,12 @@ textarea.form-control {
   background-image: url(../img/icons/link.svg);
 }
 
+.icon.warning::before {
+  background-image: url(../img/icons/warning.svg);
+  vertical-align: bottom;
+  transform: scale(.8);
```
nt1m: +  vertical-align: bottom;
+  transform: scale(.8);

shouldn't be part of the .icon.warning class.

I would prefer a more specific selector for those 2 properties.

static/css/janitor-new.css
```diff
@@ -780,3 +789,21 @@ textarea.form-control {
...
+  order: -2;
+}
+
+#announcement-bar p {
+  margin: 0;
+}
+
+@media screen and (max-width: 768px) {
+  #announcement-bar {
+    font-size: 12px;
```
nt1m: It seems a bit overkill to add a media query just for this. Is there a better solution ?

lib/routes.js
```diff
@@ -69,7 +69,7 @@ const appHeader = camp.template([
 const appHeaderNew = camp.template([
   './templates/header-new.html',
 ].concat(!security.forceInsecure ? [] : [
-  './templates/header-insecure.html',
+  './templates/announcement-bar.html',
```
nt1m: Please leave it as `header-insecure.html`, because it only appears if `security.forceInsecure` is true. For global site alerts, I'd prefer having them in `header.html` directly.

templates/announcement-bar.html
```diff
@@ -0,0 +1,6 @@
+    <div id="announcement-bar">
```
nt1m: I still don't think we should have this div.

static/css/janitor-new.css
```diff
@@ -699,14 +704,12 @@ textarea.form-control {
...
-  margin-bottom: 1.5em;
-  padding: 10px;
-  border-radius: 4px;
+  padding: 15px;
 }
 
 .alert.alert-warning {
-  background: #ffe900;
-  color: #3e2800;
+  background: #fcf8e3;
```
nt1m: This yellow feels too dull to me. How does #ffe900 at 0.5 opacity work for you ? (Also please use rgba rather than 8 char hex)

arshdkhn1: for consistincy, right? are they(8charhex) not supported by some browser?

nt1m: For compatibility and consistency. 8 char hex is a recent feature

static/css/janitor-new.css
arshdkhn1: i can't think of any atm, if you see the janitor-new.css then u ll see 3 more similar mq having just 1 rule.

static/css/janitor-new.css
arshdkhn1: #ffe900 looks quite bad on mobile screen.

static/css/janitor-new.css
```diff
@@ -170,10 +170,16 @@ main {
   display: flex;
   flex-direction: column;
   flex-grow: 1;
+  z-index: 0;
   /* viewport height - header height - 20px extra height (so that footer is visible) */
   min-height: calc(100vh - 100px);
 }
 
+.global-alert ~ main {
```
arshdkhn1: `+` will be more specific as per current markup

static/css/janitor-new.css
```diff
@@ -698,15 +705,28 @@ textarea.form-control {
...
+.alert > .icon::before {
+  width: 1em;
+  height: 1em;
+  margin-right: .5em;
 }
 
 .alert.alert-warning {
-  background: #ffe900;
-  color: #3e2800;
+  background: #fff8b2;
```
arshdkhn1: tbh, i dont like the color, on lcd screen it is too warm.

templates/project-new.html
```diff
@@ -1,19 +1,20 @@
...
+        </div>
       </div>
-      <div class="graph">
-        <div data-data="{{= project.data['update-time'] in json}}" data-title="Image update"></div>
-      </div>
-      <div class="graph">
-        <div data-data="{{= project.data['spawn-time'] in json}}" data-title="Container spawn"></div>
-      </div>
-    </div>
+    </main>
```
arshdkhn1: have to insert an endline.

arshdkhn1: done

PR #296, opened 2018-03-12
Add Datadog dashboards to data pages

Coder206:
This PR adds the links to the Datadog dashboards to the data pages. 

For the new design, I was unsuccessful in making the "Metrics" section part of the existing `data-container` with `data-topic` as this would result in a vertical line placed to the left. Do you have suggestions to fix this using CSS? 
screenshot-2018-3-12 data janitor 1

I experimented with the following design as well, it removes the `data-topic` class and this PR currently implements this.
screenshot-2018-3-12 data janitor

In any case, I think it would be nicer if the buttons would look centred (having the space between them sit centred under the word "Metrics").

Finally, instead of links, I think it would be great to embed these dashboards directly in the app, My concern is that the dashboard's content does not seem to be responsive to the size of the iframe, so you need to scroll to see the dashboard.

Would greatly appreciate feedback on this.


nt1m:
"Running containers" might be a better fit as a link in the "Containers" data section, and server metrics could be a link in the "Clusters" section ?

Coder206:
@nt1m Thanks for this suggestion, I think it's a great idea! 

Please note, as we are implementing this in both data pages (old and new), I decided to place "Running containers" under "Projects" as fits best there (there is no "Containers" section in the old data page)

jankeromnes:
Ping @nt1m or @arshdkhn1 about the review. 😄

jankeromnes:
> This is generally fine to me, I do think it might look better as links as opposed to buttons, but I don't have strong opinions here.

Thanks a lot for reviewing this change! I also think that links might be less surprising than buttons, but I also don't have a strong opinion.

So I'll merge this. Thanks again @Coder206! 💯 🥇 👍 I find these dashboards super cool!

PR #298, opened 2018-03-16
Fix WebSocket URL rewriting

ishitatsuyuki:
Theia uses the page URL as the WS endpoint, thus it has to be rewritten. ~Adding to that WebSocket handshakes get a Referer, so our logic just works.~

Some regex related code are modified for clarity (discard variables we don't care, so the variable doesn't have multiple meanings).

### DIFF
--- a/lib/proxy-heuristics.js
+++ b/lib/proxy-heuristics.js
@@ -1,6 +1,7 @@
 // Copyright © 2017 Jan Keromnes. All rights reserved.
 // The following code is covered by the AGPL-3.0 license.
 
+const http = require('http');
 const nodeurl = require('url');
 
 const log = require('./log');
@@ -31,6 +32,14 @@ exports.handleProxyUrls = function (request, response, next) {
     // However `path` is empty in URLs like '/abc123/8080?p=1', so we redirect
     // them to '/abc123/8080/?p=1' (where `path` is '/').
     if (!path) {
+      // We can only use `http.ServerResponse`s to redirect,
+      // not raw `net.Socket`s (as in WebSocket connections).
+      if (!(response instanceof http.ServerResponse)) {
+        const error = new Error('Unsupported response type (e.g. WebSocket)');
+        log('[fail] trailing slash redirect', error);
+        response.end();
+        return;
+      }
       url = url.includes('?') ? url.replace('?', '/?') : url + '/';
       routes.redirect(response, url, true);
       return;
--- a/join.js
+++ b/join.js
@@ -73,9 +73,9 @@ boot.executeInParallel([
       request.query = {};
       ensureSession(request, socket, () => {
         ensureOAuth2Access(request, socket, () => {
-          // Note: We don't handle proxy URLs here, because they don't work with
-          // WebSockets (no 'referer' HTTP header in WebSocket requests).
-          proxyRequest(request, socket);
+          proxyHeuristics.handleProxyUrls(request, socket, () => {
+            proxyRequest(request, socket);
+          });
         });
       });
     });


jankeromnes:
Exciting! I was pretty disappointed when I saw that websockets never specified a `referer` header. Has this changed?

I look forward to taking a closer look to your changes, and to take them for a spin. Thanks a lot for digging deep into our proxy code to fix Theia! :+1:

ishitatsuyuki:
Broke some styling, fixing soon.

ishitatsuyuki:
Sorry seems the Referer thing was my mistake (I probably confused when testing using curl).

lib/proxy-heuristics.js
```diff
@@ -81,7 +90,7 @@ exports.requestLikelyForPort = {
     }
     const match = exports.cloud9VfsUrlPrefix.exec(url);
     if (match) {
-      this._cloud9VfsUrlPrefix = match[0];
+      this._cloud9VfsUrlPrefix = match[ 0 ];
```
jankeromnes: Wait, what? If this is requested by eslint, then it's probably an eslint regression and we should downgrade our devDependencies. I really don't think that we want to add spaces to Array item lookups.

ishitatsuyuki: Uh, no. I'm using IntelliJ Idea and seems I misconfigured the formatter.

jankeromnes:
Checked out and rebased your branch on Pianosa. There seem to be a few errors. 😋 

- https://pianosa.janitor.technology/5f429a8d7b3c3e72/8089/ide.html gives me "404 Cannot GET /5f429a8d7b3c3e72/8089"
- https://pianosa.janitor.technology/5f429a8d7b3c3e72/8088/vnc.html "Error code 404. Message: File not found."
- https://pianosa.janitor.technology/5f429a8d7b3c3e72/8090/ gives me a hard 500 error with no content

From the logs, it looks like you're eating a forward slash somewhere:

```
[2018-03-16T14:24:05.928Z] [fail] invalid proxy path: 5f429a8d7b3c3e72/8090/
```

lib/proxy-heuristics.js
```diff
@@ -31,19 +32,27 @@ exports.handleProxyUrls = function (request, response, next) {
...
+        return;
+      }
       url = url.includes('?') ? url.replace('?', '/?') : url + '/';
       routes.redirect(response, url, true);
       return;
     }
 
     // Locally remove the prefix from `request.url`.
-    request.url = request.url.replace('/' + containerId + '/' + port, '');
+    request.url = path;
```
jankeromnes: Why did you need to change this line? I feel like it's best to stay as close as possible to the original `request.url` (e.g. if one day we make our `path` regex more specific, we may start dropping things like query parameters during such redirects).

lib/proxy-heuristics.js
```diff
@@ -31,19 +32,27 @@ exports.handleProxyUrls = function (request, response, next) {
...
   } else if (request.headers.referer) {
     // Look for a container ID and port in `request.headers.referer`.
     const referer = nodeurl.parse(request.headers.referer);
     match = exports.proxyUrlPrefix.exec(referer.pathname);
     if (match) {
-      [url, containerId, port, path] = match;
+      [, containerId, port] = match;
```
jankeromnes: Nit: I guess this is a matter of preference. I like to always use the same de-structuring, so that it's always obvious what format you expect for `match` (and it's always the same, which is easier for copy/pasting or refactoring), even if `url` and `path` aren't actually used. But ok if you insist on doing it like this.

lib/proxy-heuristics.js
```diff
@@ -2,6 +2,7 @@
 // The following code is covered by the AGPL-3.0 license.
 
 const nodeurl = require('url');
+const http = require('http');
```
jankeromnes: Super-nit: Please sort `require`s of the same type alphabetically (`http` < `nodeurl`).

lib/proxy-heuristics.js
ishitatsuyuki: Yeah, it's not strictly required to change this. But on the other hand: we don't constraint the replacement to the start of the path, which means it's somewhat flawed too.

Anyway, if you don't like this change to be coupled, I will happily swap it out and do it separately in future.

ishitatsuyuki:
Status: fixed one nit, others need your decision

lib/proxy-heuristics.js
jankeromnes: > But on the other hand: we don't constraint the replacement to the start of the path, which means it's somewhat flawed too.

Maybe we can replace it with something like `request.url.replace(new RegExp('^/' + containerId + '/' + port), '')`?

ishitatsuyuki: That's a magnitude longer :) From the viewpoint of code length, this is the most simple solution we can have.

lib/proxy-heuristics.js
jankeromnes: If you don't mind, I'd prefer having the exact same destructuring code for `match` everywhere, even if the variables aren't used.

lib/proxy-heuristics.js
jankeromnes: It is, but I continue to believe that using `path` isn't safe for things like query parameters. It seems it will work as intended today, but I don't think such a subtle dependency on what the rest of the code does is very future proof.

ishitatsuyuki:
I decided to swap the controversial changes out.

PR #307, opened 2018-03-24
Pre-configure hub

jankeromnes:


jankeromnes:
@ishitatsuyuki Could you please take a look at these changes, when you get a chance?

jankeromnes:
The only thing that's not clear with this pull request is that, when you enable the GitHub integration in `/settings/integrations/`, you then need to `Reset to default` both configurations `~/.netrc` and `~/.config/hub`, so that the GitHub token can be added to it.

Maybe we should add some UI with explanations in the "Integrations" section (e.g. for the GitHub integration, we could say "Once enabled, please reset `~/.config/hub` and `~/.netrc` configurations. Also, you can revoke tokens at any time [here](https://github.com/settings/applications) and [here](https://github.com/settings/tokens)."?

api/user-api.js
```diff
@@ -156,8 +156,11 @@ configurationsAPI.patch({
       try {
         const json = Buffer.concat(chunks).toString();
         const operations = JSON.parse(json);
-        const changedFiles = operations
-          .map(operation => operation.path.replace(/^\//, ''));
+        // Compute file names by un-escaping the JSON Pointer tokens (RFC 6901).
```
ishitatsuyuki: Can you use an external library instead of rolling your own implementation?

lib/configurations.js
```diff
@@ -5,10 +5,13 @@
 
 const fleau = require('fleau');
 const fs = require('fs');
+const nodepath = require('path');
```
ishitatsuyuki: Nit: prefer camel-case like `nodePath`. Though, can we use just `path` here?

api/user-api.js
jankeromnes: Would you really prefer an external dependency for 3 super simple JS lines, that can be matched by anyone against a very simple spec at any time?

I think that would be overkill, would unnecessarily obfuscate the code while not making it much shorter, and I don't think we want another `left-pad`-like drama.

(That being said, JSON Pointer token is a tiny part of the JSON Patch mechanism, for which we do use an external library `fast-json-patch` that's efficient, well-tested and self-contained.)

ishitatsuyuki: Actually, I couldn't understand why we suddenly have to parse those tokens. It sounds like a dumb question, but can you describe the rationale?

jankeromnes: JSON Pointer allows you to reference keys inside an object, e.g. `/user/email` would reference:

```js
{
  "user": {
    "email": "user@example.com"
  }
}
```

However, if you want to point to a key that contains a `/`, you're in trouble (we do that for configuration templates like `".config/hub"`):

```js
{
  "user": {
    "email": "user@example.com"
  },
  "user/email": "trolling you"
}
```

This is why you escape characters like  `/` with `~1`, making an un-ambiguous JSON Pointer token `/user~1email`.

Then you get the same question with:

```js
{
  "user": {
    "email": "user@example.com"
  },
  "user/email": "trolling you",
  "user~1email": "trolling you harder"
}
```

... which you solve by escaping `~` characters with `~0`: `/user~01email` (now this is clearly asking for trouble, but you get the idea, we're now able to un-ambiguously reference keys in objects even if they contain special JSON Pointer characters like `/` and `~`.)

Note that we didn't actually need to properly unescape before, because we didn't need to reference keys containing `/` until now (but now we have `.config/hub`).

ishitatsuyuki: Thanks for the explanation; please use https://github.com/Starcounter-Jack/JSON-Patch/blob/master/src/helpers.ts#L65

lib/configurations.js
jankeromnes: > Nit: prefer camel-case like `nodePath`.

Ok fair enough, I'll switch to camelCase. Although, please note that lowercase `node`-prefixing is commonplace in our code base (see [nodepath](https://github.com/JanitorTechnology/janitor/search?utf8=%E2%9C%93&q=nodepath&type=) and [nodeurl](https://github.com/JanitorTechnology/janitor/search?utf8=%E2%9C%93&q=nodeurl&type=)). Would you like me to change other occurrences to camelCase as well?

> Though, can we use just `path` here?

Unfortunately not, because line 43 has `const path = nodepath.join(templatesDirectory, file);`.

ishitatsuyuki: Let's do the rest of change either in a batch commit, or whenever the change looks necessary.

jankeromnes:
> Regarding the "update configurations" interface, I think it definitely should prompt the user to push the changes to container. We should do this with a flow of a few clicks, but do note that the change may be destructive.

The fact that they're destructive is why I'd prefer to tell users "please go reset your configs yourself" instead of "please click here to let me destroy stuff". But then, maybe we could have a flag that says "the user modified this config file", and as long as it's off we can auto-reset them? (And if they're hand-modified, but we need to reset them, we can say "you appear to have customized ~/.config/hub, but your token has changed. Please save your changes and then reset to default" or so).

jankeromnes:
@ishitatsuyuki do you think it's OK to merge this as-is, or should we first improve the token/configuration refresh UX?

ishitatsuyuki:
This is a quite long requested feature, we should just merge and improve the UI later.

jankeromnes:
Agreed, thanks! Could you please approve the pull request?

jankeromnes:
Thanks a lot for the review @ishitatsuyuki! 👍

PR #308, opened 2018-03-25
gitignore: add watchman config

ishitatsuyuki:
See: https://facebook.github.io/watchman/docs/config.html

I think we're going to integrate watchman to improve speed of VCS. This is the first step forward.


jankeromnes:
Thanks! :+1:

PR #309, opened 2018-03-25
Always use an explicit tag when pulling a Docker image.

jankeromnes:
Otherwise, Docker might pull all available image tags:
"Tag or digest. If empty when pulling an image, this causes all tags for the
given image to be pulled."
Source: https://docs.docker.com/engine/api/v1.24/#create-an-image

### DIFF
--- a/lib/machines.js
+++ b/lib/machines.js
@@ -71,7 +71,7 @@ exports.pull = function (projectId, callback) {
     return;
   }
 
-  const { host, image } = project.docker;
+  const { host, _baseImage: image } = project.docker;
   const time = Date.now();
 
   docker.pullImage({ host, image }, (error, stream) => {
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
API examples need update. Good otherwise.

jankeromnes:
Oops, this looks like a bug:

```js
[2018-03-25T15:53:06.672Z] pull image:latest:latest started
[2018-03-25T15:53:06.673Z] pull-tag image:latest:latest image:janitor-production success
[fail] Janitor API self-tests
Error: 31/32 API tests passed. Failed tests: [
  {
    "handler": "Pull a project",
    "method": "post",
    "uri": "/api/projects/test-project/pull",
    "request": {
      "urlParameters": {
        "project": "test-project"
      }
    },
    "expectedResponse": {
      "body": "{\n  \"image\": \"image:latest\",\n  \"created\": 1500000000000\n}"
    },
    "actualResponse": {
      "status": 200,
      "body": "{\n  \"image\": \"image:latest:latest\",\n  \"created\": 1500000000000\n}"
    }
  }
]
```

lib/machines.js
```diff
@@ -482,6 +482,15 @@ function getProject (projectId, create) {
     project.description = '';
   }
 
+  // Get a full Docker image ID, with an explicit tag (by default ':latest').
+  if (!project.docker.hasOwnProperty('_baseImage')) {
+    Object.defineProperty(project.docker, '_baseImage', {
+      get () {
+        return this.image + (this.image.includes[':'] ? '' : ':latest');
```
jankeromnes: I don't think `.includes[':']` is the proper way to call methods... (should be `.includes(':')` instead).

jankeromnes:
All nits addressed (no need to update API examples). @ishitatsuyuki please approve this pull request if you like it. 😄

PR #317, opened 2018-04-06
Add default .ssh/authorized_keys user configuration template.

jankeromnes:
This replaces the deprecated https://c9.io SSH public key and the ad-hoc `deploySSHAuthorizedKeys` mechanism.

Also, this gives users more control over which SSH keys should be authrorized to log into their containers (hint: you can find a handy SSH one-liner for your containers at https://janitor.technology/containers-new/).


jankeromnes:
@nt1m please have a look if you have some time. 😄

nt1m:
Seems like it is, since it's a separate commit :)

PR #321, opened 2018-04-10
Add production Dockerfile to ease Janitor deployments.

jankeromnes:


ishitatsuyuki:
Don't forget to mount the credentials as bind mounts; this may be a bit hard since we have multiple data files in repo root.

jankeromnes:
> Don't forget to mount the credentials as bind mounts; this may be a bit hard since we have multiple data files in repo root.

Thanks! But since this is mostly to accommodate Cloud Ops deployments, I think we'll wait and see how they expect credentials to be handled (e.g. via a mounted volume, or environment variables, or some command or API call...)

PR #325, opened 2018-04-25
 Make containers expire after 6 months, but don't auto-destroy them just yet

jankeromnes:


lib/machines.js
```diff
@@ -275,6 +287,16 @@ exports.destroy = function (user, projectId, machineId, callback) {
   });
 };
```
ishitatsuyuki: Can we use camelCase with a lower initial letter here?

lib/machines.js
```diff
@@ -215,6 +218,15 @@ exports.spawn = function (user, projectId, callback) {
 
     const now = Date.now();
     metrics.push(project, 'spawn-time', [ now, now - time ]);
+
+    const expires = now + maximumContainerLifetime;
+    metrics.set(machine, 'expires', expires);
```
ishitatsuyuki: Please perform batch scan instead of scheduling an event for away. This way the logic applies to already spawned containers and we'll be able to change the expiry period in the future.

jankeromnes: What do you mean by "batch scan"? I.e. frequently query "container creation time" of every container to detect expired containers? Or back process all existing containers in order to set an expiration time? Please briefly outline the algorithm you have in mind.

lib/machines.js
jankeromnes: Can do, and I'd find that more natural too, but I took inspiration from existing event systems that apparently often have capitalized event types (e.g. GitHub events from their API).

lib/machines.js
jankeromnes: I agree that we'd need a cron like job that scans & removes expired containers, instead of scheduling events far into the future. For now, I've removed any scheduling altogether, and I'll leave that as a follow-up.

lib/machines.js
jankeromnes: Event removed for now.

Coder206:
@jankeromnes I was wondering what happens to containers after the expired time is reached (without the implementation of the auto-destroy). Does it flag it for you to manually remove?

jankeromnes:
> LGTM!

Thanks!

> I was wondering what happens to containers after the expired time is reached (without the implementation of the auto-destroy). Does it flag it for you to manually remove?

I'm not exactly sure yet, so I'm open to ideas. Ideally, I'd like to soon show a preliminary UI saying something like "Expires in X days" and then "Expired!", encouraging users to delete them manually. Then, I'd also like to send email reminders when a container is about to expire. Finally, when all that's sufficiently obvious, and users have been given advance notice, we can start auto-deleting containers.

@Coder206 Is this something you'd like to help with? :slightly_smiling_face:

Coder206:
> @Coder206 Is this something you'd like to help with? slightly_smiling_face

Yes, count me in :)

jankeromnes:
@Coder206 cool, thanks!

As a proof of concept, I attempted to display container expiration in the `Advanced` tab. Please take a look when you have some time. :smile: 

expired-container

nt1m:
@jankeromnes I think it should just be in the "Info" tab.

jankeromnes:
@nt1m @Coder206 Thanks for your input on my earlier design!

In this new attempt, I simply replace the `Built X months ago` string with `Expired Y months ago` in the "Info" section of a container (having both felt a bit noisy and confusing):

expired-info

Please let me know what you think.

Coder206:
@jankeromnes I like this new design and your plan, it's clean! Would this lead to abandoning the "freshness" metric ("Build x months ago")?

jankeromnes:
> Would this lead to abandoning the "freshness" metric ("Build x months ago")?

Yes, I simply replaced it. Showing both side by side was too confusing, and actually the purpose of "Built X months ago" was just to indicate how old your container image is, while in a new world where containers expire, hopefully all images and containers will always be fresh and recent.

(Once we've successfully enforced 6 months, I'd like to see how low we can go. This will probably be gradual, with added work necessary to make migrating between containers easier, but in an ideal world, containers don't stay alive for too long. They're good as work environments for when you work, but valuable state should be stored somewhere else, e.g. in GitHub branches, or maybe a "safe" within your container that gets auto-migrated, not sure.)

jankeromnes:
(Sorry, I didn't mean to close the pull request!)

jankeromnes:
Thanks again for the feedback! I'll merge this as is, but please feel free to share other opinions, I'll be happy to address anything in follow-ups.

nt1m:
@jankeromnes The freshness metric was a good indicator of whether you needed to rebase your current patch or not. Now that it's not there, there's no way of knowing whether you need to get a new container or not.

While the freshness metric isn't necessarily the best way to display the age information (it would be more useful to have the output of `hg log -r tip` for example), it was the only place where you could find that info.

jankeromnes:
Hm, thanks, you make a good point @nt1m.

Would you have any design suggestions how to display both the "freshness metric" ("Built X months ago", or maybe "Image built X months ago") and the expiration time ("Expires in Y months")? Should one of them be in "Advanced" rather than "Info"?

nt1m:
@jankeromnes I'd display the built date if the expiration date is far off, and the expiration date if the expiration date is arriving soon, or if it's already expired.

PR #340, opened 2018-06-07
Set mode to 0600 to all config files

beaufortfrancois:
I wasn't able to try this in my local janitor container. Please test first.

@jankeromnes

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -229,7 +229,7 @@ exports.copyIntoContainer = function (parameters, callback) {
     // Add the files to a tar stream for Docker's Remote API.
     const pack = tar.pack();
     for (const name in files) {
-      pack.entry({ name, mode: 600 }, files[name]);
+      pack.entry({ name, mode: 0600 }, files[name]);
     }
     pack.finalize();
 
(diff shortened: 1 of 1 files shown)


lib/docker.js
```diff
@@ -229,7 +229,7 @@ exports.copyIntoContainer = function (parameters, callback) {
     // Add the files to a tar stream for Docker's Remote API.
     const pack = tar.pack();
     for (const name in files) {
-      pack.entry({ name }, files[name]);
+      pack.entry({ name, mode: 600 }, files[name]);
```
jankeromnes: Nit: `mode` seems to require an octal value: see `tar-stream`'s [readme example](https://github.com/mafintosh/tar-stream#headers) and its [tests](https://github.com/mafintosh/tar-stream/blob/c0b13f7329d0459310d75e4dacf16aa9e960b7b5/test/pack.js#L15).

lib/docker.js
```diff
@@ -229,7 +229,7 @@ exports.copyIntoContainer = function (parameters, callback) {
     // Add the files to a tar stream for Docker's Remote API.
     const pack = tar.pack();
     for (const name in files) {
-      pack.entry({ name }, files[name]);
+      pack.entry({ name, mode: 0600 }, files[name]);
```
nt1m: I prefer 0o600 if you're going to write octal

jankeromnes:
Also fixed @nt1m's nit, and merging soon.

PR #353, opened 2018-07-23
Add link tag for RSS feed

nt1m:


templates/blog.html
```diff
@@ -1,3 +1,4 @@
+    <link rel="alternate" type="application/rss+xml" href="https://discourse.janitor.technology/c/propaganda/blog.rss"/>
```
jankeromnes: Nit: `<link>` is a [void element](https://www.w3.org/TR/2012/WD-html-markup-20120320/syntax.html#syntax-elements), so the extra `/` at the end is just decoration/noise. Feel free to remove it.

PR #355, opened 2018-07-24
Redirect old /*-new/ URLs to the right place.

jankeromnes:
Examples from our logs:

```
[2018-07-24T08:16:47.030Z] 404 /projects-new/
[2018-07-24T08:16:47.649Z] 404 /about/
[2018-07-24T08:16:47.996Z] 404 /projects-new
[2018-07-24T08:16:51.123Z] 404 /reference/api-new/
[2018-07-24T08:16:51.510Z] blog
[2018-07-24T08:16:52.020Z] 404 /settings-new/
[2018-07-24T08:16:55.996Z] 404 /blog-new/
```

### DIFF
--- a/app.js
+++ b/app.js
@@ -92,6 +92,11 @@ boot.executeInParallel([
     routes.landingPage(query.res, user);
   });
 
+  // Public landing page (legacy).
+  app.route(/^\/landing-new\/?$/, (data, match, end, query) => {
+    routes.redirect(query.res, '/');
+  });
+
   // Public API (when wrongly used with a trailing '/').
   app.route(/^\/api\/(.+)\/$/, (data, match, end, query) => {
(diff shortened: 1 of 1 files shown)


jankeromnes:
Thanks @notriddle!

jankeromnes:
Hm, I wonder if we shouldn't have edited the existing route RegExps to optionally support `-new`, because this approach won't work with sub-pages, nor will it keep query parameters or anchors (e.g. `/containers-new/?key=1234`, `/blog-new/#janitor-news-5`)

notriddle:
What does the `?key` even do on containers-new?

jankeromnes:
It's the login key from email URLs (but ok, it's a very unlikely example as we've never shared URLs like these).

PR #361, opened 2018-08-05
centering feedback icon

arshdkhn1:


arshdkhn1:
oops

nt1m:
I'd be fine with some specific CSS for each individual page (using the individual CSS files).

PR #373, opened 2018-08-21
UX: Make deleting containers faster and more consistent.

jankeromnes:
There is no need to make users wait for Docker when removing a container.

We can quickly remove it from the user's account, then wait for Docker to finish
cleaning up in the background. If cleaning up fails, we log the error and save
container ID and host, so that we can try again later.


jankeromnes:
@nt1m Could you please have a look when you have time? :smile:

PR #378, opened 2018-08-28
Add a search box to the projects page

notriddle:


static/css/projects.css
```diff
@@ -0,0 +1,14 @@
+.search-section {
+    display: flex;
+    flex-direction: row;
+}
+.search-section label {
+    line-height: 30px;
+}
+.search-section input {
+    flex-grow: 1;
```
jankeromnes: Nit: Please use 2-space indents (here and above).

static/js/projects.js
```diff
@@ -48,3 +48,28 @@ Array.forEach(timestampElements, function (element) {
   // Use live-updating timeago plugin.
   timeago().render(element);
 });
+
+var cardSearchBox = document.querySelector('[data-search-cards]');
+var cards = document.querySelectorAll('.card');
+Array.forEach(cards, function (element) {
+  var header = element.querySelectorAll('h2');
+  element._janitor_header = header[0].innerText.toLowerCase();
```
jankeromnes: Nit: Why search in just the `h2` and not the entire `card.innerText.toLowerCase()`?

templates/projects.html
```diff
@@ -1,5 +1,9 @@
       <main class="container">
-        <div class="grid">{{
+        <section class="search-section">
+          <label for='project-search'>Search</label>
+          <input class='form-control' id='project-search' data-search-cards='projects-grid'>
```
jankeromnes: Nit: Please use double quotes `"` instead of single quotes `'` in HTML code when possible.

jankeromnes: Also, have you considered adding the `<input>` attribute `type="search"`?

static/js/projects.js
```diff
@@ -48,3 +48,28 @@ Array.forEach(timestampElements, function (element) {
...
+});
+cardSearchBox.addEventListener('input', function() {
+  var words = cardSearchBox.value.toLowerCase().split(/\s+/);
+  Array.forEach(cards, function(element) {
```
jankeromnes: Nit: Please add a space between `function` and `(` (and for the other functions as well).

static/js/projects.js
```diff
@@ -48,3 +48,28 @@ Array.forEach(timestampElements, function (element) {
...
+    var cardHeader = element._janitor_header;
+    var hide = !Array.some(words, function(word) {
```
jankeromnes: Nit: Please polyfill `Array.some` [here](https://github.com/JanitorTechnology/janitor/blob/12b711e3cd5dbd3e6355407bd030ed4bc9432063/static/js/janitor-new.js#L19-L24) for older browsers.

templates/projects.html
jankeromnes: Also, maybe it's worth adding a `placeholder` attribute to explain that it matches any word (e.g. `Search projects including any of these words…`)?

nt1m:
Here's a mockup of how the searchbox could look like:

screen shot 2018-08-30 at 13 29 59

nt1m:
The icon is https://design.firefox.com/icons/icons/desktop/search-16.svg with fill `rgba(12,12,13,0.4)`

static/js/projects.js
```diff
@@ -48,3 +48,27 @@ Array.forEach(timestampElements, function (element) {
...
+  Array.forEach(cards, function (element) {
+    var cardHeader = element._janitor_header;
+    var hide = !Array.some(words, function (word) {
+      return cardHeader.indexOf(word) < 0;
+    });
+    if (hide) {
+      element.removeAttribute('hidden');
+    } else {
+      element.setAttribute('hidden', true);
+    }
```
nt1m: nit: element.hidden = !matchesQuery;

static/js/projects.js
```diff
@@ -48,3 +48,27 @@ Array.forEach(timestampElements, function (element) {
...
+var cardSearchBox = document.querySelector('[data-search-cards]');
+var cards = document.querySelectorAll('.card');
+Array.forEach(cards, function (element) {
+  element._janitor_header = element.innerText.toLowerCase();
+});
+cardSearchBox.addEventListener('input', function () {
+  var words = cardSearchBox.value.toLowerCase().split(/\s+/);
+  Array.forEach(cards, function (element) {
+    var cardHeader = element._janitor_header;
+    var hide = !Array.some(words, function (word) {
```
nt1m: nit: Can we call this variable`matchesQuery` ?

static/js/projects.js
```diff
@@ -48,3 +48,27 @@ Array.forEach(timestampElements, function (element) {
   // Use live-updating timeago plugin.
   timeago().render(element);
 });
+
+var cardSearchBox = document.querySelector('[data-search-cards]');
+var cards = document.querySelectorAll('.card');
+Array.forEach(cards, function (element) {
+  element._janitor_header = element.innerText.toLowerCase();
```
nt1m: Can we compute this server-side in the projects.html template, as a data-query attribute ?

static/js/projects.js
```diff
@@ -48,3 +48,27 @@ Array.forEach(timestampElements, function (element) {
   // Use live-updating timeago plugin.
   timeago().render(element);
 });
+
+var cardSearchBox = document.querySelector('[data-search-cards]');
```
nt1m: Are we actually respecting the value of the data attribute ? If not, I prefer that we use `document.getElementById("searchbox");`

static/js/projects.js
```diff
@@ -48,3 +48,27 @@ Array.forEach(timestampElements, function (element) {
...
+      return cardHeader.indexOf(word) < 0;
+    });
+    if (hide) {
+      element.removeAttribute('hidden');
+    } else {
+      element.setAttribute('hidden', true);
+    }
+  });
+});
+cardSearchBox.addEventListener('submit', function (e) {
```
nt1m: Do we need this ? the searchbox doesn't seem like it's inside a form.

static/js/projects.js
notriddle: We weren't, but we should. Fixed that.

static/js/projects.js
notriddle: Yeah. That's probably better, anyway, since we don't want to count the "New Container" button as part of the text to search.

static/js/projects.js
notriddle: It used to be, but yeah, it's not any more.

static/js/projects.js
```diff
@@ -48,3 +48,16 @@ Array.forEach(timestampElements, function (element) {
...
 });
+
+var cardSearchBox = document.querySelector('[data-search-cards]');
+cardSearchBox.addEventListener('input', function () {
+  var words = cardSearchBox.value.toLowerCase().split(/\s+/);
+  var cardsContainerId = cardSearchBox.getAttribute('data-search-cards');
+  var cards = document.getElementById(cardsContainerId).querySelectorAll('.card');
+  Array.forEach(cards, function (element) {
+    var cardHeader = element.getAttribute('data-search-text');
+    element.hidden = Array.some(words, function (word) {
```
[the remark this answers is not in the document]
notriddle: No. `cardHeader.indexOf(word) < 0` means *word is not present in cardHeader*.

If some word is not present in cardHeader, then element should be hidden.

notriddle:
Fixed all the nits.

notriddle:
Rebased.

jankeromnes:
Thanks! Will merge and deploy later today. :smile:

PR #407, opened 2018-12-14
Add Visible Link to RSS Feed

### LINKED ISSUE #299: add an RSS/Atom feed to the blog
Not sure where else to report this; please close this if this is the wrong place!

It'd be great to have an RSS and/or Atom feed for the blog, so people would have an easier time following along in their favorite feed reader.

Coder206:
Fixes #299

### DIFF
--- a/templates/blog.html
+++ b/templates/blog.html
@@ -1,6 +1,8 @@
     <link rel="alternate" type="application/rss+xml" href="https://discourse.janitor.technology/c/propaganda/blog.rss">
     <main class="container">
-      <section class="blog">{{
+      <section class="blog">
+        <a class="btn btn-primary" href="https://discourse.janitor.technology/c/propaganda/blog.rss">RSS Feed</a>
+        {{
           for (const post of posts) {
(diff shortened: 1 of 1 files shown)


templates/blog.html
```diff
@@ -1,6 +1,8 @@
     <link rel="alternate" type="application/rss+xml" href="https://discourse.janitor.technology/c/propaganda/blog.rss">
     <main class="container">
-      <section class="blog">{{
+      <section class="blog">
+        <a class="btn btn-primary" href="https://discourse.janitor.technology/c/propaganda/blog.rss">RSS Feed</a>
+        {{
```
Coder206: Not sure I like moving this to a new line

notriddle: Shouldn't it be this?

        <a class="btn btn-primary" href="https://discourse.janitor.technology/c/propaganda/blog.rss">RSS Feed</a>{{

Coder206: I like that more, thank you @notriddle

PR #412, opened 2019-01-07
[WIP] Automatically alert about and then remove expired containers

jankeromnes:
(Another old commit that I salvaged from an old Janitor container. Will work more on it soon. Probably doesn't work.)


jankeromnes:
There was also this:

https://gist.github.com/jankeromnes/1581bb88b076b17c98b8fe15a514be29

PR #417, opened 2019-01-21
Order topics by creation instead of update

notriddle:
This prevents topics from being un-published

https://discourse.janitor.technology/t/our-latest-newsletter-janitor-news-11-keeps-getting-unpublished/241

### DIFF
--- a/lib/blog.js
+++ b/lib/blog.js
@@ -28,7 +28,7 @@ function fetchDiscourseAPI (url) {
     const urlj = url + auth_param;
     const options = {
       headers: {
-        "Accept": "application/json"
+        'Accept': 'application/json'
       }
     };
     https.get(urlj, options, response => {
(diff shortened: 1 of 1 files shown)


lib/blog.js
```diff
@@ -14,8 +14,7 @@ const metrics = require('./metrics');
 // [^1] "published" is a staff-only-tag.
 // [^2] "blog" is a category, and is not staff-only.
 
-const discourseBlogCategoryID = 9;
-const discourseBlogTopicsUrl = 'https://discourse.janitor.technology/tags/published';
+const discourseBlogTopicsUrl = 'https://discourse.janitor.technology/tags/published?category=9&parent_category=8&order=created&ascending=false';
```
jankeromnes: Nit: When visiting https://discourse.janitor.technology/tags/published?category=9&parent_category=8&order=created&ascending=false, the post order looks mostly right, except:

- `Janitor Alpha is here!` seems to be the second most recent
- `Janitor News #9` is all the way down as the oldest

jankeromnes: Also, nit, shouldn't that be `published.json`, since you've removed `.json` below?

notriddle: Looks like it was a problem with the old article migration. It's correctly sorted now.

jankeromnes: Newsletter 9 is indeed back in the right place, but not Janitor Alpha. But that can probably be fixed in Discourse, and is orthogonal to this pull request.

PR #440, opened 2019-03-13
Code Review: Amend the License Owners

Coder206:
This PR is mostly just changing the names of the owners of the project's files to "Team Janitor" as discussed with @jankeromnes in IRC.

### DIFF
--- a/lib/streams.js
+++ b/lib/streams.js
@@ -1,4 +1,4 @@
-// Copyright © 2016 Jan Keromnes. All rights reserved.
+// Copyright © 2016 Team Janitor. All rights reserved.
 // The following code is covered by the AGPL-3.0 license.
 
 const stream = require('stream');
@@ -9,7 +9,6 @@ const log = require('./log');
 const objectStreams = new Map();
 
 // Get `object[key]` as a stream (may still be receiving new data).
-
 exports.get = function (object, key) {
   const passthrough = new stream.PassThrough();
   const streams = objectStreams.get(object);
@@ -28,7 +27,6 @@ exports.get = function (object, key) {
 };
 
 // Read a stream into `object[key]`.
-
 exports.set = function (object, key, readable) {
   let streams = objectStreams.get(object);
 
@@ -61,7 +59,6 @@ exports.set = function (object, key, readable) {
 };
 
 // Remove any stream affected to `object[key]`.
-
 exports.remove = function (object, key) {
   const streams = objectStreams.get(object);
 
--- a/static/js/blog.js
+++ b/static/js/blog.js
@@ -1,3 +1,6 @@
+// Copyright © Team Janitor. All rights reserved.
+// The following code is covered by the AGPL-3.0 license.
+
 function expandPost (hash) {
   // If hash is an empty string then select first blog
   var blogId = hash.slice(1) || document.querySelector('.article-wrapper h1').id;
--- a/static/js/janitor-new.js
+++ b/static/js/janitor-new.js
@@ -1,4 +1,4 @@
-// Copyright © 2015 Jan Keromnes. All rights reserved.
+// Copyright © 2015 Team Janitor. All rights reserved.
 // The following code is covered by the AGPL-3.0 license.
 
 // Helpers
@@ -50,7 +50,7 @@ Array.forEach($$('.tabs'), function (element) {
 Array.forEach(document.querySelectorAll('form[method]'), function (form) {
   setupAsyncForm(form);
   form.addEventListener('submit', function (event) {
-    // set form to pending status
+    // Set form to pending status
     updateFormStatus(form, 'pending');
 
     var elements = Array.filter(form.elements, function (element) {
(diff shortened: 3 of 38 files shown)


lib/streams.js
```diff
@@ -9,7 +9,6 @@ const log = require('./log');
 const objectStreams = new Map();
 
 // Get `object[key]` as a stream (may still be receiving new data).
-
```
Coder206: I want to keep things consistent with the other files

lib/streams.js
```diff
@@ -28,7 +27,6 @@ exports.get = function (object, key) {
 };
 
 // Read a stream into `object[key]`.
-
```
Coder206: See previous

lib/streams.js
```diff
@@ -61,7 +59,6 @@ exports.set = function (object, key, readable) {
 };
 
 // Remove any stream affected to `object[key]`.
-
```
Coder206: See previous

static/js/blog.js
```diff
@@ -1,3 +1,6 @@
+// Copyright © Team Janitor. All rights reserved.
+// The following code is covered by the AGPL-3.0 license.
+
```
Coder206: This should be under the license as well because it seems to be ours.

static/js/janitor-new.js
```diff
@@ -50,7 +50,7 @@ Array.forEach($$('.tabs'), function (element) {
 Array.forEach(document.querySelectorAll('form[method]'), function (form) {
   setupAsyncForm(form);
   form.addEventListener('submit', function (event) {
-    // set form to pending status
+    // Set form to pending status
```
Coder206: I want to keep the comments consistent

jankeromnes:
> In other news, when looking over the code, I noticed a few occurrences similar to this: https://github.com/JanitorTechnology/janitor/blob/master/lib/machines.js#L410
Could we have these sort of statements at the top of the files like so:
https://github.com/JanitorTechnology/janitor/blob/master/lib/machines.js#L12
Or is there location strategic to get up-to-date data?

Thanks for noticing this! (Tip: When you click on a line to get these links, you can also type the letter `Y` on your keyboard to get commit-specific URLs that remain stable over time.)

I think it's OK to use `db.get('thing')` both locally (e.g. when you just need it in that function), and at the top of a file (e.g. when you use `thing` in many places in that file).

But in `lib/machines.js` for example, moving `const ports = db.get('ports')` to the top of the file would conflict with:

https://github.com/JanitorTechnology/janitor/blob/f9dc2d88e2645d923307db59c9023631a989dfe1/lib/machines.js#L195

This can of course be addressed by renaming it one of the conflicting symbols. TL;DR feel free to hoist `db.get('thing')`s to the top of files, but please make sure to fix any conflicts in the process. 😄

Coder206:
@jankeromnes Thank you for the review!

> (Tip: When you click on a line to get these links, you can also type the letter Y on your keyboard to get commit-specific URLs that remain stable over time.)
Neat, thanks! The URLs are https://github.com/JanitorTechnology/janitor/blob/f9dc2d88e2645d923307db59c9023631a989dfe1/lib/machines.js#L410 and https://github.com/JanitorTechnology/janitor/blob/f9dc2d88e2645d923307db59c9023631a989dfe1/lib/machines.js#L12 in that case :)

> > In other news, when looking over the code, I noticed a few occurrences similar to this: https://github.com/JanitorTechnology/janitor/blob/master/lib/machines.js#L410
> > Could we have these sort of statements at the top of the files like so:
> > https://github.com/JanitorTechnology/janitor/blob/master/lib/machines.js#L12
> > Or is there location strategic to get up-to-date data?
> 
> Thanks for noticing this! (Tip: When you click on a line to get these links, you can also type the letter `Y` on your keyboard to get commit-specific URLs that remain stable over time.)
> 
> I think it's OK to use `db.get('thing')` both locally (e.g. when you just need it in that function), and at the top of a file (e.g. when you use `thing` in many places in that file).
> 
> But in `lib/machines.js` for example, moving `const ports = db.get('ports')` to the top of the file would conflict with:
> 
> [janitor/lib/machines.js](https://github.com/JanitorTechnology/janitor/blob/f9dc2d88e2645d923307db59c9023631a989dfe1/lib/machines.js#L195)
> 
> Line 195 in [f9dc2d8](/JanitorTechnology/janitor/commit/f9dc2d88e2645d923307db59c9023631a989dfe1)
>  const ports = {}; 
> 
> This can of course be addressed by renaming it one of the conflicting symbols. TL;DR feel free to hoist `db.get('thing')`s to the top of files, but please make sure to fix any conflicts in the process. smile

Good catch, I missed that. Thanks! :)


> Approved, many thanks for addressing this!
> 
> I think the `All rights reserved.` doesn't actually mean anything, but it also doesn't hurt to have it, so +1 .
> 
> > Submodule dockerfiles updated 19 files
> 
> Did you also sneak in a Dockerfiles submodule upgrade? :) That's cool with me too.

Thank you! I think the Dockerfiles submodule upgrade was an artefact of using `git pull` for this keeping this PR in sync with master.

